PR #18, opened 2022-06-17
Fail build on wildcard imports

### LINKED ISSUE #3: [BUG] Fail build when `*` imports are used
### What is the bug?
Currently `*` imports are being used and should have a way to fail the build if found. 

_Steps to reproduce the behavior._

_A clear and concise description of what you expected to happen._

### What is your host/environment?
_Operating system, version._

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

_Add any other context about the problem._

dbwiddis:

### Description
Fails the build if a regex matching a wildcard import is encountered.

### Issues Resolved
Closes #3  

### DIFF
--- a/gradle/formatting.gradle
+++ b/gradle/formatting.gradle
@@ -10,6 +10,13 @@ allprojects {
             eclipse().configFile rootProject.file('formatter/formatterConfig.xml')
             trimTrailingWhitespace()
 
+            custom 'Refuse wildcard imports', {
+                // Wildcard imports can't be resolved; fail the build
+                if (it =~ /\s+import .*\*;/) {
+                    throw new AssertionError("Do not use wildcard imports.  'spotlessApply' cannot resolve this issue.")
+                }
+            }
+
             // See DEVELOPER_GUIDE.md for details of when to enable this.
             if (System.getProperty('spotless.paddedcell') != null) {
                 paddedCell()
--- a/src/main/java/org/opensearch/sdk/ActionListener.java
+++ b/src/main/java/org/opensearch/sdk/ActionListener.java
@@ -15,7 +15,11 @@ import org.opensearch.transport.TransportRequestOptions;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
-import java.net.*;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.ServerSocket;
+import java.net.Socket;
+import java.net.UnknownHostException;
 
 public class ActionListener {
 
--- a/src/main/java/org/opensearch/sdk/netty4/Netty4MessageChannelHandler.java
+++ b/src/main/java/org/opensearch/sdk/netty4/Netty4MessageChannelHandler.java
@@ -33,7 +33,12 @@
 package org.opensearch.sdk.netty4;
 
 import io.netty.buffer.ByteBuf;
-import io.netty.channel.*;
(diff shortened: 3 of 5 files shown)


peternied (commented):
What about pulling in checkstyle to enforce this [check] instead of writing it on your own(https://checkstyle.org/config_imports.html#AvoidStarImport)?

owaiskazi19:
> What about pulling in checkstyle to enforce this [check] instead of writing it on your own(https://checkstyle.org/config_imports.html#AvoidStarImport)?

We have removed CheckStyle from OpenSearch to have one formatting tool(Spotless) throughout the repo and following the same here as well.

dbwiddis:
Ready for review.  [Initial test failure after adding the check](https://github.com/opensearch-project/opensearch-sdk/runs/6942878760?check_suite_focus=true).  Tests pass after removing star imports in c0f119d99d550d1a84ffa02cc7ededc1f428d252

dbwiddis:
Note: the error message isn't very informative and I had to hunt down the star imports; however, for future failures they'll likely be part of the diff and much easier to identify.   [It may be possible to get line numbers](https://stackoverflow.com/questions/27089509/how-to-get-access-to-line-number-in-file-copy-filter-task) but that adds some complexity.

peternied:
> We have removed CheckStyle from OpenSearch to have one formatting tool(Spotless) throughout the repo and following the same here as well.

From https://github.com/opensearch-project/OpenSearch/issues/1362, I see 'it caused issue on plugins side '.  Seems like this is a cause of tight build system coupling rather than the tool itself being a problem.

While I agree that removing * imports is a good thing, lack of built-in IDE integration and a host of features that might need to be duplicated for future coverage will be the cost of this pattern.

gradle/formatting.gradle
```diff
@@ -10,6 +10,13 @@ allprojects {
             eclipse().configFile rootProject.file('formatter/formatterConfig.xml')
             trimTrailingWhitespace()
 
+            custom 'Refuse wildcard imports', {
+                // Wildcard imports can't be resolved; fail the build
+                if (it =~ /\s+import .*\*;/) {
```
peternied, line 15:
```suggestion
                if (it =~ /\s*import.*\*;/) {
```

dbwiddis: Imports nearly always start after a new line and the `\s+` was in preference to `[\n|\r|\n\r]\s*`.  There should always be a `package` declaration prior to the imports, although I'm not sure we have a style check to enforce declaring a package.

I think `import*\*;` would behave identically to `\s*import.*\*;` and don't think that's what we want.

peternied (approved):
Keeping this commit unblocked for merging to keep progress moving on this pull request.  If we encounter 'checks' that are already implemented via checkstyle, I think it might be worthwhile to revisit some of these comments.

Thanks for the great back-and-forth @dbwiddis

owaiskazi19:
> From [opensearch-project/OpenSearch#1362](https://github.com/opensearch-project/OpenSearch/issues/1362), I see 'it caused issue on plugins side '. Seems like this is a cause of tight build system coupling rather than the tool itself being a problem.

We wanted to have only one formatting tool for OpenSearch and that's why asked the Plugin teams to follow the same. Does CheckStyle provides more benefits than Spotless? Something we can think of for the future definitely.

PR #29, opened 2022-06-28
Document security scope for extensions

peternied:
### Description
While  working on extensions there are many considerations for security, this document captures what those areas are as a way to discuss design approaches and codify the current state of affairs.

To approve this document means that we have `agreement of areas and 'as-is' state of OpenSearch Plugins and Extensions.` There are more details as to the phases of this document over time within the doc itself

### Issues Resolved
* Related https://github.com/opensearch-project/opensearch-sdk/issues/24
* Related https://github.com/opensearch-project/security/issues/1895


peternied:
@opensearch-project/security @CEHENKLE @dblock @saratvemulapalli @dbwiddis What do you think of using a document like this for extensions security related features?

SECURITY.md
```diff
@@ -0,0 +1,57 @@
+# Extensions Security Guidelines
+
+OpenSearch's support for extensions allows for taking already powerful use cases and expanding on them, this creates a larger surface area for misuse, vunerabilities, and malicious interactions.  This document outlines several areas for enhancements, features, and practices to incorperate into extensions for OpenSearch
```
dbwiddis, line 3: ...expanding on them. With this increased functionality comes a larger surface area....

SECURITY.md
```diff
@@ -0,0 +1,57 @@
...
+
+To keep concepts consistant, this document is using terminology from [NIST Glossary](https://csrc.nist.gov/glossary).
+
+Additional terms:
+* **Plugin** - reference to the existing functionality to extend OpenSearch functionality.
+* **Extension** - reference to the in development functionality to extend OpenSearch.
+
+## Host security
+
+The Java Security Manager is the mechanism for ensuring plugins are limited in what they can do to the host operation system resources (cpu/disk/memory/network/...).  As there are limitations and its deprecated with removal scheduled in the next release of the JVM.
```
dbwiddis, line 13: I'm unclear about "is the mechanism" vs. "it's deprecated".    If it's on its way out this should be much more past-tense focused, e.g., "While the Java Security Manager has been.... " and add on "Additional measures are needed to protect system resources."

SECURITY.md
```diff
@@ -0,0 +1,57 @@
...
+
+The current extensions design they operate via Rest APIs, by isolating extensions from using host they are prevented from executing operation system calls directly on hosts of the cluster.
```
dbwiddis, line 15: REST should be all caps.  While isolating is a benefit it also brings with it the requirement for extensions to follow strict APIs which may enforce limits.

SECURITY.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Communications security (COMSEC)
+
+Data is transferred from the OpenSearch cluster to the extensions.  This is done via https requests between the nodes on the cluster and the extensions endpoint(s).
```
dbwiddis, line 19: Let's avoid passive voice here: The OpenSearch cluster transfers data to the extensions via https requests between the nodes on the cluster and the extensions endpoint(s).

Does this explain two way communications and how the extensions themselves communicate?  I see Extension <--->  Extension points <----> Cluster.

SECURITY.md
```diff
@@ -0,0 +1,57 @@
...
+
+Extensions should never directly communicate with other extensions, cross extensions work should always be proxied through OpenSearch.
```
dbwiddis, line 21: Let's add the why.  "To ensure security and identity best practices are followed ..."

SECURITY.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Data Security
+
+OpenSearch stores data in its memory and local file system storage, the security plugin provides mechanisms to control data access within OpenSearch.  Extensions have independent data storage.
```
dbwiddis, line 25: Is the security plugin going to remain as-is? That's not clear here.  I think it's a good idea if so but we should articulate that.

SECURITY.md
```diff
@@ -0,0 +1,57 @@
...
+
+Plugins store data inside of the OpenSearch cluster itself such as in system/hidden indices.
```
dbwiddis, line 27: Plugins do store, or may store?

SECURITY.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Access Control
+
+OpenSearch offers access control through the security plugin, with checks action names and filters.  Actions registered within OpenSearch that are not permitted never reach the handler for a plugin or extension execution.
```
dbwiddis, line 31: This needs a lot more detail, or perhaps just a link to the security plugin docs.

SECURITY.md
```diff
@@ -0,0 +1,57 @@
...
+
+Resource level access control is governed by the extension, when requests are processed the [user](https://github.com/opensearch-project/common-utils/blob/main/src/main/java/org/opensearch/commons/authuser/User.java) object from common-utils is checked for matching backendroles/roles.  Access control checks are managed wholy in the plugin.  Example from anomaly detection, [checkUserPermissions](https://github.com/opensearch-project/anomaly-detection/blob/875b03c1c7596cb34d74fea285c28d949cfb0d19/src/main/java/org/opensearch/ad/util/ParseUtils.java#L568).
```
dbwiddis, line 33: This needs an "As is" and "To Be" summary.  We are planning to do something else with User, aren't we?  You're also mixing extension and plugin here, not sure if this is documenting current vs. future.   I would like to put RBAC as a goal.

SECURITY.md
```diff
@@ -0,0 +1,57 @@
...
+
+Available permissions and roles are defined in the security plugin, every extension needs to update the security plugin to provide these values, eg. [roles.xml](https://github.com/opensearch-project/security/blob/main/config/roles.yml).
```
dbwiddis, line 35: Possibly currently.  In the future I would want permissions and roles to be identity-based using open standards for RBAC.

SECURITY.md
```diff
@@ -0,0 +1,57 @@
...
+
+With the security plugin installed, when actions are performed on the OpenSearch cluster they are recorded if filtering criteria are meet to configurable audit log sinks.
+
+## Installation
+
+Plugin installation is managed by using a binary on the node that extract plugin.zip files into the file system, this is done outside the active running of OpenSearch itself.  When OpenSearch starts it loads installed plugins into its JVM runtime.
+
+## Versioning
+
+OpenSearch systems have ways to deprecate unsupported patterns, feature, and APIs.
```
dbwiddis, line 47: Should we mention Semantic Versioning standards?

SECURITY.md
```diff
@@ -0,0 +1,57 @@
...
+
+Available permissions and roles are defined in the security plugin, every extension needs to update the security plugin to provide these values, eg. [roles.xml](https://github.com/opensearch-project/security/blob/main/config/roles.yml).
+
+## Auditing
+
+With the security plugin installed, when actions are performed on the OpenSearch cluster they are recorded if filtering criteria are meet to configurable audit log sinks.
+
+## Installation
+
+Plugin installation is managed by using a binary on the node that extract plugin.zip files into the file system, this is done outside the active running of OpenSearch itself.  When OpenSearch starts it loads installed plugins into its JVM runtime.
```
dbwiddis, line 43: This seems to be an implementation detail that we don't need to specify here.

dbwiddis (commented):
Good first start. I like the distinction between plugin and extension up front but then I'm starting to miss the usage later on as everything seems to say "plugin". Is it documenting current state?  

It may be good to organize the sections into "current state" (plugin) and "future state" (extension + select plugins) to make clear what we are scoping for in the future.

dbwiddis:
> @opensearch-project/security @CEHENKLE @dblock @saratvemulapalli @dbwiddis What do you think of using a document like this for extensions security related features?

I think a _template_ with a subset of important decision points similar to a document like this is a good idea once this document is ironed out.   This document can serve more as a guiding principle while extensions should complete the template with implementation specifics for that extension.

SECURITY.md
peternied: 💯 on the passive voice comment, thanks for noticing

SECURITY.md
peternied: Rewording to make this clearer - let me know what you think of the revision

SECURITY.md
peternied: Reworked this section, does this sound better?

SECURITY.md
peternied: > I would like to put RBAC as a goal.

Me too, I think there is a version of this document that includes lists of issues to be implemented/decisions to make.  Before we come to the possible implementations I'd like to lay out the surface area for prioritization.  How would you feel about holding off on individual features until we codify this framework?

SECURITY.md
```diff
@@ -0,0 +1,71 @@
+# Extensions Security Guidelines
+
+OpenSearch's support for extensions allows for taking already powerful use cases and expanding on them. With this increased functionality comes a larger surface area for misuse, vunerabilities, and malicious interactions.
+
+By capturing the current state of OpenSearch ecocsystem and the plans for extensions this document outlines several areas for enhancements, features, and practices to incorperate into extensions for OpenSearch.
+
+## Document Phases
```
peternied, line 7: @dbwiddis Following your feedback added this section to capture how to approve/reject this document.

SECURITY.md
peternied: OpenSearch hardly has requirements in this space right now, here are some areas of future recommendations:
* authenticity at installation, dev plugin vs self-signed vs from a registry
* authenticity during runtime, how OpenSearch is sure the extension endpoint didn't change, or vice-versa
* installation gates, allow some plugins to be install but not any plugin
* controlling installed extensions capabilities, restrict what a plugin has access to do

SECURITY.md
```diff
@@ -0,0 +1,79 @@
...
+* **Plugin** - reference to the existing functionality to extend OpenSearch functionality.
+* **Extension** - reference to the in development functionality to extend OpenSearch.
+
+# Areas
+
+## Host security
+
+Plugins depend on use of the Java Security Manager is use to limit interactions on the host operation system resources (cpu/disk/memory/network/...).  JSM has been deprecated, with its removal scheduled in the next release of the JVM, see [OpenSearch discussion](https://github.com/opensearch-project/OpenSearch/issues/1687). Additional measures are needed to protect system resources.
+
+Extensions are sandboxed from the host system by operating via REST APIs.  This security boundary isolates extensions from executing operation system calls directly on OpenSearch hosts.
```
davidlago, line 29: This is only true for out-of-process extensions. How about the ones that can run in-process? I think we are still open to having those supported down the line

SECURITY.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Data Security
+
+OpenSearch stores data in memory and local file system storage.  This data is stored unencrypted.
+
+Plugins can use the existing data systems of the OpenSearch.  Several classes of plugins extend storage options out to external services.
+
+### Access Control
+
+With the security plugin installed, role based access control (RBAC) is available with a proprietary policy document format.  Access control over native OpenSearch data is possible with this plugin installed.
```
davidlago, line 45: I think the word proprietary might not be the one you're looking for here (see https://en.wikipedia.org/wiki/Proprietary_file_format). Perhaps "specific to OpenSearch", "non-standard" (if there is a standard for these...?

SECURITY.md
```diff
@@ -0,0 +1,79 @@
...
+
+For resource that are managed by plugins, access control is governed within individual plugin, by examining [user](https://github.com/opensearch-project/common-utils/blob/main/src/main/java/org/opensearch/commons/authuser/User.java) object from OpenSearch's thread context permissions are available for approval/denial. Example from anomaly detection, [checkUserPermissions](https://github.com/opensearch-project/anomaly-detection/blob/875b03c1c7596cb34d74fea285c28d949cfb0d19/src/main/java/org/opensearch/ad/util/ParseUtils.java#L568).  Uniform resource controls and models are needed to protect from misconfiguration and code defects.
+
+As Extensions do not have access OpenSearch's thread context, identity and its associated privileges must be communicated through the REST APIs.
```
davidlago, line 49: this is another statement that might need to be revisited if we allow in-process extensions

SECURITY.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Auditing
+
+With the security plugin installed, when actions are performed on the OpenSearch cluster they are recorded if filtering criteria are meet to configurable audit log sinks.
+
+## Installation
+
+Plugin installation is managed by using a binary on the node that extract plugin zip files into the file system, this is done outside the active running of OpenSearch itself.  When OpenSearch starts it loads installed plugins into its JVM runtime.
+
+Extensions installation is managed through on disk configuration.
```
davidlago, line 59: I'm not clear on what this means (and specifically how it relates to security)

SECURITY.md
```diff
@@ -0,0 +1,79 @@
...
+
+Extensions installation is managed through on disk configuration. 
+
+## Versioning
+
+OpenSearch has a version number following [semver](https://semver.org/).
+
+Plugins for OpenSearch must match their version exactly the version of OpenSearch.  Older version numbers are not compatible.
+
+Extensions version information is not tied to OpenSearch's version.
```
davidlago, line 67: Is this a security aspect?

SECURITY.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Configuration
+
+Configuration of OpenSearch uses on disk yml configuration files.  Other settings are manage in-memory through settings that are modifiable at runtime through APIs or indirectly.
```
davidlago, line 71: Do we also use OpenSearch indexes to store configuration elements too?

SECURITY.md
peternied: AFAIK in process extensions are not being designed or implemented at this time.  How do you feel about not including it until this effort is more concrete?

SECURITY.md
peternied: This is relevant in that CVEs for plugins are tightly coupled with one another and OpenSearch.  Whereas extensions and OpenSearch are be able to perform minor/patch revisions for CVEs without coordination.  I've update the section to better capture this.

SECURITY.md
peternied: Good point, I know the security plugin built functionality for this pattern, but I'm not sure if other plugins do the same.  I'll follow up on this.

SECURITY.md
```diff
@@ -0,0 +1,79 @@
...
+
+Additional terms:
+* **Plugin** - reference to the existing functionality to extend OpenSearch functionality.
+* **Extension** - reference to the in development functionality to extend OpenSearch.
+
+# Areas
+
+## Host security
+
+Plugins depend on use of the Java Security Manager is use to limit interactions on the host operation system resources (cpu/disk/memory/network/...).  JSM has been deprecated, with its removal scheduled in the next release of the JVM, see [OpenSearch discussion](https://github.com/opensearch-project/OpenSearch/issues/1687). Additional measures are needed to protect system resources.
```
dbwiddis, line 27:
```suggestion
Plugins depend on use of the Java Security Manager (JSM) to limit interactions on the host operation system resources (cpu/disk/memory/network/...).  JSM has been deprecated, with its removal scheduled in the next release of the JVM, see [OpenSearch discussion](https://github.com/opensearch-project/OpenSearch/issues/1687). Additional measures are needed to protect system resources.
```

SECURITY.md
dbwiddis: I think defining "REST API" is too specific for the potential application. The key point to make here is that interaction is limited to a defined set of API locations.  For external extensions, they will use REST.  If REST is not needed (e.g., in-process) there still should be another method of using the same API point.

SECURITY.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Communications security (COMSEC)
+
+Plugins are loaded into the same java virtual machine instance allowing communicate to OpenSearch through in process java APIs.  Plugins can issue REST API requests to the OpenSearch hosts reusing the standard node-to-node communications, internally called the transport client.
+
+Extensions of OpenSearch communicate via https requests between the nodes on the cluster and the extensions endpoint(s).  This is a bi-direction communication also allows extensions to contact the OpenSearch cluster through its available REST APIs.
+
+## Data Security
+
+OpenSearch stores data in memory and local file system storage.  This data is stored unencrypted.
```
dbwiddis, line 39: If it's unencrypted we should explain how it is presently "secured".

SECURITY.md
```diff
@@ -0,0 +1,79 @@
...
+
+Plugins can use the existing data systems of the OpenSearch.  Several classes of plugins extend storage options out to external services.
+
+### Access Control
+
+With the security plugin installed, role based access control (RBAC) is available with a policy document format specific to OpenSearch.  Access control over native OpenSearch data is possible with this plugin installed.
+
+For resource that are managed by plugins, access control is governed within individual plugin, by examining [user](https://github.com/opensearch-project/common-utils/blob/main/src/main/java/org/opensearch/commons/authuser/User.java) object from OpenSearch's thread context permissions are available for approval/denial. Example from anomaly detection, [checkUserPermissions](https://github.com/opensearch-project/anomaly-detection/blob/875b03c1c7596cb34d74fea285c28d949cfb0d19/src/main/java/org/opensearch/ad/util/ParseUtils.java#L568).  Uniform resource controls and models are needed to protect from misconfiguration and code defects.
```
dbwiddis, line 47:
```suggestion
For resource that are managed by plugins, access control is governed within individual plugin. By examining the [user](https://github.com/opensearch-project/common-utils/blob/main/src/main/java/org/opensearch/commons/authuser/User.java) object from OpenSearch's thread context, permissions are available for approval/denial. An example from anomaly detection is [checkUserPermissions](https://github.com/opensearch-project/anomaly-detection/blob/875b03c1c7596cb34d74fea285c28d949cfb0d19/src/main/java/org/opensearch/ad/util/ParseUtils.java#L568).  Uniform resource controls and models are needed to protect from misconfiguration and code defects.
```

dbwiddis:
I've been doing some thinking about the integration of security, extensions, and versioning and think they all inter-relate a bit.  Tossing the following diagram and content which you might consider including in the scope document in some way.  Presently an outline, but each of the below bullet points should be fleshed out.  @peternied @saratvemulapalli interested in your thoughts on this.

image

## API Controls

All of the below interactions will use the API/SDK interaction points as the primary means of enabling/enforcing the points below.
 - SDK will ensure Extensions identify themselves with a unique version which is permitted, Users authenticate appropriately, and requests are rejected if they would potentially harm the cluster

### Security & Identity

 - The OpenSearch cluster is protected from improper use of its resources
 - Only authorized users have accesess to data that they need in their current role (RBAC)

### Versioning

 - All components of the OpenSearch ecosystem (core, extensions) can be uniquely identified in a way that represents their features as well as code, dependencies, etc.

### Extensibility

 - Code external to OpenSearch can communicate with OpenSearch via a defined API/SDK

#### Security + Versioning

 - Individual versions of extensions can be declared unsecure (alternately, extensions must be "approved" to be allowed) due to vulnerabilities

#### Security + Extensibility

 - Extensions need a common framework to communicate user identities to the API, including certificate stores, identity server, etc.

#### Extensiblity + Versioning

 - Semantic Versioning and feature sets should combine to provide compatible version ranges

SECURITY.md
peternied: The host security is the only check on this data access; we could callout this overlap.  As it occurs to me, the data is not secured in a sufficient manor to justify the use of the term 'secure'.  

Services such as AWS' managed OpenSearch do have at rest encryption for data, and I wanted to make sure that readers of this doc did not conflate the two as I did when I first started writing this doc.

What do you think of leaving this text as it is?

SECURITY.md
peternied: If extensions are available in-process, it seems like we are recreating the existing plugin feature with its same architectural weaknesses.   With a solid security boundary between OpenSearch and its extensions, we can build extensibility in alignment with trustworthy functionality.  I think there is a discussion to be had for in-process extensions and I would prefer to wait to add those considerations to this document until we are committing more thought in this space, does this seem like a worthwhile trade-off in the near term?

I am only away of using REST APIs for extensions, if we want to broaden the language do you have a suggestion?

SECURITY.md
dbwiddis: As mentioned in earlier comment, we do need an individual feature list somewhere. We can limit this particular document to the overall framework/scope but identifying at least major categories of features is needed for any actual planning and scoping of future work

SECURITY.md
dbwiddis: Looked into this further and I'm fine with leaving this as REST API for simplicity.

saratvemulapalli (commented):
Thanks for writing this up @peternied.

Couple of more thoughts/questions:
- How do we protect inter-extension communication? 
- How can we enable extensions to expose resources, perform authentication and authorization?
- I see across the docs, the assumption is that the communication based on REST API's. So far extensions are just built on transport(TCP). For security, if REST APIs are preferred we could weigh in pros and cons.
- I see this doc scope as security for plugins and extensions. What are we thinking for Security for OpenSearch? 
- When we define scope, i believe it would be the time to define tenets we would want to stick to. 
- We did close: https://github.com/opensearch-project/security/issues/1895 in favor for this PR. I dont see the tactical items on the plate to execute for supporting AD as an extension. How do you want to take on this?

SECURITY.md
```diff
@@ -0,0 +1,57 @@
+# Extensions Security Guidelines
+
+OpenSearch's support for extensions allows for taking already powerful use cases and expanding on them, this creates a larger surface area for misuse, vunerabilities, and malicious interactions.  This document outlines several areas for enhancements, features, and practices to incorperate into extensions for OpenSearch
+
+To keep concepts consistant, this document is using terminology from [NIST Glossary](https://csrc.nist.gov/glossary).
```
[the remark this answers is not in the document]
peternied: It looks like this was resolved in a later revision, could you double check?

SECURITY.md
```diff
@@ -0,0 +1,79 @@
...
+
+Additional terms:
+* **Plugin** - reference to the existing functionality to extend OpenSearch functionality.
+* **Extension** - reference to the in development functionality to extend OpenSearch.
+
+# Areas
+
+## Host security
+
+Plugins depend on use of the Java Security Manager (JSM) to limit interactions on the host operation system resources (cpu/disk/memory/network/...).  JSM has been deprecated, with its removal scheduled in the next release of the JVM, see [OpenSearch discussion](https://github.com/opensearch-project/OpenSearch/issues/1687). Additional measures are needed to protect system resources.
```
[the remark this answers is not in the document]
peternied: Instead of going into this detail, I've linked the blog post.  As we plan to invest more energy into areas like this we can flesh them out better.

peternied:
Great questions @saratvemulapalli thanks for the feedback.

> How do we protect inter-extension communication?

This sounds like a future-feature request.  For the moment, proxied via OpenSearch seems like the path, but there is much potential in this space.  When we consider adding this functionality I think it warrants an in-depth discussion.  Should we file an issue on this now to track, or is there another document that captures this?  I'm leaning on having this come back up organically instead of creating an issue we forget about.

> How can we enable extensions to expose resources, perform authentication and authorization?

I'm not sure that we would implement all of these in this SDK.  For example in the identity space there is [Open ID](https://openid.net/what-is-openid/), I do not think that functionality would below in this SDK, as implementing the standard in a native plugin might be the better way.

> I see across the docs, the assumption is that the communication based on REST API's. So far extensions are just built on transport(TCP). For security, if REST APIs are preferred we could weigh in pros and cons.

If we want to be security focused with extensions, I think we need to strongly consider moving to only communicate via REST API.  While there are likely trade-offs from existing functionality/perf, there are huge benefits to quickly onboard simple extensions written in python, rust, or golang.  Additionally, by limited communicates with an ephemeral connections we can be sure that authorization is granted at the time of a given request.  It creates a well defined trust boundary.  I've created https://github.com/opensearch-project/opensearch-sdk/issues/34 for additional discussion, as well as updating this doc in the COMSEC section.

> I see this doc scope as security for plugins and extensions. What are we thinking for Security for OpenSearch?

This is a good callout, I'll start an effort to connect OpenSearch / Security as well as this doc together.

> When we define scope, i believe it would be the time to define tenets we would want to stick to.

I like this in principle, but I am struggling to see how we could roll this out.  I don't think we have existing engineering tenets in OpenSearch, this might be exactly the time to create them.  This said I think that those will be documented in the OpenSearch codebase and I'd like to see about merging this change without them, how does that occur to you?

> We did close: https://github.com/opensearch-project/security/issues/1895 in favor for this PR. I dont see the tactical items on the plate to execute for supporting AD as an extension. How do you want to take on this?

Thanks for bringing this up, after I merge this version of the document, there should be a fast follow for calling out the investigations and work that needs to be done to support Anomaly Detections feature set.  I'll reopen that issue and close it when this has been done.

dbwiddis (approved):

saratvemulapalli (approved):

dbwiddis:
> @dbwiddis @saratvemulapalli how does this look in its current form?

I'm happy with it in it's current form.  We may tweak a few things in the future but it's a good working document at this point.

Good questions regarding REST and auth/security.  I think it's a solvable problem, and all these are implementation details rather than scope which the doc outlines.
 - We can use SSL/TLS for the REST APIs to encrypt and then use (initially) the existing security mechanisms (user/password, certificate based) while also enabling (future) identity/RBAC capabilities.
   - May consider whether to make this user configurable, e.g., force https by default/allow insecure opt-in.
 - In-process comms can still use the REST API; localhost avoids any network traffic
 - Inter-process comms we'll need to figure out how to register availability, push vs. pull, robustness to failures, etc.

SECURITY.md
peternied: Lets revisit this in a follow up PR to make sure we've got the detail we need

Issue #27, opened 2022-06-23
[FEATURE] Add Java docs and mandate for all new classes


saratvemulapalli:
### Is your feature request related to a problem?
We are missing Java docs for this repository. 

### What solution would you like?
Lets add support for Java docs and mandate all new classes to have them. 
We could run this check as part of the PR check workflow.

dbwiddis:
+1. We are small enough right now to fix this without too much effort.

dbwiddis:
When I submitted a PR on the core repo yesterday the PR template included a check-off for Javadocs.  I'll add that to this repo; may not be an automated check but it will be something that submitters will see when they submit a PR, and should prompt code reviewers to do so as well.

Also can we be clear on the policy for Javadocs?  My thoughts:
 - Required for all classes
 - Required for all `public` and `protected` methods, _with the exception of getters and setters_ where the only javadoc would be "gets the foo"; "sets the foo".  Optional for private methods if the usage isn't obvious.
   - First line should be short and end in a period.
   - Separate paragraphs with `<p>`
   - `@param` required if any parameters. Explicitly state if null is allowed.
   - `@param` required for generics, e.g., `<T>`
   - `@return` required including what return values represent error conditions. Explicitly state if null is returned.
   - `@throws` required for runtime exceptions that aren't caught

dbwiddis:
I should have added to the above list. If a method overrides a method in a superclass, it automatically inherits that javadoc so it isn't required -- but can be added with the inheritdoc annotation if there's anything additional to say that the superclass/interface javadoc doesn't cover.

dbwiddis:
https://github.com/plume-lib/require-javadoc

peternied:
I would rather add dependencies on maintained components, I'd cast my vote with either `require-javadoc` or `checkstyle`.

If there are missing features in either tool, this would be a good reason to improve them with a proposal and pull request(s) to unblock adoption.

saratvemulapalli:
> @saratvemulapalli and @owaiskazi19 what are your thoughts?
I would always prefer relying on standard implementation vs something for us to maintain down the line. 
I would go with `require-javadoc`, also take our learnings to opensearch repository.

owaiskazi19:
> @saratvemulapalli and @owaiskazi19 what are your thoughts?

Discussed offline. +1 to @saratvemulapalli

dbwiddis:
[...]

PR #36, opened 2022-07-10
Copy PR template from main project

dbwiddis:

### Description
This PR copies over the PR template from the main OpenSearch project site, which includes a check list prompting the submitter to write tests and add javadocs, and prompts reviewers to check for the same.

### Issues Resolved
Part of a strategy to address #27.  Additional checks will be added in a later PR.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,15 @@
+### Description
+[Describe what this change achieves]
+ 
+### Issues Resolved
+[List any issues this PR will resolve]
+ 
+### Check List
+- [ ] New functionality includes testing.
+  - [ ] All tests pass
+- [ ] New functionality has been documented.
+  - [ ] New functionality has javadoc added
+- [ ] Commits are signed per the DCO using --signoff 
+
+By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
+For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).


.github/pull_request_template.md
```diff
@@ -0,0 +1,15 @@
...
+ 
+### Issues Resolved
+[List any issues this PR will resolve]
+ 
+### Check List
+- [ ] New functionality includes testing.
+  - [ ] All tests pass
+- [ ] New functionality has been documented.
+  - [ ] New functionality has javadoc added
+- [ ] Commits are signed per the DCO using --signoff
```
peternied, line 12: I don't think we should include this line item; its explicitly enforced by the PR DCO check enabled on all opensearch-project repositories and the statement at the bottom

.github/pull_request_template.md
```diff
@@ -0,0 +1,15 @@
+### Description
+[Describe what this change achieves]
+ 
+### Issues Resolved
+[List any issues this PR will resolve]
+ 
+### Check List
+- [ ] New functionality includes testing.
+  - [ ] All tests pass
```
peternied, line 9: Shouldn't this enforced by the PR check, can we remove this line if its already checked?

peternied (changes requested):
Doesn't look like this follows the template from [opensearch-project/.github](https://github.com/opensearch-project/.github/blob/main/PULL_REQUEST_TEMPLATE.md) is cleaner.  What do you think about using it instead?

dbwiddis:
> Doesn't look like this follows the template from [opensearch-project/.github](https://github.com/opensearch-project/.github/blob/main/PULL_REQUEST_TEMPLATE.md) is cleaner. What do you think about using it instead?

It follows that template and adds the checklist to it. It's currently active for PRs on OpenSearch site (where I copied it from and it takes priority over the organization-wide default):
https://github.com/opensearch-project/OpenSearch/blob/main/.github/pull_request_template.md

owaiskazi19:
> OK, so this is probably a bigger question/comment/concern.
> 
> We should have a standardized template across all `opensearch-project` repos. The template @dblock has here accomplishes this: https://github.com/opensearch-project/.github/blob/main/PULL_REQUEST_TEMPLATE.md

+1 to this.

> I don't think it's a good idea to have differing templates across the projects. @peternied raises some good points about whether we need to include items that are already part of the automated checks (tests pass; DCO signoff) and I'll be adding another PR to include javadoc checks as part of the gradle build.

Regarding the checklist
### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass.  -> This is covered by gradle check. We can remove this for OpenSearch and opensearch-sdk.
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added. -> Now since OpenSearch has strict javadoc rule as a part of gradle check. We can remove this as well for OpenSearch repo. SDK repo doesn't have the rule integrated yet.
- [ ] Commits are signed per the DCO using --signoff -> We have DCO bot for all the repos of opensearch-project. We can remove this too.

saratvemulapalli:
+1 to @peternied. Lets just use whats the standard for our project. 
@dbwiddis there are great questions here like redundant DCO, test pass etc. Probably a PR to `.github` upstream?

PR #43, opened 2022-07-11
Project plan for anomaly detection features

peternied:
### Description
Project plan for anomaly detection features

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -52,7 +52,7 @@ Extensions are sandboxed from the host system by operating via APIs.  This secur
 
 ## Communications security (COMSEC)
 
-Plugins are loaded into the same java virtual machine instance allowing communication to OpenSearch through in-process java APIs, although this behavior will be deprecated over time in favor of using a common API endpoint.  Plugins can issue API requests to the OpenSearch hosts reusing the standard node-to-node communications, internally called the transport client.
+Plugins are loaded into the same java virtual machine instance allowing communicate to OpenSearch through in process java APIs. Plugins can issue API requests to the OpenSearch hosts reusing the standard node-to-node communications, internally called the transport client.
 
 Extensions of OpenSearch communicate via https requests between the nodes on the cluster and the extensions endpoint(s).  This is a bi-directional communication and also allows extensions to contact the OpenSearch cluster through its available APIs.
 
@@ -112,7 +112,7 @@ To stretch out the design process while fulfilling scenarios some security effor
  `:white_check_mark:`.
 
 ## Anomaly Detection Plugin
-Overall project is tracked with [[FEATURE] Migrate Anomaly Detector plugin to work as an Extension](https://github.com/opensearch-project/opensearch-sdk/issues/24).  By migrating this plugin it will exercise the general extensions and security specific scenarios.
+Overall project is tracked with [[FEATURE] Migrate Anomaly Detector plugin to work as an Extension](https://github.com/opensearch-project/OpenSearch/issues/3635).  By migrating this plugin it will exercise the general extensions and security specific scenarios.
 
 Additional background avaliable from [Security#1895](https://github.com/opensearch-project/security/issues/1895)
 
@@ -123,7 +123,7 @@ Replace [commons.authuser.User](https://github.com/opensearch-project/common-uti
 When OpenSearch sends a request to an extension, the identity of the requestor should be included with the request. More discussion in [Handling identity in extensions](https://github.com/opensearch-project/opensearch-sdk/issues/14).
 
 #### Resource user/role checks [sdk#40](https://github.com/opensearch-project/opensearch-sdk/issues/40)
-anomaly Detection has detectors that analyzer data and store its results so it can be inspected or alerted on, [more details](https://opensearch.org/docs/latest/monitoring-plugins/ad/index/). OpenSearch should be responsible for inspecting the user, roles, resources to ensure standard practices are used.  A permissions check API should be designed and implemented to offload this work from extensions creators.
+anomaly Detection has detectors that analyzer data and store its results so it can be inspected or alerted on, [more details](https://opensearch.org/docs/latest/monitoring-plugins/ad/index/). OpenSearch should be responsible for inspecting the user, roles, resources to ensure standard practices are used.  An access check API should be designed and implemented to offload this work from extensions creators.
 
 ### Delegate Authority  [OpenSearch#3850](https://github.com/opensearch-project/OpenSearch/issues/3850) :negative_squared_cross_mark:
 anomaly Detection runs background jobs to scan for anamolies and alerts that trigger if conditions are detected.  Background tasks should be tied to an idenity and a delegated identity so permissions can be verified.  The underlying systems depends on the [Job Scheduler](https://github.com/opensearch-project/job-scheduler/blob/main/src/main/java/org/opensearch/jobscheduler/scheduler/JobScheduler.java) plugin to execute these requests. 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -52,11 +52,11 @@ Extensions are sandboxed from the host system by operating via APIs.  This secur
 
 ## Communications security (COMSEC)
 
-Plugins are loaded into the same java virtual machine instance allowing communicate to OpenSearch through in-process java APIs.  Plugins can issue API requests to the OpenSearch hosts reusing the standard node-to-node communications, internally called the transport client.
+Plugins are loaded into the same java virtual machine instance allowing communication to OpenSearch through in-process java APIs, although this behavior will be deprecated over time in favor of using a common API endpoint.  Plugins can issue API requests to the OpenSearch hosts reusing the standard node-to-node communications, internally called the transport client.
 
-Extensions of OpenSearch communicate via https requests between the nodes on the cluster and the extensions endpoint(s).  This is a bi-direction communication also allows extensions to contact the OpenSearch cluster through its available APIs.
+Extensions of OpenSearch communicate via https requests between the nodes on the cluster and the extensions endpoint(s).  This is a bi-directional communication and also allows extensions to contact the OpenSearch cluster through its available APIs.
 
-* :warning: The communication protocal has not been locked-in, following up with [Extensions to OpenSearch communication #34](https://github.com/opensearch-project/opensearch-sdk/issues/34).
+* :warning: The communication protocol has not been locked-in. Further discussion in [Extensions to OpenSearch communication #34](https://github.com/opensearch-project/opensearch-sdk/issues/34).
 
 ## Data Security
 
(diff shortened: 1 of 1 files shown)


peternied:
@opensearch-project/security Could I get a review of this set of changes as well?

SECURITY.md
```diff
@@ -30,10 +52,12 @@ Extensions are sandboxed from the host system by operating via APIs.  This secur
 
 ## Communications security (COMSEC)
 
-Plugins are loaded into the same java virtual machine instance allowing communicate to OpenSearch through in process java APIs.  Plugins can issue API requests to the OpenSearch hosts reusing the standard node-to-node communications, internally called the transport client.
+Plugins are loaded into the same java virtual machine instance allowing communicate to OpenSearch through in-process java APIs.  Plugins can issue API requests to the OpenSearch hosts reusing the standard node-to-node communications, internally called the transport client.
```
dbwiddis, line 55:
```suggestion
Plugins are loaded into the same java virtual machine instance allowing communication to OpenSearch through in-process java APIs, although this behavior will be deprecated over time in favor of using a common API endpoint.  Plugins can issue API requests to the OpenSearch hosts reusing the standard node-to-node communications, internally called the transport client.
```

SECURITY.md
```diff
@@ -30,10 +52,12 @@ Extensions are sandboxed from the host system by operating via APIs.  This secur
...
 
 Extensions of OpenSearch communicate via https requests between the nodes on the cluster and the extensions endpoint(s).  This is a bi-direction communication also allows extensions to contact the OpenSearch cluster through its available APIs.
```
dbwiddis, line 57:
```suggestion
Extensions of OpenSearch communicate via https requests between the nodes on the cluster and the extensions endpoint(s).  This is a bi-directional communication and also allows extensions to contact the OpenSearch cluster through its available APIs.
```

SECURITY.md
```diff
@@ -30,10 +52,12 @@ Extensions are sandboxed from the host system by operating via APIs.  This secur
...
 
+* :warning: The communication protocal has not been locked-in, following up with [Extensions to OpenSearch communication #34](https://github.com/opensearch-project/opensearch-sdk/issues/34).
```
dbwiddis, line 59:
```suggestion
* :warning: The communication protocol has not been locked-in. Further discussion in [Extensions to OpenSearch communication #34](https://github.com/opensearch-project/opensearch-sdk/issues/34).
```

SECURITY.md
```diff
@@ -76,4 +105,31 @@ Extensions configuration setup is tied to OpenSearch settings, extensions config
...
+ `:white_check_mark:`.
+
+## Anomaly Detection Plugin
+Overall project is tracked with [[FEATURE] Migrate Anomaly Detector plugin to work as an Extension](https://github.com/opensearch-project/opensearch-sdk/issues/24).  By migrating this plugin it will exercise the general extensions and security specific scenarios.
+
+Additional background avaliable from [Security#1895](https://github.com/opensearch-project/security/issues/1895)
+
+### User identity [OpenSearch#3846](https://github.com/opensearch-project/OpenSearch/issues/3846) :negative_squared_cross_mark:
+Replace [commons.authuser.User](https://github.com/opensearch-project/common-utils/blob/main/src/main/java/org/opensearch/commons/authuser/User.java) the common identity object used by plugins.  The new object should be obtainable by the extension through other means than [InjectSecurity](https://github.com/opensearch-project/common-utils/blob/main/src/main/java/org/opensearch/commons/InjectSecurity.java#L69) depends on the thread context.
```
dbwiddis, line 120:
```suggestion
Replace [commons.authuser.User](https://github.com/opensearch-project/common-utils/blob/main/src/main/java/org/opensearch/commons/authuser/User.java) which is presently the common identity object used by plugins.  The new object should conform to open identity standards and be obtainable by the extension through other means than [InjectSecurity](https://github.com/opensearch-project/common-utils/blob/main/src/main/java/org/opensearch/commons/InjectSecurity.java#L69) which depends on the thread context.
```

SECURITY.md
```diff
@@ -76,4 +105,31 @@ Extensions configuration setup is tied to OpenSearch settings, extensions config
...
+
+#### Aquiring User objects [sdk#37](https://github.com/opensearch-project/opensearch-sdk/issues/37)
+When OpenSearch sents a request to an extension the identiy should be avaliable, more discussion see [Handling identity in extensions](https://github.com/opensearch-project/opensearch-sdk/issues/14).
```
dbwiddis, line 123:
```suggestion
When OpenSearch sends a request to an extension, the identity of the requestor should be included with the request. More discussion in [Handling identity in extensions](https://github.com/opensearch-project/opensearch-sdk/issues/14).
```

SECURITY.md
```diff
@@ -76,4 +105,31 @@ Extensions configuration setup is tied to OpenSearch settings, extensions config
...
+
+#### Resource user/role checks [sdk#40](https://github.com/opensearch-project/opensearch-sdk/issues/40)
+anomaly Detection has detectors that analyzer data and store its results so it can be inspected or alerted on, [more details](https://opensearch.org/docs/latest/monitoring-plugins/ad/index/). OpenSearch should be responsible for inspecting the user, roles, resources to ensure standard practices are used.  A permissions check API should be designed and implemented to offload this work from extensions creators.
```
dbwiddis, line 126: Given we're touching on RBAC here with "user, roles, resources" I think the term "permissions check API" may be better termed "access check API" or at least "access permissions check API"... open to ideas.

SECURITY.md
```diff
@@ -76,4 +105,31 @@ Extensions configuration setup is tied to OpenSearch settings, extensions config
...
+anomaly Detection runs background jobs to scan for anamolies and alerts that trigger if conditions are detected.  Background tasks should be tied to an idenity and a delegated identity so permissions can be verified.  The underlying systems depends on the [Job Scheduler](https://github.com/opensearch-project/job-scheduler/blob/main/src/main/java/org/opensearch/jobscheduler/scheduler/JobScheduler.java) plugin to execute these requests. 
+
+#### Extension identity [sdk#41](https://github.com/opensearch-project/opensearch-sdk/issues/41)
+There should be different levels of permissions granularity interactive allowing for disgushing a user actions or user action through an extension.  Extensions should have an identity and there should be a way that the identity of action is layered with all the parties that have triggered it. 
+
+#### Delayed action API [sdk#42](https://github.com/opensearch-project/opensearch-sdk/issues/42)
+When actions are triggered without an interactive user session OpenSearch will need to permit the action to occur or not.  Create an API for these background tasks to get an identity associated with the session.
```
dbwiddis, line 135: Probably want to mention `Principal` identities here.

dbwiddis (dismissed):
This looks really good!  I've made a few suggested tweaks.

SECURITY.md
peternied: What do you think about moving this detail into the associated issue?  Its a great requirement that should be part of use cases and acceptance criteria.

SECURITY.md
peternied: > "access check API"

I like it

SECURITY.md
```diff
@@ -76,4 +105,31 @@ Extensions configuration setup is tied to OpenSearch settings, extensions config
...
-OpenSearch plugins can create cluster or node instability if incorrectly configured or by software defects.
\ No newline at end of file
+OpenSearch plugins can create cluster or node instability if incorrectly configured or by software defects.
+
+# Projects
+To stretch out the design process while fulfilling scenarios some security efforts will be tracked as longer running projects.  There will be tracking issues in github tied to the work, documentation here is justification for a project and how it ties into the security space.  Tasks will be denotated as incomplete with :negative_squared_cross_mark: `:negative_squared_cross_mark:` or completed with :white_check_mark:
+ `:white_check_mark:`.
+
+## Anomaly Detection Plugin
+Overall project is tracked with [[FEATURE] Migrate Anomaly Detector plugin to work as an Extension](https://github.com/opensearch-project/opensearch-sdk/issues/24).  By migrating this plugin it will exercise the general extensions and security specific scenarios.
```
owaiskazi19, line 115: We can mention the meta issue for AD plugin to run as an extension. The first part of which is to start with the create detector functionality : https://github.com/opensearch-project/OpenSearch/issues/3635

SECURITY.md
```diff
@@ -30,9 +52,11 @@ Extensions are sandboxed from the host system by operating via APIs.  This secur
 
 ## Communications security (COMSEC)
 
-Plugins are loaded into the same java virtual machine instance allowing communicate to OpenSearch through in process java APIs.  Plugins can issue API requests to the OpenSearch hosts reusing the standard node-to-node communications, internally called the transport client.
+Plugins are loaded into the same java virtual machine instance allowing communication to OpenSearch through in-process java APIs, although this behavior will be deprecated over time in favor of using a common API endpoint.  Plugins can issue API requests to the OpenSearch hosts reusing the standard node-to-node communications, internally called the transport client.
```
saratvemulapalli, line 55: Its lovely to see the new world, but we dont know what it looks like for plugins. 
I would rather avoid talking about deprecation at this point because Plugins still add a lot of value and Java APIs might still be supported.

SECURITY.md
```diff
@@ -30,9 +52,11 @@ Extensions are sandboxed from the host system by operating via APIs.  This secur
...
+
+Extensions of OpenSearch communicate via https requests between the nodes on the cluster and the extensions endpoint(s).  This is a bi-directional communication and also allows extensions to contact the OpenSearch cluster through its available APIs.
```
saratvemulapalli, line 57: ~So are we agreeing upon having HTTPs for extension <-> OpenSearch communication ? 
The design built today is via transport/TCP.~
I see the scary warning at bottom :)

SECURITY.md
```diff
@@ -30,9 +52,11 @@ Extensions are sandboxed from the host system by operating via APIs.  This secur
...
 
-Extensions of OpenSearch communicate via https requests between the nodes on the cluster and the extensions endpoint(s).  This is a bi-direction communication also allows extensions to contact the OpenSearch cluster through its available APIs.
+* :warning: The communication protocol has not been locked-in. Further discussion in [Extensions to OpenSearch communication #34](https://github.com/opensearch-project/opensearch-sdk/issues/34).
```
saratvemulapalli, line 59: 👍🏻

SECURITY.md
```diff
@@ -76,4 +105,31 @@ Extensions configuration setup is tied to OpenSearch settings, extensions config
...
+# Projects
+To stretch out the design process while fulfilling scenarios some security efforts will be tracked as longer running projects.  There will be tracking issues in github tied to the work, documentation here is justification for a project and how it ties into the security space.  Tasks will be denotated as incomplete with :negative_squared_cross_mark: `:negative_squared_cross_mark:` or completed with :white_check_mark:
+ `:white_check_mark:`.
+
+## Anomaly Detection Plugin
+Overall project is tracked with [[FEATURE] Migrate Anomaly Detector plugin to work as an Extension](https://github.com/opensearch-project/opensearch-sdk/issues/24).  By migrating this plugin it will exercise the general extensions and security specific scenarios.
+
+Additional background avaliable from [Security#1895](https://github.com/opensearch-project/security/issues/1895)
+
+### User identity [OpenSearch#3846](https://github.com/opensearch-project/OpenSearch/issues/3846) :negative_squared_cross_mark:
```
saratvemulapalli, line 119: AD exposes a bunch of actions. 
For example: `cluster:admin/opensearch/ad/detector/delete`[1] is a Transport Action API exposed by AD. 
How do we plan to support authorization of user for extensions?

[1] https://github.com/opensearch-project/anomaly-detection/blob/4baf75ca46da02c4e70016d34982f2c5a34d0acd/src/main/java/org/opensearch/ad/transport/DeleteAnomalyDetectorAction.java#L20

SECURITY.md
peternied: Thanks, I've updated the link to the more appropriate issue

SECURITY.md
peternied: Thanks removing the change to this paragraph.

SECURITY.md
peternied: With the existing security plugin in place, no additional action is needed as these are already filtered.  The assumption is that the security plugins hooks will still execute before the extensions hooks are triggered so the API calls can be intercepted as is already implemented.

If/When we rewrite the permissions evaluation system I think part of the backwards compatibility story will need to account for scenarios like this where no/minimal action is needed by plugin authors.

saratvemulapalli (approved):
Thanks @peternied for this.
Excited to see these pieces in action.

peternied:
@dbwiddis / @owaiskazi19 I'd like to get this change merged, any thing you'd like to see update beforehand?

owaiskazi19 (approved):

PR #44, opened 2022-07-11
Added tests for createComponent workflow and removed the logs of gradle test

owaiskazi19:
Signed-off-by: Owais Kazi <owaiskazi19@gmail.com>

### Description
Added tests for createComponent workflow, ExtensionsSettings and removed the logs of gradle test

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-sdk/issues/25 and closes https://github.com/opensearch-project/OpenSearch/issues/3082

### DIFF
--- a/src/main/java/org/opensearch/sdk/ExtensionSettings.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionSettings.java
@@ -16,9 +16,9 @@ package org.opensearch.sdk;
  */
 public class ExtensionSettings {
 
-    private String extensionname;
-    private String hostaddress;
-    private String hostport;
+    private String extensionName;
+    private String hostAddress;
+    private String hostPort;
 
     /**
      * Placeholder field. Change the location to extension.yml file of the extension.
@@ -32,21 +32,21 @@ public class ExtensionSettings {
         super();
     }
 
-    public String getExtensionname() {
-        return extensionname;
+    public String getExtensionName() {
+        return extensionName;
     }
 
-    public String getHostaddress() {
-        return hostaddress;
+    public String getHostAddress() {
+        return hostAddress;
     }
 
-    public String getHostport() {
-        return hostport;
+    public String getHostPort() {
+        return hostPort;
     }
 
     @Override
     public String toString() {
-        return "\nnodename: " + extensionname + "\nhostaddress: " + hostaddress + "\nhostPort: " + hostport + "\n";
+        return "\nnodename: " + extensionName + "\nhostaddress: " + hostAddress + "\nhostPort: " + hostPort + "\n";
     }
 
 }
--- a/src/main/java/org/opensearch/sdk/ExtensionSettings.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionSettings.java
@@ -25,28 +25,24 @@ public class ExtensionSettings {
      */
     public static final String EXTENSION_DESCRIPTOR = "src/test/resources/extension.yml";
 
-    public String getExtensionname() {
-        return extensionname;
+    public ExtensionSettings(String extensionname, String hostaddress, String hostport) {
+        this.extensionname = extensionname;
+        this.hostaddress = hostaddress;
+        this.hostport = hostport;
     }
 
-    public void setExtensionname(String extensionname) {
-        this.extensionname = extensionname;
+    public ExtensionSettings() {}
+
+    public String getExtensionname() {
+        return extensionname;
     }
 
     public String getHostaddress() {
         return hostaddress;
     }
 
-    public void setHostaddress(String hostaddress) {
-        this.hostaddress = hostaddress;
-    }
-
     public String getHostport() {
         return hostport;
-    }
-
-    public void setHostport(String hostport) {
-        this.hostport = hostport;
     }
 
     @Override
--- a/src/main/java/org/opensearch/sdk/ExtensionSettings.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionSettings.java
@@ -25,13 +25,12 @@ public class ExtensionSettings {
      */
     public static final String EXTENSION_DESCRIPTOR = "src/test/resources/extension.yml";
 
-    public ExtensionSettings(String extensionname, String hostaddress, String hostport) {
-        this.extensionname = extensionname;
-        this.hostaddress = hostaddress;
-        this.hostport = hostport;
+    /**
+     * Jackson requires a default constructor.
+     */
+    private ExtensionSettings() {
+        super();
     }
-
-    public ExtensionSettings() {}
 
     public String getExtensionname() {
         return extensionname;
--- a/src/test/java/org/opensearch/sdk/TestExtensionSettings.java
+++ b/src/test/java/org/opensearch/sdk/TestExtensionSettings.java
@@ -10,24 +10,21 @@ public class TestExtensionSettings extends OpenSearchTestCase {
 
     @BeforeEach
     public void setUp() {
-        extensionSettings = new ExtensionSettings();
+        extensionSettings = new ExtensionSettings("test-extension", "127.0.0.1", "1234");
     }
 
     @Test
     public void testExtensionName() {
-        extensionSettings.setExtensionname("test-extension");
         assertEquals(extensionSettings.getExtensionname(), "test-extension");
     }
 
     @Test
     public void testHostAddress() {
-        extensionSettings.setHostaddress("127.0.0.1");
         assertEquals(extensionSettings.getHostaddress(), "127.0.0.1");
     }
 
     @Test
     public void testHostPort() {
-        extensionSettings.setHostport("1234");
         assertEquals(extensionSettings.getHostport(), "1234");
     }
 }
--- a/src/test/java/org/opensearch/sdk/TestExtensionSettings.java
+++ b/src/test/java/org/opensearch/sdk/TestExtensionSettings.java
@@ -22,16 +22,16 @@ public class TestExtensionSettings extends OpenSearchTestCase {
 
     @Test
     public void testExtensionName() {
-        assertEquals(extensionSettings.getExtensionname(), "extension");
+        assertEquals(extensionSettings.getExtensionName(), "extension");
     }
 
     @Test
     public void testHostAddress() {
-        assertEquals(extensionSettings.getHostaddress(), "127.0.0.1");
+        assertEquals(extensionSettings.getHostAddress(), "127.0.0.1");
     }
 
     @Test
     public void testHostPort() {
-        assertEquals(extensionSettings.getHostport(), "4532");
+        assertEquals(extensionSettings.getHostPort(), "4532");
     }
 }
--- a/src/test/java/org/opensearch/sdk/TestExtensionSettings.java
+++ b/src/test/java/org/opensearch/sdk/TestExtensionSettings.java
@@ -1,0 +1,33 @@
+package org.opensearch.sdk;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.opensearch.test.OpenSearchTestCase;
+
+public class TestExtensionSettings extends OpenSearchTestCase {
+
+    private ExtensionSettings extensionSettings;
+
+    @BeforeEach
+    public void setUp() {
+        extensionSettings = new ExtensionSettings();
+    }
+
+    @Test
+    public void testExtensionName() {
+        extensionSettings.setExtensionname("test-extension");
+        assertEquals(extensionSettings.getExtensionname(), "test-extension");
+    }
+
+    @Test
+    public void testHostAddress() {
+        extensionSettings.setHostaddress("127.0.0.1");
+        assertEquals(extensionSettings.getHostaddress(), "127.0.0.1");
+    }
+
+    @Test
+    public void testHostPort() {
+        extensionSettings.setHostport("1234");
+        assertEquals(extensionSettings.getHostport(), "1234");
+    }
+}
--- a/src/test/java/org/opensearch/sdk/TestExtensionSettings.java
+++ b/src/test/java/org/opensearch/sdk/TestExtensionSettings.java
@@ -1,21 +1,28 @@
 package org.opensearch.sdk;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.opensearch.test.OpenSearchTestCase;
+
+import java.io.File;
+import java.io.IOException;
 
 public class TestExtensionSettings extends OpenSearchTestCase {
 
     private ExtensionSettings extensionSettings;
 
     @BeforeEach
-    public void setUp() {
-        extensionSettings = new ExtensionSettings("test-extension", "127.0.0.1", "1234");
+    public void setUp() throws IOException {
+        File file = new File(ExtensionSettings.EXTENSION_DESCRIPTOR);
+        ObjectMapper objectMapper = new ObjectMapper(new YAMLFactory());
+        extensionSettings = objectMapper.readValue(file, ExtensionSettings.class);
     }
 
     @Test
     public void testExtensionName() {
-        assertEquals(extensionSettings.getExtensionname(), "test-extension");
+        assertEquals(extensionSettings.getExtensionname(), "extension");
     }
 
     @Test
@@ -25,6 +32,6 @@ public class TestExtensionSettings extends OpenSearchTestCase {
 
     @Test
     public void testHostPort() {
-        assertEquals(extensionSettings.getHostport(), "1234");
+        assertEquals(extensionSettings.getHostport(), "4532");
     }
 }
--- a/src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
+++ b/src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
@@ -11,55 +11,124 @@
 
 package org.opensearch.sdk;
 
+import static java.util.Collections.emptySet;
+import static java.util.Collections.emptyMap;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.anyBoolean;
+import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.mock;
 
-import java.io.IOException;
+import java.net.InetAddress;
+import java.util.Collections;
+
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.mockito.Mockito;
+import org.opensearch.Version;
+import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.common.transport.TransportAddress;
+import org.opensearch.discovery.PluginRequest;
+import org.opensearch.discovery.PluginResponse;
 import org.opensearch.test.OpenSearchTestCase;
 import org.opensearch.transport.TransportService;
+import org.opensearch.transport.Transport;
 
 public class TestExtensionsRunner extends OpenSearchTestCase {
 
     private ExtensionsRunner extensionsRunner;
     private Settings settings;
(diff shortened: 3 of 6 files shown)


src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -11,55 +11,124 @@
...
+        verify(transportService, times(1)).start();
     }
 
     // test manager method invokes accept incoming requests on transport service
     @Test
-    public void testTransportServiceAcceptedIncomingRequests() throws IOException {
-
-        // retrieve and mock transport service
-        TransportService transportService = Mockito.spy(extensionsRunner.createTransportService(settings));
+    public void testTransportServiceAcceptedIncomingRequests() throws Exception {
```
dbwiddis, line 79: `startTransportService()` is not documented as throwing an exception. Should it be?  And similar to above comment should this be a more narrow exception, or is "throws" needed at all?

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -11,55 +11,124 @@
...
         // verify mocked object interaction in manager method
         extensionsRunner.startTransportService(transportService);
-        Mockito.verify(transportService, times(1)).acceptIncomingRequests();
+        verify(transportService, times(1)).acceptIncomingRequests();
     }
+
+    @Test
+    public void testRegisterRequestHandler() throws Exception {
+
+        extensionsRunner.startTransportService(transportService);
```
dbwiddis, line 89: Same comment here. Will an exception actually be thrown?

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -11,55 +11,124 @@
...
+        verify(transportService, times(3)).registerRequestHandler(anyString(), anyString(), anyBoolean(), anyBoolean(), any(), any());
+    }
+
+    @Test
+    public void testHandlePluginsRequest() throws Exception {
```
dbwiddis, line 94: I don't see where `handlePluginsRequest()` throws an exception?

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -11,55 +11,124 @@
...
+        PluginRequest pluginRequest = new PluginRequest(sourceNode, null);
+        PluginResponse response = extensionsRunner.handlePluginsRequest(pluginRequest);
+        assertEquals(response.getName(), "RealExtension");
+
+        // Test if the source node is set after handlePluginRequest() is called during OpenSearch bootstrap
+        assertEquals(extensionsRunner.getOpensearchNode(), sourceNode);
+    }
+
+    @Test
+    public void testClusterStateRequest() throws Exception {
```
dbwiddis, line 111: All exceptions are explicitly caught in `sendClusterStateRequest()` (which is a different problem!) so this ~will~ should never throw an Exception.

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -11,55 +11,124 @@
...
+
+        extensionsRunner.sendClusterStateRequest(transportService);
+
+        verify(transportService, times(1)).sendRequest(any(), anyString(), any(), any());
+    }
+
+    @Test
+    public void testClusterSettingRequest() throws Exception {
```
dbwiddis, line 119: `sendClusterSettingRequest()` catches all Exceptions (should it?) so how will it throw one here?

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -11,55 +11,124 @@
...
+
+        extensionsRunner.sendClusterSettingRequest(transportService);
+
+        verify(transportService, times(1)).sendRequest(any(), anyString(), any(), any());
+    }
+
+    @Test
+    public void testLocalNodeRequest() throws Exception {
```
dbwiddis, line 127: You know exactly what comment I'm going to make here so I won't repeat myself :)

dbwiddis (changes requested):
It seems we've added `throws Exception` to every test method whether it's needed or not.  If there's a good reason for this, I'd like to learn it!  But from my understanding:
 - Tests will fail on exception anyway (they are "caught" and "fail" in Junit)
 - In some cases there are no exceptions expected
 - In other cases we know the type of exception expected

src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -31,7 +31,7 @@ public String getHostaddress() {
         return hostaddress;
     }
 
-    public void getHostaddress(String hostaddress) {
+    public void setHostaddress(String hostaddress) {
```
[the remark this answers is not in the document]
owaiskazi19: Added for the class

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -11,55 +11,124 @@
...
+                null,
+                Collections.emptySet()
+            )
+        );
     }
 
     // test ExtensionsRunner getTransportService return type is transport service
     @Test
-    public void testGetTransportService() throws IOException {
+    public void testGetTransportService() throws Exception {
```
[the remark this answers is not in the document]
owaiskazi19: Removed IOException from `createTransoprtService()`. @joshpalis let me know if you think otherwise when you added it

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
owaiskazi19: We require an exception for `new TransportAddress(InetAddress.getByName("localhost"), 9876)` here and not for `handlePluginsRequest()`.

src/test/java/org/opensearch/sdk/TestExtensionSettings.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.opensearch.test.OpenSearchTestCase;
+
+public class TestExtensionSettings extends OpenSearchTestCase {
+
+    private ExtensionSettings extensionSettings;
+
+    @BeforeEach
```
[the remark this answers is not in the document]
peternied: It looks like `extensionSettings` is modified during the tests, this seems like it could lead to unpredictable behavior, I would recommend to reinitialize the value before every test run with the `@BeforeEach` annotation

saratvemulapalli (commented):
Most of the changes look good. 
Just a tiny comment. Thanks @owaiskazi19 for these changes!

src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -31,7 +34,7 @@ public String getHostaddress() {
...
```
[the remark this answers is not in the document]
peternied: It sounds ExtensionSettings should be immutable, and if that is the case maybe there could be a builder to create objects of a specific configuration for testing?

Note, for tests this can be worked around by creating a mock `ExtensionSettings` object that returns the expected values.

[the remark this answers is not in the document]
owaiskazi19: That's a good call out. Removed setters and found a way ObjectMapper can initialize in the constructor.

src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -19,30 +19,26 @@ public class ExtensionSettings {
     // Change the location to extension.yml file of the extension
     public static final String EXTENSION_DESCRIPTOR = "src/test/resources/extension.yml";
 
-    public String getExtensionname() {
-        return extensionname;
+    public ExtensionSettings(String extensionname, String hostaddress, String hostport) {
```
dbwiddis, line 22: I generally like the use of a constructor like this to set the fields, but this also implies these will always be the only fields we'll be using.  If we add one more we'd have to create a 4-arg constructor but because of SemVer we couldn't delete the 3-arg one; we could deprecate it but then we'd have to deal with what the appropriate default is etc.

All this is to say that this is only a good idea if we are absolutely sure these are the only 3 fields we will ever likely have in this class.

src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -19,30 +19,26 @@ public class ExtensionSettings {
...
+        this.extensionname = extensionname;
+        this.hostaddress = hostaddress;
+        this.hostport = hostport;
     }
 
-    public void setExtensionname(String extensionname) {
-        this.extensionname = extensionname;
+    public ExtensionSettings() {}
```
dbwiddis, line 28: Without any setters (since you've removed them) this constructor will leave the fields permanently uninitialized.

src/test/java/org/opensearch/sdk/TestExtensionSettings.java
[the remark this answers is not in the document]
dbwiddis: Well in the older behavior it was modified but only each individual field so they wouldn't step on each other.

And now it's an immutable object so we're creating an identical object three times.

dbwiddis (changes requested):

src/main/java/org/opensearch/sdk/ExtensionSettings.java
owaiskazi19: Actually, this is the way of initialization required by ObjectMapper to load yml file and the other was using setters but since we don't want to expose or modify host address and port, this is the approach I went for. Open to suggestions though.

src/main/java/org/opensearch/sdk/ExtensionSettings.java
owaiskazi19: Without a default constructor, Jackson will throw an exception

src/main/java/org/opensearch/sdk/ExtensionSettings.java
dbwiddis: Hard-coding inputs (as in constructors) is nearly always disconnected both logically and programmatically from reading in an arbitrary number of configurations key/value pairs via an external file. 

This is an interesting use case in that we want a POJO which is what we currently have but we also want to reserve the right to later generate a different POJO with more fields.

I took a look at ObjectMapper docs and it seems that it uses reflection and you can actually use private constructors for this purpose. That may help settle the SemVer argument as it is obvious non-API.  Alternately if there is some way we can mark this class as "internal" (a package with internal in the name is the common approach) then we also aren't bound by SemVer.

So the approach may be correct but the file location may not be. We need special treatment (package location) for these POJO objects.

src/main/java/org/opensearch/sdk/ExtensionSettings.java
dbwiddis: I believe you but I also suspect there's more to this story about how Jackson wants you to write classes.

dbwiddis (commented):
I think we need a bit more clarity on exactly what Jackson's expectations are for deserializing. I'm not sure what the correct implementation is, but I'm skeptical the current state is it.

src/main/java/org/opensearch/sdk/ExtensionSettings.java
owaiskazi19: As discussed. Removed the arg constructor and called super for no arg constructor

src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -25,30 +25,25 @@ public class ExtensionSettings {
      */
     public static final String EXTENSION_DESCRIPTOR = "src/test/resources/extension.yml";
 
-    public String getExtensionname() {
-        return extensionname;
+    /**
+     * Jackson requires a default constructor.
+     */
+    public ExtensionSettings() {
```
dbwiddis, line 31: You can make this constructor `private` and Jackson can still use it. This will avoid any confusion with users trying to instantiate the object (in fact, they won't be able to, as the presence of the private default one will prevent Java from adding its own default constructor).  May want to add "ObjectMapper" to the explanation in the javadoc as well, for clarity.

src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -25,30 +25,25 @@ public class ExtensionSettings {
...
+        super();
     }
 
-    public void setExtensionname(String extensionname) {
-        this.extensionname = extensionname;
+    public String getExtensionname() {
```
dbwiddis, line 35: (Style preference/suggestion) I know you're using the existing field names but I think using camelCase may improve readability for this one in particular because it's so long and has the double n's: `extensionName`.  Also consider for `hostAddress` and `hostPort`.

src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -25,30 +25,25 @@ public class ExtensionSettings {
      */
     public static final String EXTENSION_DESCRIPTOR = "src/test/resources/extension.yml";
```
dbwiddis, line 26: Just noticed that this path is in `src/test/resources` so this constant really shouldn't be under `src/main/java`.  Not sure we need to move it now as this is placeholder code, but where will we eventually be getting this?   

May want to make it package private at least.

dbwiddis (changes requested):

src/main/java/org/opensearch/sdk/ExtensionSettings.java
owaiskazi19: This is just a placeholder file. Once we integrate SDK with any extension, this will point to the extenions.yml of the extensions

src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -16,42 +16,37 @@
...
-    public void setExtensionname(String extensionname) {
-        this.extensionname = extensionname;
-    }
-
-    public String getHostaddress() {
-        return hostaddress;
+    /**
+     * Jackson requires a default constructor.
+     */
+    private ExtensionSettings() {
```
saratvemulapalli, line 31: How does `ObjectMapper` use this?

src/main/java/org/opensearch/sdk/ExtensionSettings.java
saratvemulapalli: Lets open an issue to track this problem. 
When we ship the SDK customers cannot use this, it should be a file path which is passed to `ExtensionsRunner`

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -11,55 +11,125 @@
...
         extensionsRunner.startTransportService(transportService);
-        Mockito.verify(transportService, times(1)).acceptIncomingRequests();
+        verify(transportService, times(1)).acceptIncomingRequests();
     }
+
+    @Test
+    public void testRegisterRequestHandler() {
+
+        extensionsRunner.startTransportService(transportService);
+        verify(transportService, times(3)).registerRequestHandler(anyString(), anyString(), anyBoolean(), anyBoolean(), any(), any());
```
saratvemulapalli, line 91: Instead of `any`, lets verify the respective handlers are being registered.

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -11,55 +11,125 @@
...
+        DiscoveryNode sourceNode = new DiscoveryNode(
+            "test_node",
+            new TransportAddress(InetAddress.getByName("localhost"), 9876),
+            emptyMap(),
+            emptySet(),
+            Version.CURRENT
+        );
+        PluginRequest pluginRequest = new PluginRequest(sourceNode, null);
+        PluginResponse response = extensionsRunner.handlePluginsRequest(pluginRequest);
+        assertEquals(response.getName(), "RealExtension");
```
saratvemulapalli, line 105: Not for this PR, but can you open up a follow up issue to respond back with the name we get from `extensions.yml` instead of static `RealExtension`

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -11,55 +11,125 @@
...
+        // Test if the source node is set after handlePluginRequest() is called during OpenSearch bootstrap
+        assertEquals(extensionsRunner.getOpensearchNode(), sourceNode);
+    }
+
+    @Test
+    public void testClusterStateRequest() {
+
+        extensionsRunner.sendClusterStateRequest(transportService);
+
+        verify(transportService, times(1)).sendRequest(any(), anyString(), any(), any());
```
saratvemulapalli, line 116: Same as above, lets actually verify the params are what we expect.

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -11,55 +11,125 @@
...
+    }
+
+    @Test
+    public void testClusterSettingRequest() {
+
+        extensionsRunner.sendClusterSettingsRequest(transportService);
+
+        verify(transportService, times(1)).sendRequest(any(), anyString(), any(), any());
```
saratvemulapalli, line 124: You know my comment :)

src/main/java/org/opensearch/sdk/ExtensionSettings.java
dbwiddis: > How does `ObjectMapper` use this?

Reflection!

owaiskazi19: It uses reflection to fill the attributes with the value present in the yml file

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
owaiskazi19: When we invoke `startTransportService` above it expects all the 3 handlers to be present at the same time. If we try to match it one by one, it causes MatchersException. I tried this before using `any`. I'm open to suggestions

src/main/java/org/opensearch/sdk/ExtensionSettings.java
dbwiddis: Which introduces its own problems with Java Modules, if we ever decide to publish a module descriptor we'll have to open this package to Jackson's module.  But for now, let's enjoy the classpath...

dbwiddis (dismissed):
LGTM with the note that we should track the YAML file location in another issue.

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
saratvemulapalli: Probably I didn't understand it. Why does it expect all 3 handlers at the same time?

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
owaiskazi19: Tried to match it with the Handler class

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
owaiskazi19: As [startTransportService](https://github.com/opensearch-project/opensearch-sdk/blob/main/src/main/java/org/opensearch/sdk/ExtensionsRunner.java#L217) is registering all 3 handlers

src/main/java/org/opensearch/sdk/ExtensionSettings.java
owaiskazi19: https://github.com/opensearch-project/opensearch-sdk/issues/52

dbwiddis (approved):

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
owaiskazi19: I covered it as a part of this PR

dbwiddis (approved):

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
saratvemulapalli: I am sure there is a way. I dont want to block this PR for this one. 
See if you find a solution.

saratvemulapalli (approved):
Thank you @owaiskazi19 for this PR!!

owaiskazi19:
Thanks @saratvemulapalli and @dbwiddis for the thorough review.

Issue #20, opened 2022-06-20
[FEATURE] Utilize the codebase required to create a detector from Anomaly Detector Plugin


owaiskazi19:
### Is your feature request related to a problem?
Since we are focusing on just creating the detector from AD, we can comment out the rest of the code and make sure AD works as usual. The build should be successful taking into tests also in consideration.

### What solution would you like?
_A clear and concise description of what you want to happen._

### What alternatives have you considered?
_A clear and concise description of any alternative solutions or features you've considered._

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

PR #67, opened 2022-07-29
Created SDK Client to interact with OpenSearch

owaiskazi19:
### Description
Created SDK Client to interact with OpenSearch for request calls required by the extensions. Once #20 is done, the next step would be to integrate SDK with the `feature/extensions` branch of AD to convert the AD plugin to extension. This client will help us to make the API calls from AD extension to OpenSearch.

Build will fail as https://github.com/opensearch-project/OpenSearch/pull/4197 should be merged first

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-sdk/issues/66


src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Creates client for SDK
+     * @throws IOException if client failed
+     */
+    public void createClient() throws IOException {
+        String endpoint = "localhost";
+        String username = "admin";
+        String password = "admin";
+        String protocol = "http";
+        int port = 9200;
```
owaiskazi19, line 46: Open to suggestion for a better way to store these attributes.

src/test/java/org/opensearch/sdk/TestSDKClient.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class TestSDKClient {
+    private OpenSearchTransport transport = new FailingTransport();
+
+    @Test
+    public void testCreateIndex() throws Exception {
+        OpenSearchClient client = new OpenSearchClient(transport);
+
+        //tag::builders
+        assertThrows(TransportException.class, () -> client.indices().create(
```
dbwiddis, line 28: I see a test here for failing, but where is the test that it works?

src/test/java/org/opensearch/sdk/FailingTransport.java
```diff
@@ -0,0 +1,112 @@
...
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
```
dbwiddis, line 12: Was this sourced from other code that had an Elasticsearch license header?  Probably want to add a short comment explaining why its header is different than everything else in this package, and probably want to put such code in its own package so it can be excluded from the spotless header rewrite.

dbwiddis (changes requested):
Need to decide what connection and authentication protocols we want to use so we can build this securely from the start.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,87 @@
...
+    private OpenSearchClient client;
+
+    /**
+     * Creates client for SDK
+     * @throws IOException if client failed
+     */
+    public void createClient() throws IOException {
+        String endpoint = "localhost";
+        String username = "admin";
+        String password = "admin";
```
saratvemulapalli, line 44: I understand this is we suggest to implement. 
To make this iterative, I would remove everything security for now and have a bare bones client i.e HTTP and everything insecure. 

Long term, depending on how security is going to fan out we could explore options:
- As @dbwiddis we might have token based access. 
- If we are looking for stop gap solution, I would get this information from our `.yml` config file, very likely load certificates and present them to OpenSearch.

@dbwiddis @peternied @owaiskazi19 thoughts?

src/main/java/org/opensearch/sdk/SDKClient.java
saratvemulapalli: I think the right way to do this is when OpenSearch sends an ExtensionsInitRequest, it should pass this information instead of hard coding in our configuration files. It makes it simple and easy to use. 

Using the configuration file, we have to configure OpenSearch with `extensions.yml`, and again configure a lot of params for the extension. We should be relying on transport as much as possible since the init communication happens.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,87 @@
...
+                restClient.close();
+            }
+        }
+    }
+
+    /**
+     * Creates index on OpenSearch
+     * @throws IOException if request failed
+     */
+    public CreateIndexResponse createIndex(String index) throws IOException {
```
saratvemulapalli, line 79: Sounds like this is the first one which is nice. 
Do you have a list of features we would like to expose to extensions? 

My thoughts, I would rather share the instantiation of client to the extension use. 
This will help us avoid writing these abstractions for now. Eventually we can make it better. 
WDYT?

src/test/java/org/opensearch/sdk/FailingTransport.java
saratvemulapalli: +1

src/main/java/org/opensearch/sdk/SDKClient.java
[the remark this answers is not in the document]
owaiskazi19: @peternied any thoughts here for authentication?

src/test/java/org/opensearch/sdk/FailingTransport.java
owaiskazi19: Yes, this is from the old ES code and that's why the header.

src/main/java/org/opensearch/sdk/SDKClient.java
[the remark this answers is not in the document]
peternied: The less instances of the word `password` in your codebase the better - how hard is it to remove?

BTW this is one of the reasons why I prefer connection strings to a resource, you _could_ imbed basic auth credentials via `https://username:password@localhost:9081/` and do not need to document this functionality anywhere

src/test/java/org/opensearch/sdk/FailingTransport.java
peternied: Can we avoid copying in like this an instead pull references from the OpenSearch core project?

src/main/java/org/opensearch/sdk/SDKClient.java
peternied: Kerberos is actual AuthN so its a great option - however I have never set it up with OpenSearch so I don't know how burdensome it will be to configure/use.

I would recommend using connection strings the work with AuthN/Z disabled (good for setting up quickly), then you can embed within the connection string basic auth credentials - this is the standard way to setup an OpenSearch cluster quickly (like pulling in a docker image)

src/test/java/org/opensearch/sdk/FailingTransport.java
owaiskazi19: This is from the [java-client](https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/test/java/org/opensearch/client/documentation/FailingTransport.java) and is not published to maven to consume.

build.gradle
```diff
@@ -46,6 +46,8 @@ dependencies {
     implementation "org.opensearch:opensearch:3.0.0-SNAPSHOT"
     implementation group: 'org.apache.logging.log4j', name: 'log4j-api', version: '2.17.1'
     implementation group: 'org.apache.logging.log4j', name: 'log4j-core', version: '2.17.1'
+    implementation 'org.opensearch.client:opensearch-rest-client:2.0.0'
+    implementation 'org.opensearch.client:opensearch-java:2.0.0'
```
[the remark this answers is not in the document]
owaiskazi19: Sure. SDK -> https://github.com/opensearch-project/opensearch-sdk/issues/89
Client -> https://github.com/opensearch-project/opensearch-java/issues/205

src/test/java/org/opensearch/sdk/TestSDKClient.java
owaiskazi19: We don't have a running OS to test when it works.

src/main/java/org/opensearch/sdk/SDKClient.java
owaiskazi19: Makes sense. I have removed the `createIndex` from the SDK and let the extension make the request.

src/main/java/org/opensearch/sdk/SDKClient.java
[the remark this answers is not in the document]
owaiskazi19: I have removed the security attributes and created an [issue](https://github.com/opensearch-project/opensearch-sdk/issues/90) to decide the right approach for adding security for SDKClient.

src/main/java/org/opensearch/sdk/SDKClient.java
owaiskazi19: Created https://github.com/opensearch-project/OpenSearch/pull/4197 to address the hostAddress and Port in the `ExtensionsInitRequest`.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,87 @@
...
+        String password = "admin";
+        String protocol = "http";
+        int port = 9200;
+        RestClient restClient = null;
+        try {
+            RestClientBuilder builder = RestClient.builder(new HttpHost(endpoint, port, protocol));
+            builder.setStrictDeprecationMode(true);
+            builder.setHttpClientConfigCallback(httpClientBuilder -> {
+                final CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
+                credentialsProvider.setCredentials(AuthScope.ANY, new UsernamePasswordCredentials(username, password));
```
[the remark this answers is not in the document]
owaiskazi19: Removed Credentials as a part of SDKClient without security.

src/main/java/org/opensearch/sdk/SDKClient.java
owaiskazi19: Sounds good. Removed security from the SDKClient and will be addressed [here](https://github.com/opensearch-project/opensearch-sdk/issues/90)

src/test/java/org/opensearch/sdk/FailingTransport.java
dbwiddis: > This is from the [java-client](https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/test/java/org/opensearch/client/documentation/FailingTransport.java) and is not published to maven to consume. Can we remove the ES header ...

The link you gave to the java-client includes an ES header, so the answer requires further research.

If the code you copied was part of the original fork from ES then no, you cannot remove the ES header.  ES maintains the copyright and licenses it to you.

If the code you copied was added under AL2.0 after the fork, then it is under the "modifications" clause of that copyright header and you could remove it.

Reviewing the history of this contribution it looks like it originated from this commit under ES license:
https://github.com/opensearch-project/opensearch-java/commit/4f3a394136652d5191eddb22aa6a944455fd7128

So you must retain the ES header.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -108,6 +110,8 @@ InitializeExtensionsResponse handleExtensionInitRequest(InitializeExtensionsRequ
         logger.info("Registering Extension Request received from OpenSearch");
         InitializeExtensionsResponse initializeExtensionsResponse = new InitializeExtensionsResponse(extensionSettings.getExtensionName());
         opensearchNode = extensionInitRequest.getSourceNode();
+        port = extensionInitRequest.getPort();
+        hostAddress = opensearchNode.getHostAddress();
```
dbwiddis, line 114: See similar comment on https://github.com/opensearch-project/OpenSearch/pull/4197 about this being redundant with a field in `opensearchNode`, so it doesn't need a separate field here.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,73 @@
...
+        OpenSearchTransport transport = new RestClientTransport(restClient, new JacksonJsonpMapper());
+        client = new OpenSearchClient(transport);
+        return client;
+    }
+
+    /**
+     *
+     * @throws IOException if closing the client fails
+     */
+    public void doClose() throws IOException {
```
dbwiddis, line 68: Tracing the call to `restClient.close()` it leads to `CloseableHttpAsyncClientBase` where any `IOException` is caught. So you can (and should) remove this `throws` clause so callers don't need to waste time handling an exception that won't occur.  (@saratvemulapalli please confirm this suggestion.)
https://jar-download.com/artifacts/org.apache.httpcomponents/httpasyncclient/4.1.4/source-code/org/apache/http/impl/nio/client/CloseableHttpAsyncClientBase.java

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
owaiskazi19: Keeping it separate here for better availability to extension developers as [this](https://github.com/opensearch-project/opensearch-sdk/pull/67/files#diff-da305de522ebd01a3b7821d9d3ee90d514a0b92d29a272fd5bb09bcb3a3e2c16R42) method requires `hostAddress` and `port`

src/main/java/org/opensearch/sdk/SDKClient.java
owaiskazi19: `restClient.close()` needs an exception otherwise it gives an error. If `IOException` doesn't look here we can change it to `Exception`.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public OpenSearchClient createClient(String hostAddress, int port) throws IOException {
+        RestClientBuilder builder = RestClient.builder(new HttpHost(hostAddress, port));
+        builder.setStrictDeprecationMode(true);
+        builder.setHttpClientConfigCallback(httpClientBuilder -> {
+            final CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
+            // credentialsProvider.setCredentials(AuthScope.ANY, new UsernamePasswordCredentials(username, password));
+            try {
+                return httpClientBuilder.setSSLHostnameVerifier(NoopHostnameVerifier.INSTANCE)
+                    .setSSLContext(SSLContextBuilder.create().loadTrustMaterial(null, (chains, authType) -> true).build());
```
saratvemulapalli, line 50: Do we really need to `setSSLContext`? Can we just have a simple httpClient?

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,73 @@
...
+     * @param hostAddress The address client can connect to
+     * @param port The port of the address
+     * @throws IOException if client failed
+     * @return SDKClient
+     */
+    public OpenSearchClient createClient(String hostAddress, int port) throws IOException {
+        RestClientBuilder builder = RestClient.builder(new HttpHost(hostAddress, port));
+        builder.setStrictDeprecationMode(true);
+        builder.setHttpClientConfigCallback(httpClientBuilder -> {
+            final CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
```
saratvemulapalli, line 46: Looks like `credentialsProvider` it's not being used. 
Probably remove it?

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Creates client for SDK
+     * @param hostAddress The address client can connect to
+     * @param port The port of the address
+     * @throws IOException if client failed
+     * @return SDKClient
+     */
+    public OpenSearchClient createClient(String hostAddress, int port) throws IOException {
+        RestClientBuilder builder = RestClient.builder(new HttpHost(hostAddress, port));
+        builder.setStrictDeprecationMode(true);
```
saratvemulapalli, line 44: Curious, what does this do?
Side note: I could only find elasticsearch documentation[1]. 
@VachaShah 

[1] https://javadoc.io/static/org.elasticsearch.client/elasticsearch-rest-client/6.8.3/org/elasticsearch/client/RestClientBuilder.html

src/test/java/org/opensearch/sdk/TestSDKClient.java
saratvemulapalli: Can we open an issue to add integration tests? 
We should be able to spin up opensearch nodes and make calls. 

@dbwiddis question makes a lot of sense, a.k.a How do we really know this client works other than manual testing?

src/main/java/org/opensearch/sdk/SDKClient.java
saratvemulapalli: Not sure how you've traced it @dbwiddis. 
I see this goes to:

https://github.com/opensearch-project/OpenSearch/blob/3a97d4cad65fa19126340aabf7e8bd2b1ce2b822/client/rest/src/main/java/org/opensearch/client/RestClient.java#L587

Which does throw IOException. Either we have to handle it or throw it back to the caller.

dbwiddis: Yes, `RestClient` has `throws` in the signature because the `client` in the call is based on an interface.  However, the specific implementation of that interface does not throw an exception. That was my original.

However, I retract my comment as we should not rely on implementation details and thus it is correct to treat `restClient.close()` as possibly throwing an exception.   (I had some doubt which was why I asked for a second opinion here).

src/test/java/org/opensearch/sdk/TestSDKClient.java
owaiskazi19: https://github.com/opensearch-project/opensearch-sdk-java/issues/97

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * This class creates a Client for SDK to make requests to OpenSearch
+ */
+public class SDKClient {
+    private final Logger logger = LogManager.getLogger(SDKClient.class);
+    private OpenSearchClient client;
+    private RestClient restClient = null;
+
+    /**
+     * Creates client for SDK
```
dbwiddis, line 33: Per other comments, the word "client" has at least 3 meanings in this class alone, and probably more in the larger OpenSearch context.  Make it more clear.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param hostAddress The address client can connect to
```
dbwiddis, line 34: Which address, the extension itself or opensearch? Make it clear.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param port The port of the address
+     * @throws IOException if client failed
+     * @return SDKClient
```
dbwiddis, line 37: What kind of client? You're returning an `OpenSearchClient` not an `SDKClient`. What am I supposed to do with it?  Am I responsible for closing it when I'm done with it?  Call that out here, e.g., "the user is responsible for calling `doClose()` when finished with it." (Except that method is on this class not the `OpenSearchClient` class, so I'm confused.)

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,67 @@
...
+import org.opensearch.client.RestClientBuilder;
+import org.opensearch.client.json.jackson.JacksonJsonpMapper;
+import org.opensearch.client.opensearch.OpenSearchClient;
+import org.opensearch.client.transport.OpenSearchTransport;
+import org.opensearch.client.transport.rest_client.RestClientTransport;
+
+/**
+ * This class creates a Client for SDK to make requests to OpenSearch
+ */
+public class SDKClient {
```
dbwiddis, line 27: Should this be extending some other class, like `OpenSearchClient`?

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,67 @@
...
+
+        // Create Client
+        OpenSearchTransport transport = new RestClientTransport(restClient, new JacksonJsonpMapper());
+        client = new OpenSearchClient(transport);
+        return client;
+    }
+
+    /**
+     *
+     * @throws IOException if closing the client fails
```
dbwiddis, line 60: Which client, the OpenSearchClient (`client`) or the RestClient (`restClient`)?  Do we need to close them both?

src/test/java/org/opensearch/sdk/FailingTransport.java
dbwiddis, line 12: suggest putting these files requiring the ES license header in their own package so they are easily excluded from our license header checker (similar to how we used to exclude the Netty4 code).

src/test/java/org/opensearch/sdk/TestSDKClient.java
```diff
@@ -0,0 +1,39 @@
...
+import org.opensearch.client.transport.TransportException;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class TestSDKClient {
+    private OpenSearchTransport transport = new FailingTransport();
+
+    @Test
+    public void testCreateIndexException() throws Exception {
+        OpenSearchClient client = new OpenSearchClient(transport);
```
dbwiddis, line 25: This isn't testing SDKClient at all, it's creating an `OpenSearchClient` from scratch.  Should this be `SDKClient.createClient()`?

dbwiddis: We aren't testing `doClose()`.

dbwiddis (changes requested):

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -143,6 +145,8 @@ InitializeExtensionsResponse handleExtensionInitRequest(InitializeExtensionsRequ
             return initializeExtensionsResponse;
         } finally {
             // After sending successful response to initialization, send the REST API
+            port = extensionInitRequest.getPort();
```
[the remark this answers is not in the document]
owaiskazi19: We don't need this [PR](https://github.com/opensearch-project/OpenSearch/pull/4197) then.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -143,6 +145,8 @@ InitializeExtensionsResponse handleExtensionInitRequest(InitializeExtensionsRequ
...
+            hostAddress = opensearchNode.getHostAddress();
```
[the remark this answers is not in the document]
owaiskazi19: We don't need this [PR](https://github.com/opensearch-project/OpenSearch/pull/4197) then.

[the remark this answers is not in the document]
dbwiddis: > We don't need this [PR](https://github.com/opensearch-project/OpenSearch/pull/4197) then.

Yup.  https://github.com/opensearch-project/OpenSearch/pull/4277

Sorry I didn't notice that earlier.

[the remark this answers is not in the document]
dbwiddis: (Actually, `opensearchNode.getHostAddress()` is fine for this, it does exactly what I suggested. But the comment above about the port doesn't have that delegate getter.)

src/main/java/org/opensearch/sdk/SDKClient.java
owaiskazi19: Based on the [sample](https://opensearch.org/docs/latest/clients/java-rest-high-level/#sample-code), doesn't look like we should extend it.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,67 @@
...
+    private RestClient restClient = null;
+
+    /**
+     * Creates client for SDK
+     * @param hostAddress The address client can connect to
+     * @param port The port of the address
+     * @throws IOException if client failed
+     * @return SDKClient
+     */
+    public OpenSearchClient createClient(String hostAddress, int port) throws IOException {
```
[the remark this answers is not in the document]
dbwiddis: You are creating (and returning) an `OpenSearchClient` though.  And also a `RestClient`, internally.   It's fine if we say we are creating an `SDKClient` which wraps these other ones, but then why have a create method instead of just a constructor which wraps this?  

I'd be fine with `initializeClient()` if it was documented to say the other two clients being internally created.  But I don't understand the return type.

src/test/java/org/opensearch/sdk/FailingTransport.java
owaiskazi19: As changed the test. This file is not needed anymore :)

src/test/java/org/opensearch/sdk/TestSDKClient.java
owaiskazi19: Made the change.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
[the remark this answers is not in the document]
owaiskazi19: Thanks for taking care of the revert PR. Let's just use `port` and `hostAddress` from the `openSearchNode`.

src/main/java/org/opensearch/sdk/SDKClient.java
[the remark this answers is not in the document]
owaiskazi19: Let's not put it in the constructor and make it heavy. Changed the method name to `initializeClient()`.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,67 @@
...
+import org.opensearch.client.opensearch.OpenSearchClient;
+import org.opensearch.client.transport.OpenSearchTransport;
+import org.opensearch.client.transport.rest_client.RestClientTransport;
+
+/**
+ * This class creates SDKClient for an extension to make requests to OpenSearch
+ */
+public class SDKClient {
+    private final Logger logger = LogManager.getLogger(SDKClient.class);
+    private OpenSearchClient client;
```
dbwiddis, line 29: Took me a bit to wrap my mind around the name collision. This is an interface in "server" but a standalone class in "java-client".  To help future-me, suggest changing this variable name to be "javaClient" or at least adding a comment.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -0,0 +1,67 @@
...
+    private RestClient restClient = null;
+
+    /**
+     * Creates OpenSearchClient for SDK. It also creates a restClient as a wrapper around OpenSearchClient
```
dbwiddis, line 33: Similarly here make this clear which OpenSearchClient we're wrapping.

dbwiddis (approved):
LGTM.  I'm still a bit confused on the client naming stuff, but that can be clarified in later PRs.

dbwiddis (approved):

saratvemulapalli:
@dblock thats a good question. 
SDK uses RestAPIs to talk to OpenSearch instead of using traditional internal client built by OpenSearch and handed over to plugins[1].

Do you think the java client would ever need to use the SDK? 

[1] https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/plugins/Plugin.java#L141

dblock:
I can't come up with an example, I came from an angle of wanting the SDK to have its dependencies to a minimum, but I suppose until I have a good reason things should remain as is ;)

Issue #69, opened 2022-07-30
[FEATURE] Register REST API and map to associated extension in ExtensionsOrchestrator


dbwiddis:
### Is your feature request related to a problem?
See steps 4, 6, and 7 described in #64.

### What solution would you like?
Upon receipt of REST API listing provided in #68, The `ExtensionsOrchestrator` should:
 * Register the APIs with the `RestHandler`
 * Store the response from the RestHandler in a map with the API as the key and the requesting extension as the value.

### What alternatives have you considered?
The best location for the map and API redirects to be stored is still an open question.  

### Do you have any additional context?
See also the DESIGN.md file (draft in #60)

dbwiddis:
I think we might be able to re-use the vast majority of existing Plugin code for the REST actions.  However, we need to carefully address the question of "which JVM is the code executed in".   Directly applying the code may end up with OpenSearch's JVM processing the actions, so the registered "actions" need to prompt network communication to a separate request and associated handler.

Details below.

-----

Current Plugin setup uses a base URI containing `_plugins`:
```java
public static final String AD_BASE_URI = "/_plugins/_anomaly_detection";
```
So we can follow this model and use an `_extensions` base URI:
```java
String restApiBaseString = "/_extensions/_" + extensionNameGoesHere;
String restActionStringToRegister = restApiBaseString + restActionString;
```
Only concern here might be conflicting/duplicate extension names so we may consider using a uniqueID, or an expanded "fully qualified name" (e.g., similar to JPMS module names which are supposed to be globally unique) to deconflict.  I'll go ahead with the extension name for now.

-----

REST Handlers correspond to `Action` classes which extend `BaseRestHandler`.  For the AD plugin, these are returned by this method:
```java
public List<RestHandler> getRestHandlers( ... ) {
  // ...
  RestGetAnomalyDetectorAction restGetAnomalyDetectorAction = new RestGetAnomalyDetectorAction();
  // ...
  return ImmutableList.of(
    restGetAnomalyDetectorAction,
  // ...
  )
}
}
```

These are eventually registered with OpenSearch (the end goal of this issue) in the `ActionModule` class
```java
for (ActionPlugin plugin : actionPlugins) {
    for (RestHandler handler : plugin.getRestHandlers(
        settings,
        restController,
        clusterSettings,
        indexScopedSettings,
        settingsFilter,
        indexNameExpressionResolver,
        nodesInCluster
    )) {
        registerHandler.accept(handler);
    }
[...]

Issue #64, opened 2022-07-28
[FEATURE] [META] Register RestHandlers in OpenSearch


owaiskazi19:
### Is your feature request related to a problem?
Currently, extensions are initialized in OpenSearch through extensions.yml file. For OpenSearch to process any functionality of the extension be it creating an index, deleting it or any other extension's functionality, the REST APIs of the extension should be registered in OpenSearch (ExtensionsOrchestrator) for OpenSearch to understand which extension it should send request to after receiving it from the user. The RestHandler should do the job of registering those APIs and map it in the sample format mentioned below
```
{extensionNode1: [API1, API2, API3], extenionNode2: [API1, API4, API5]}
```
extensionNode will be the node of the extension and values will be the APIs of the extension.

This issue requires us to
1. Register extension REST APIs to ExtensionOrchestrator by building Rest Handler model [3][4]
2. Map the extension node with the APIs it has  to register [6] [7]
3. Process the API call sent by the user and send a transport request to the right extension which has the request registered for that API [5] [8]

![Rest_new](https://user-images.githubusercontent.com/16099877/181392993-c9a82238-83a2-42d4-99f9-f26afaae657b.png)

### What solution would you like?
_A clear and concise description of what you want to happen._

### What alternatives have you considered?
_A clear and concise description of any alternative solutions or features you've considered._

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

dbwiddis:
Some quick questions:
 - You have "OpenSearch SDK" labeled on the Extension, but my understanding is that the SDK is what will define how the extension and OpenSearch communicate with each other.  So my thought is the SDK would be more appropriate either on the ExtensionsOrchestrator (as the "receiving end of the arrows coming from the extensions", with the "SDK Client" fulfilling a similar role for the Extension), or as some sort of dotted-line box around all the back-and-forth communication (1, 2, 3, 8, and 9) indicating the common language that both sides need to understand. 
 - I'm not sure how the user interaction (5) falls into the sequence.  Should they be triggering the bootstrap request (1) with all the index information needed, or does the index information come from the extension, or both?

dbwiddis:
I've created separate issues to track this.
 - [ ] #68 
 - [ ] #69 
 - [ ] #70

dbwiddis:
> {extensionNode1: [API1, API2, API3], extenionNode2: [API1, API4, API5]}

@owaiskazi19 two questions on this mapping:
[...]

Issue #52, opened 2022-07-15
[FEATURE] Pass extensions.yml file to ExtensionsRunner


owaiskazi19:
### Is your feature request related to a problem?
Currently, SDK relies on a test extenions.yml file to get the metadata of the extensions. Create a workflow to read the extensions.yml of the extensions and pass it to ExtensionSettings.java

### What solution would you like?
_A clear and concise description of what you want to happen._

### What alternatives have you considered?
_A clear and concise description of any alternative solutions or features you've considered._

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

Once the location of extension.yml is finalized we have to update Developer guide.

owaiskazi19:
We will be passing one single object of the extension to SDK from OpenSearch. It will be covered here: https://github.com/opensearch-project/opensearch-sdk-java/issues/93. Closing tis in favor of that.

PR #74, opened 2022-08-02
Provide Extension API to OpenSearch

### LINKED ISSUE #68: [FEATURE] Send required REST APIs to OpenSearch
See step 3 described in #64

The initial connection from `ExtensionsOrchestrator` (in OpenSearch) to `ExtensionsRunner` (in each extension) is handled by this method:
https://github.com/opensearch-project/opensearch-sdk/blob/f9b3157af6b6b8cf05d4fb34ce6a8f529446a55f/src/main/java/org/opensearch/sdk/ExtensionsRunner.java#L100-L106

Another option is for `ExtensionsOrchestrator` to send a second request to the Extension after the initial acknowledgement, containing a different pluginRequest and then receive the list of APIs.  

Given the basic implementation at present, this seems like an unnecessary extra step, somewhat like pinging someone on chat and asking "Can I ask a question?"  Rather than simply returning the extension name, this handler could return a more detailed response including the REST APIs that the Extension will require OpenSearch to handle.

However, having two handshakes provides a bit more flexibility in the future such as implementing #52 and sending the extensions.yaml information in the second interchange.

See also the DESIGN.md file (draft in #60)

dbwiddis:

### Description
Companion PR: https://github.com/opensearch-project/OpenSearch/pull/4100

Extensions will report their APIs to the ExtensionsOrchestrator upon initial registration and potentially later in the Extension's lifecycle. These APIs will be registered and mapped so that when Users submit API requests they can be forwarded to the appropriate extensions.

This PR handles the communication of the REST API from the Extension to OpenSearch.   Future PRs will complete the process of registering the REST Handlers and redirecting inbound requests.
 
### Issues Resolved
Fixes #68


dbwiddis:
On extension orchestrator log:
> [2022-08-04T16:20:43,929][INFO ][o.o.e.ExtensionsOrchestrator] [dev-dsk-widdisd-2a-8f5e7316.us-west-2.amazon.com] Initialized extension: sample-extension
> [2022-08-04T16:20:43,930][INFO ][o.o.e.ExtensionsOrchestrator] [dev-dsk-widdisd-2a-8f5e7316.us-west-2.amazon.com] Registering: GET /_extensions/sample-extension/api_1
> [2022-08-04T16:20:43,930][INFO ][o.o.e.ExtensionsOrchestrator] [dev-dsk-widdisd-2a-8f5e7316.us-west-2.amazon.com] Registering: PUT /_extensions/sample-extension/api_2
> [2022-08-04T16:20:43,930][INFO ][o.o.e.ExtensionsOrchestrator] [dev-dsk-widdisd-2a-8f5e7316.us-west-2.amazon.com] Registering: POST /_extensions/sample-extension/api_3

On extension runner log:
> MESSAGE RECEIVED:E-ǣpRegistered node opensearch-sdk-1, extension sample-extension to handle API [GET /api_1, PUT /api_2, POST /api_3]

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -76,33 +74,67 @@ public ExtensionsRunner() throws IOException {}
...
         opensearchNode = pluginRequest.getSourceNode();
+        // Fetch the unique ID
+        for (DiscoveryExtension de : pluginRequest.getExtensions()) {
+            if (de.getName().equals(extensionSettings.getExtensionName())) {
+                setUniqueId(de.getId());
+                break;
+            }
+        }
+        if (getUniqueId() == null) {
+            throw new IllegalArgumentException(
```
joshpalis, line 130: javadoc should include an `@throws` since this throws an IllegalArguementException

joshpalis (approved):
LGTM! Just a small suggestion regarding missing javadocs

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
dbwiddis: @joshpalis Technically speaking, the `@throws` would only be required if the method itself declared `throws IllegalArgumentException`.  Which isn't required as it's an unchecked exception.  It only occurs in the case there's a misconfiguration where the extension name in `extensions.yml` doesn't match the extension name defined by the extension.

Which I might argue is an unnecessary requirement and one of the two names should be ignored.  However by passing the entire list of `DiscoveryExtension`s matching the names is the only reliable way to fetch the unique ID.  (Maybe we should have the extension itself delcare the uniqueId?)

I think the correct thing to do is not bother with the name check, but change `PluginRequest` (which should be renamed) to pass a single instance of the correct `DiscoveryExtension` instead of the list.   Since you'll be updating your NamedWritables PR based on this code you might consider altering that in the process.  (FYI @ryanbogan who has also been involved with this code.)

dbwiddis: > pass a single instance of the correct `DiscoveryExtension` instead of the list.

This implementation could probably accomplish the goals of #52.

joshpalis (dismissed):

src/main/java/org/opensearch/sdk/ExtensionApi.java
```diff
@@ -0,0 +1,40 @@
...
+ * This class encapsulates the API of an Extension.
+ */
+public class ExtensionApi {
+
+    private List<String> api = new ArrayList<>();
+
+    /**
+     * Placeholder field. Change the location to extension_api.yml file of the extension.
+     */
+    public static final String EXTENSION_API_DESCRIPTOR = "src/test/resources/extension_api.yml";
```
saratvemulapalli, line 23: This is nice. Can we generate these RestAPI configuration from a Spec[1]? 
Just something for the future. 
We'd like to have extensions defining RestAPIs and having our clients to support it. 
cc: @VachaShah

[1] https://github.com/opensearch-project/OpenSearch/issues/3090

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -83,18 +91,38 @@ public class ExtensionsRunner {
...
-        ExtensionSettings extensionSettings = objectMapper.readValue(file, ExtensionSettings.class);
-        return extensionSettings;
+        return objectMapper.readValue(file, ExtensionSettings.class);
+    }
+
+    private ExtensionApi readExtensionApi() throws IOException {
+        File file = new File(ExtensionApi.EXTENSION_API_DESCRIPTOR);
+        ObjectMapper objectMapper = new ObjectMapper(new YAMLFactory());
+        return objectMapper.readValue(file, ExtensionApi.class);
+    }
```
saratvemulapalli, line 111: Do you think its worth having its own class? This helps ExtensionsRunner to be simple and lean (and testable).

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -108,7 +136,21 @@ InitializeExtensionsResponse handleExtensionInitRequest(InitializeExtensionsRequ
...
+            }
+        }
+        if (getUniqueId() == null) {
+            throw new IllegalArgumentException(
+                "Extension " + extensionSettings.getExtensionName() + " does not match any requested extension."
+            );
+        }
         setOpensearchNode(opensearchNode);
+        extensionTransportService.connectToNode(opensearchNode);
+        sendRegisterApiRequest(extensionTransportService);
```
saratvemulapalli, line 153: We are trying to send RegisterApi before responding to the init extension request. 
Wouldn't this mess up the workflow on ExtensionsOrchestrator? 
Can we call the API after the registration is successful ? 

I am worried if the response is not sent cleanly and something fails, we might have to build fail-safe code on ExtensionOrchestrator. Instead, I think its simpler to have the API call to register them after the extension registration is successful. WDYT?

src/main/java/org/opensearch/sdk/ExtensionApi.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package org.opensearch.sdk;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This class encapsulates the API of an Extension.
+ */
+public class ExtensionApi {
```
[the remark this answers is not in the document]
dbwiddis: @saratvemulapalli Do we have a style that would prefer `RESTApi` or `RESTAPI` or `RestApi`?

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
dbwiddis: Good question.  I agree that moving this (and corresponding tests) out of `ExtensionsRunner` is a good idea, but  I'm not sure it's worth a whole class just for this. There are some other options:
 - Move this to the `ExtensionApi` class
 - Create a new `YAMLReader` class and move both this and other similar methods into it

@owaiskazi19 any thoughts on this?

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
dbwiddis: Excellent question, and I went back and forth on this.  I can easily switch the register request into a `finally` block to execute after the response.  I'll update again to do this but I think the question about how to handle failures needs further discussion and would welcome @owaiskazi19 and @joshpalis input on this.  A similar question should be asked about registering the named writ(e)ables.

One underlying thought I had:  the extension isn't properly initialized until both the REST API and the named writables are registered.  So we do have to handle failures of those methods in any case.

You saw in my initial draft PR that I simply replaced the original response with the API, eliminating the sequencing issue, but that doesn't allow for the API to be updated later during operation.

src/main/java/org/opensearch/sdk/ExtensionApi.java
dbwiddis: +1 on spec-based OpenAPI-compatible generation.

dbwiddis: See also the comments at the header of that file:
```
# This is a placeholder for a more formal API definition file, e.g., an OpenAPI
# spec file which can be used to generate documentation and streamline testing
# See https://github.com/opensearch-project/OpenSearch/issues/3090
# The API will include at least two components, a method (e.g., GET, PUT) and
# parts of a URI (e.g., /_extensions/sample-extension/api_1)
```

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
owaiskazi19: Option 1 to move it to `ExtensionApi` class looks good to me here if we don't have any similar methods reading YAML file. (AFAIK this is the only one)

dbwiddis: On further review as I encountered this -- I actually think this is a temporary method given the eventual intent to get the API from a spec file.  So both this and the `readExtensionSettings()` are placeholders for better non-YAML functionality so I think they are fine to leave as-is.

src/main/java/org/opensearch/sdk/ExtensionApi.java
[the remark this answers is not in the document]
dbwiddis: I went with `RestApi` because while a standalone `RestAPI` might work it's also used in longer classes like `RestApiResponseHandler` and I strongly prefer the lowercase there.   But @joshpalis has used `NamedWriteableRegistryAPI` so we should align on a style before we create any more APIs.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
dbwiddis: I'm going to create a new issue to raise the exception handling question in the previous comment.  For now I've reordered the responses to align with the design diagrams on #64 and elsewhere.

src/main/java/org/opensearch/sdk/ExtensionApi.java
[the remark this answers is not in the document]
peternied: Within the OpenSearch codebase, the moniker is `Rest`, as in `RestController.java`

Names / suffixes that I do not recommend using, `Framework`, `Utility`, `Helper`, and `API` they tend not to impart any additional information about the what or why of the class.  

As I see the utility of this class, it seems like its a registry of extension specific actions/operations.  Maybe something along the lines of `RegisteredActions` or `AvailableOperations` - What do you think?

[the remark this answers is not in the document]
dbwiddis: Great observations, @peternied.  I agree that for many of these we can get rid of the API and I'll work on that later today.  My thoughts (mostly notes to myself for later):
 - I think I'll leave the `ExtensionRestApi` class named as-is for now, as it's really a placeholder class for reading a YAML file, both of which will eventually go away when we eventually implement an OpenAPI-compatible spec file.  And given the strong intended tie to "OpenAPI" I'd like to keep the API (or Api) in the name for now.
 - The remainder of the operations are going to be "actions" that need to be "handled", e.g, `public class RestDeleteAnomalyDetectorAction extends BaseRestHandler`.   I think either `FooExtensionRestAction` or `FooExtensionRestHandler` or the similar for ...Request and ...Response will be a better convention.

FYI @joshpalis I think you should consider removing API from the NamedWriteables code for similar reasons.

FYI @owaiskazi19 and @saratvemulapalli when you hopefully review this PR tomorrow.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
dbwiddis: And on even further review, I moved this "temporary" code into `EstensionRestApi` class as a static method `readFromYaml()`, as well as moving the YML file to the class path in `src/main/resources` as an example to whoever does #86.

src/main/java/org/opensearch/sdk/ExtensionApi.java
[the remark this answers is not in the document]
saratvemulapalli: That makes a lot of sense. May be we can have a legend for conventions something like[1]. 
I agree it makes sense to have them as `ExtensionRestApi` 

[1] https://github.com/opensearch-project/opensearch-plugins/blob/main/CONVENTIONS.md

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
saratvemulapalli: Thanks Dan!

saratvemulapalli (commented):
Thanks @dbwiddis. This is a great start for RestAPIs. 
Could you update the `DESIGN.md`.

dbwiddis:
> Thanks @dbwiddis. This is a great start for RestAPIs. Could you update the `DESIGN.md`.

I can do that as part of the follow-on PR to address #69 (based a lot on the comment discussing my plan there) to avoid delaying this PR any further.

Any other changes needed before this and the companion PR is merged?

peternied (approved):
Thanks for the updates, looks good for my perspective

saratvemulapalli (approved):
Thanks @dbwiddis. 
Gradle check is failing probably because of the PR pending in OpenSearch.

dbwiddis:
> Thanks @dbwiddis. 
> 
> Gradle check is failing probably because of the PR pending in OpenSearch.

Indeed as it passes locally when they branch is published to maven local. 

What will it take to get that companion PR approved?  I'd like to continue work on both branches.

src/test/java/org/opensearch/sdk/TestExtensionRestApi.java
```diff
@@ -0,0 +1,44 @@
...
+    @Override
+    @BeforeEach
+    public void setUp() throws Exception {
+        super.setUp();
+        File file = new File(ExtensionRestPaths.class.getResource(ExtensionRestPaths.EXTENSION_REST_PATHS_DESCRIPTOR).getPath());
+        ObjectMapper objectMapper = new ObjectMapper(new YAMLFactory());
+        extensionRestApi = objectMapper.readValue(file, ExtensionRestPaths.class);
+    }
+
+    @Test
```
owaiskazi19, line 27: I see we have covered the happy path. Any failure cases?

owaiskazi19 (approved):
One small ask :) Rest LGTM!

src/test/java/org/opensearch/sdk/TestExtensionRestApi.java
dbwiddis: Not sure what you are asking here. Either we correctly read the three `String`s from the yml, or it fails and we get an `IOException` in the `setup()` above or one of the assertions below fails.  I see no case where the below test succeeds and there is a problem in the class.

PR #87, opened 2022-08-11
Add ActionListener onFailure to ExtensionsRunner

### LINKED ISSUE #61: [FEATURE] Create Transport API for ActionListener onFailure method
The onFailure method from ActionListener is necessary for integrating AD plugin.  This will be implemented through a Transport API between OpenSearch and the SDK.

Tasks:

- [ ] Create `sendActionListenerOnFailureRequest` method in `ExtensionsRunner.java` that will send the request to OpenSearch

- [ ] Create `handleActionListenerOnFailureRequest` method in `ExtensionsOrchestrator.java` that will call `listener.onFailure(Exception e)` and send an ack back to the SDK

- [ ] Create `handleActionListenerOnFailureRequest` method in `ExtensionsRunner.java` that will receive the ack and print a confirmation

- [ ] Add unit tests for the methods above

ryanbogan:

### Description
Add ActionListener onFailure to ExtensionsRunner

### Issues Resolved
#61 


src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -349,6 +350,26 @@ public void sendLocalNodeRequest(TransportService transportService) {
...
+     *
+     * @param transportService  The TransportService defining the connection to OpenSearch.
+     */
+    public void sendActionListenerOnFailureRequest(TransportService transportService) {
+        logger.info("Sending ActionListener onFailure request to OpenSearch");
+        ActionListenerOnFailureResponseHandler listenerHandler = new ActionListenerOnFailureResponseHandler();
+        try {
+            transportService.sendRequest(
+                opensearchNode,
+                ExtensionsOrchestrator.REQUEST_ACTION_LISTENER_ON_FAILURE,
```
joshpalis, line 364: Similar suggestion as this [comment](https://github.com/opensearch-project/OpenSearch/pull/4191#discussion_r943923921), let's change this to `REQUEST_EXTENSION_ACTION_LISTENER_ON_FAILURE`

joshpalis (changes requested):

joshpalis (dismissed):

owaiskazi19 (changes requested):
We need to identify the exceptions here rather than incrementing count

saratvemulapalli:
@owaiskazi19 @ryanbogan whats needed to keep this moving?

owaiskazi19:
> @owaiskazi19 @ryanbogan whats needed to keep this moving?

We need to catch the exception in the OpenSesrch end coming from the extension rather than incrementing/decrementing counts which is present in the current PR.
@ryanbogan can you update the issue with design, steps to be done, what's completed and what's pending?

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -485,7 +510,7 @@ private Settings getSettings() {
      *
      * @param timeout  The timeout for the listener in milliseconds. A timeout of 0 means no timeout.
      */
-    public void startActionListener(int timeout) {
+    public static void startActionListener(int timeout) {
```
owaiskazi19, line 513: Why do we have to change it to `static`?

dbwiddis: @owaiskazi19 The normal answer to that question is that it doesn't affect any instance variables so it would help the compiler (potentially) optimize.  However...

_Why_ doesn't it affect any instance variables?  It looks like we instantiate an `ActionListener` on the next line, tell it to run, and then promptly forget about it, so we can't get any info from it again, and it is eligible for garbage collection unless one of its internal variables has a reference retained.

Which it doesn't.  It does open a `ServerSocket` which looks internally like it starts a thread... which dies on handshake or after an hour, but that's not attached to any object so it looks like it just floats out there for a while, keeping the socket alive.

There doesn't appear to be any way to prevent multiple action listeners from simultaneously executing, which in the case of a static method is possibly a good thing, as each of multiple extensions may start one up... except that after starting them up they have no reference to it anymore.

All this is to say this should probably _not_ be static, but there's a bigger problem of not keeping track of the action listener at all. Shouldn't it be an instance variable and shouldn't this method only start it if it hasn't already been started?  Or as part of a constructor?

I realize this problem existed before this PR, but I suspect the failure will never be triggered (or if it is, nobody will get the result of it).

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -476,6 +477,30 @@ public void sendLocalNodeRequest(TransportService transportService) {
         }
     }
 
+    /**
+     * Requests the ActionListener onFailure method to be run by OpenSearch.  The result will be handled by a {@link ActionListenerOnFailureResponseHandler}.
+     *
+     * @param transportService  The TransportService defining the connection to OpenSearch.
+     * @param failureException The exception to be sent to OpenSearch
+     */
+    public void sendActionListenerOnFailureRequest(TransportService transportService, Exception failureException) {
```
dbwiddis, line 486: How is this Request associated with the ActionListener instantiated inside the `startActionListener()` method or anywhere else?  Where (other than the test) is this method ever called?

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -184,6 +185,14 @@ public void testLocalNodeRequest() {
         verify(transportService, times(1)).sendRequest(any(), anyString(), any(), any(LocalNodeResponseHandler.class));
     }
 
+    @Test
+    public void testActionListenerOnFailureRequest() {
+
+        extensionsRunner.sendActionListenerOnFailureRequest(transportService, new Exception("Test failure"));
+
+        verify(transportService, times(1)).sendRequest(any(), anyString(), any(), any(ActionListenerOnFailureResponseHandler.class));
```
dbwiddis, line 193: shouldn't we be able to test the value of the response?

dbwiddis (changes requested):
I'm really confused.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
ryanbogan: I don't remember changing it to static, so I'm guessing it's just something that happened with merge conflicts.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
ryanbogan: It is not called anywhere yet, but that functionality will be added later when the actual createDetector is implemented as far as I know.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
owaiskazi19: That makes sense @dbwiddis! Thanks for the detailed explanation.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
owaiskazi19: We are creating these APIs for SDK like:
https://github.com/opensearch-project/opensearch-sdk-java/blob/main/src/main/java/org/opensearch/sdk/ExtensionsRunner.java#L443

https://github.com/opensearch-project/opensearch-sdk-java/blob/main/src/main/java/org/opensearch/sdk/ExtensionsRunner.java#L463

They will be called once we integrate SDK with AD extension.

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
ryanbogan: Resolved offline

owaiskazi19 (approved):

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -209,6 +210,7 @@ TransportResponse handleOpenSearchRequest(OpenSearchRequest request) throws Exce
     IndicesModuleResponse handleIndicesModuleRequest(IndicesModuleRequest indicesModuleRequest, TransportService transportService) {
         logger.info("Registering Indices Module Request received from OpenSearch");
         IndicesModuleResponse indicesModuleResponse = new IndicesModuleResponse(true, true, true);
+        sendActionListenerOnFailureRequest(transportService, new Exception("Test failure"));
```
dbwiddis, line 213: Per comments on the companion PR, really try to avoid throwing broad Exceptions.  

Also I don't understand why we are throwing a "test" failure in the main code.  This even looks like it's in a non-failure path.

dbwiddis (dismissed):
Approved with a big question on why there's a test failure being sent in the production side of code.  Is this leftover debug code?

dbwiddis (approved):

owaiskazi19 (approved):

PR #103, opened 2022-08-25
Only send one extension info when initializing

### LINKED ISSUE #93: [FEATURE] Only send one extension's information in InitializeExtensionsRequest

Currently, the `ExtensionsOrchestrator` iterates over all extensions when initializing, passing the `DiscoveryNode` representing a single extension:
```java
public void extensionsInitialize() {
    for (DiscoveryNode extensionNode : extensionsList) {
        extensionInitialize(extensionNode);
    }
}
```
However, inside this method, rather than using just the extension being initialized, (a copy of) the full `extensionsList` is passed as a transport parameter:
```java
transportService.sendRequest(
    extensionNode,
    REQUEST_EXTENSION_ACTION_NAME,
    new InitializeExtensionsRequest(transportService.getLocalNode(), new ArrayList<DiscoveryExtension>(extensionsList)),
    extensionResponseHandler
);
```
In order to obtain information from that List, the extension is forced to iterate over the entire list to match the name as present in the extension to get any of the configuration information.
```java
for (DiscoveryExtension de : extensionInitRequest.getExtensions()) {
    if (de.getName().equals(extensionSettings.getExtensionName())) {
        setUniqueId(de.getId());
        break;
    }
}
```
This results in needless list iteration, copying, serializing/deserializing, and sending excess information between nodes, along with introducing the need for exception handling in the case there is a name mismatch between the name configured in the list of extensions and the name configured in the extension itself.

Change the second argument in the constructor for `InitializeExtensionsRequest` from a `List<DiscoveryExtension>` to a single `DiscoveryExtension` object, and pass that argument during initialization.  (The initialization loop should iterate over `DiscoveryExtension` rather than the `DiscoveryNode` superclass).

This will require a decision on which extension name configuration takes precedence (or removal of one of the duplicate configs).

1. Leaving it as is:  wasteful of CPU and network bandwidth
2. Having a common extension list in a separate file and only passing a unique key to all methods that need to gain information for it: possibly a better solution but more complex and error-prone than the simple change proposed.

See https://github.com/opensearch-project/opensearch-sdk-java/pull/74#discussion_r938321402
> It only occurs in the case there's a misconfiguration where the extension name in `extensions.yml` doesn't match the extension name defined by the extension, so terminating the extension on a runtime exception is appropriate.
> 
> Which I might argue is an unnecessary requirement and one of the two names should be ignored.  However by passing the entire list of `DiscoveryExtension`s matching the names is the only reliable way to fetch the unique ID.  (Maybe we should have the extension itself declare the uniqueId?)
>
> I think the correct thing to do is not bother with the name check, but change `PluginRequest` (which should be renamed) to pass a single instance of the correct `DiscoveryExtension` instead of the list.  

See also #52 which this may solve.

dbwiddis:
Companion PR: https://github.com/opensearch-project/OpenSearch/pull/4302

### Description
Sends a single `DiscoveryExtension` object when initializing an extension, rather than the entire list.

### Issues Resolved
Fixes #93

### DIFF
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -22,7 +22,6 @@ import org.opensearch.common.network.NetworkModule;
 import org.opensearch.common.network.NetworkService;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.util.PageCacheRecycler;
-import org.opensearch.extensions.DiscoveryExtension;
 import org.opensearch.extensions.ExtensionBooleanResponse;
 import org.opensearch.discovery.InitializeExtensionsRequest;
 import org.opensearch.discovery.InitializeExtensionsResponse;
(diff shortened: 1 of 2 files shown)


dbwiddis:
Gradle check failing due to changes in feature/extension branch. Passes after publishToMavenLocal:

> BUILD SUCCESSFUL in 16s
> 12 actionable tasks: 8 executed, 4 up-to-date

owaiskazi19 (approved):

saratvemulapalli (approved):

Issue #86, opened 2022-08-11
[BUG] Package extension.yml file


joshpalis:
### What is the bug?
Building and running the opensearch-sdk artifact would look for the `extension.yml` file within the `opensearch-sdk/build/distributions/opensearch-sdk-1.0.0-SNAPSHOT/src/test/resources/` directory instead of the `opensearch-sdk/src/test/resources` directory

### How can one reproduce the bug?
Run `./gradlew build` in the `opensearch-sdk` directory to produce the artifact. Navigate to `opensearch-sdk/build/distributions/` and untar the generated artifact using `tar -xvf opensearch-sdk-1.0.0-SNAPSHOT.tar`. Navigate to `opensearch-sdk-1.0.0-SNAPSHOT/` and run `./bin/opensearch-sdk`

This will produce the following error : 
`Exception in thread "main" java.io.FileNotFoundException: src/test/resources/extension.yml (No such file or directory)`

In order to fix this issue, we must create the directory within the package and generate the extension.yml file here so that the artifact has access.

EDIT: Following discussion, the extensions.yml file will be moved to the config directory where opensearch.yml is currently stored.

dbwiddis:
This is not as simple as just changing the directory paths.  

Both "file paths" are accessible with the current setup but neither file path will work in a compiled jar file.

The file should be moved to `src/main/resources` which will put it at the root of the classpath.  You'll then have to use some variant of `getResource()` or `getResourceAsStream()` from the classloader, such as:
```java
// reads from the root of the classpath so will find something in src/main/resources
InputStream in = this.getClass().getClassLoader().getResourceAsStream("extension.yml");
// or get the filepath
Path ymlPath = this.getClass().getClassLoader().getResource("extension.yml").getPath();
```

You can also get it from the class itself using NIO, but this is relative to the package, so if you're in the `org.opensearch.sdk` package you need to navigate up 3 directories:
```java
// reads from the root of the package
Path ymlPath = Paths.get(this.getClass().getResource("../../../extension.yml").toURI());
// or
File ymlFile = new File(this.getClass().getResource("../../../extension.yml").toURI());
// I think if you lead with a slash you can work around the relative package directory things, so try
Path ymlPath = Paths.get(this.getClass().getResource("/extension.yml").toURI());
// or
String ymlPath = this.getClass().getResource("../../config/dev/example.xml").getPath();
```
or variations of the above.

[...]

Issue #37, opened 2022-07-11
When a request comes into an Extension the identity information should be available for use


peternied:
When requests from OpenSearch -> Extension arrive there should be a payload that includes the identity of the current user.  Given a request there should be a way to extract this information.  When requests are made from Extension -> OpenSearch this information should be included.

* [ ] Extract a JWT token through the transport interface on an incoming request
* [ ] Include properties
   * [ ] identifier
   * [ ] expiration
* [ ] Outbound requests to OpenSearch include the identity information.

peternied:
@dbwiddis I didn't get around to making a pull request into OpenSearch for an example, but for the moment I've been liking some concepts from [Principal](https://docs.oracle.com/javase/8/docs/api/java/security/Principal.html) from java's implementation, here is what I was thinking:

```java
public interface Principal {
   /** Unique and immutable identifier */
   public String name();

   /** Returns true if this principle exists within the identity */
   public boolean implies(final Identity ident);
}
```

In a use case inside an extension it could look like

```java
public Response createResource(Request r) {
   final ExtensionResource d = new ExtensionResource(...);
   d.setOwner(r.getPrincipalIdentity().getName());
   // Other interesting things
   return ...;
}

What do you think about this?

DarshitChanpura:
Initial Steps to follow:
1. Extract user from Opensearch core, and Create a user identity with relevant info
2. Add principal identifier to be able to pass it to the extension
3. Figure out how to handle this principal information inside extension

dbwiddis:
> there should be a payload that includes the identity of the current user

I'm trying to be precise and keep "identity" (logged in user) and "access" (permissions) separate.

Consider this scenario:  User X (identity) authenticates (logs in) and OpenSearch determines them to have a role permitting access to Y data.  They set up a scheduled job to do something with that data, and then they log out.

Now, later, the extension needs access to that data.  There is no "identity" but the job still has permission to run.

> I've been liking some concepts from [Principal](https://docs.oracle.com/javase/8/docs/api/java/security/Principal.html) from java's implementation

I agree this is the correct way to go.  This is the superclass of both a user and a group (where the group may be more closely aligned to "roles").

> Initial Steps to follow:
> 
> 1. Extract user from Opensearch core, and Create a user identity with relevant info

Yes with the caveat that this isn't necessarily a user, it's a "principal" which may be a group, may be a non-user service object, etc.

> 2. Add principal identifier to be able to pass it to the extension

[...]

Issue #105, opened 2022-08-26
GetActions: OpenSearch to add support for registering Actions


saratvemulapalli:
- ExtensionsOrchestrator to handle register Transport Actions.
- SDK to make a request to OpenSearch, register actions.

PR #101, opened 2022-08-23
Register REST API and forward REST requests to associated extension

dbwiddis:
Companion PR: https://github.com/opensearch-project/OpenSearch/pull/4282

### Description

This and the companion PR linked above completes the work begun in #74 by registering the received API with the `RestController`.  When OpenSearch receives a `RestRequest` matching one of these registered routes, it forwards it to the appropriate extension for further action.

See the design schematic and description in #64
 - Steps 1 and 2 were already completed
 - [#4100](https://github.com/opensearch-project/OpenSearch/pull/4100) and #74 implemented step 3
 - This and the companion PR implement steps 4 through 8
 - Step 9 is outside the scope of this PR.

### Issues Resolved
Closes #64, Closes #69, and Closes #70

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #109, opened 2022-08-29
Add Extension API interfaces and implement in sample Hello World extension

### LINKED ISSUE #102: [FEATURE] Extension-local means to register handlers and act on them

In Plugins, the REST API is implemented as Action objects implementing the `RestHandler` interface.  The `routes()` method is implemented to identify which REST requests the Action should handle.  

Plugins have a `public List<RestHandler> getRestHandlers()` which provides these handlers to OpenSearch where they are registered with the RestController.

In the Extensions model, the API is communicated from the extension to OpenSearch, where all registered `Route`s for that extension are forwarded to the extension to handle.  We need a local registry of the `RestHandler` implementations of an extension so we can:
 - generate the API list to send to OpenSearch
 - handle requests forwarded from OpenSearch 

EIther an instance of the RestController or a new object with similar (but reduced) functionality to still correlate REST requests (method+URI) with an action.

There is no specific need to continue to use the RestHandler interface for extensions; an entirely new API could be developed. However, re-using the RestHandler interface will speed the transition of existing plugins to extensions by allowing their existing Actions to be re-used with minimal alteration.

See comments [here](https://github.com/opensearch-project/opensearch-sdk-java/issues/69#issuecomment-1217203666).

dbwiddis:
### Description

This PR creates new interfaces `Extension` and `ExtensionRestHandler` which developers will implement when writing new Extensions, and demonstrates their use in a sample `HelloWorldExtension` which responds as expected to its registered OpenSearch REST request:
```
[~/git/OpenSearch] % curl -X GET localhost:9200/_extensions/_opensearch-sdk-1/hello
Hello, World!%
```

The `Extension` interface requires two methods to be implemented by developers.
 - `getExtensionSettings()` which the `ExtensionsRunner` needs to bind the host and port.
 - `getExtensionRestHandlers()` which obtains a list of objects implementing the `ExtensionRestHandler` interface, which will handle the REST requests registered with OpenSearch.

The `ExtensionRestHandler` interface requires two methods to be implemented by developers.
 - `routes()` which corresponds directly to the `routes()` method required by the `RestHandler` interface implemented by plugins.
 - `handleRequest()` which is analogous to the logic eventually called by the `RestHandler`'s `handleRequest()` method, implemented in plugins with the `prepareRequest()` method. 

#102 discussed possibly implementing the `RestHandler` interface in extensions to speed migrating plugins, but presented the following challenges:
 - The full method signature of `prepareRequest()` is `public RestChannelConsumer prepareRequest(final RestRequest request, final NodeClient client)`.  This involves a `Channel` and `Client` to handle a `RestRequest`:
   - The channel and end user are on the OpenSearch side of the transport mechanism so can't be used directly
   - The `RestRequest` isn't yet configured to be deserialized to send over transport
     - Even if we add that, it includes HTTP headers, potentially including basic authentication (user/password). For security purposes these are **_not_** being sent to extensions, and it's difficult or impossible to remove or filter them from an existing object.
 - We could create new `Channel`, `Client`, and `RestRequest` objects with appropriate information, but that introduces a lot of complexity and overhead (and duplication) when instead we can just send and return only the information that we really need to.

The `ExtensionRestHandler` interface still corresponds very closely to the Plugin action implementation of `BaseRestHandler`.

Presently the action method takes only the method and URI and returns a String, as the scope of this PR is to implement that functionality.  This signature is envisioned to change over time as we add new features.  For example:
 - We will probably add a new field for communicating a lookup key for an Identity object as discussed in #37
 - We will need to handle URL queries with parameters. I've added #111 to address this.
 - We will eventually need to change the return type to an object encapsulating both a message and a REST response code as we implement more actions. I've added #112 to address this.
 - We will possibly need some support for NamedWriteables added (not sure yet).
 - Other items from the `RestRequest` object (excluding the HTML headers) may be relevant (not sure yet).

In addition to the interface implementations and sample extension, this PR also:
 - Makes multiple changes in `ExtensionsRunner` to:
   - Add a `run()` method to instantiate using the settings of an Extension.
   - Use the Extension's settings to configure its hostaddress/port
 - Removes the `ExtensionRestPaths` class (and yml and test files) as it is replaced by the `Extension` API implementation (and tests).
 - Moves the `extension.yml` file to the classpath at `/sample/extension-settings.yml` as part of the sample extension implementation.  This may make #86 no longer needed.
 - Updates DESIGN with details
 - Changes `build.gradle` to execute the `main()` method of the sample extension rather than running `ExtensionsRunner`. I believe this is closer to the intent of the SDK.
 - Updates the DEVELOPER_GUIDE appropriately for the additional configuration

### Issues Resolved

Closes #102.

Removes the need for #86.


ryanbogan (approved):
I agree that #86 is no longer necessary with these changes.

src/test/java/org/opensearch/sdk/sample/rest/TestRestHelloAction.java
```diff
@@ -0,0 +1,45 @@
...
+    @Test
+    public void testRoutes() {
+        List<Route> routes = helloAction.routes();
+        assertEquals(1, routes.size());
+        assertEquals(Method.GET, routes.get(0).getMethod());
+        assertEquals("/hello", routes.get(0).getPath());
+    }
+
+    @Test
+    public void testExtensioResponse() {
```
joshpalis, line 39: Small change, fix name and change to `testExtensionResponse`

joshpalis (changes requested):
Besides the test name change, LGTM! Thanks for adding to the design document, the ExtensionAction diagram definitely helped me gain a better understanding of the workflow. I also agree that this change makes #86 obsolete, should I go ahead and close this bug out now?

src/test/java/org/opensearch/sdk/sample/rest/TestRestHelloAction.java
dbwiddis: Silencio!

joshpalis (dismissed):

dbwiddis:
@owaiskazi19 and @saratvemulapalli, after reviewing #80 and #105 I can already see a need for name changes in my implementation.  Holding off on making them until you review but adding some context here so you can make the appropriate comments when you get to them in the API.

What I have implemented in the `ExtensionAction` interface directly correlates to a `RestHandler` implementation, and the code enables a the equivalent of `getRestHandlers()`.  I chose the `Action` name because of classes like `RestIndexAnomalyDetectorAction`.    However, I see this is separate from `getActions()` which returns classes like `IndexAnomalyDetectorAction` <insert pun about "No Rest for the weary.">.   Clearly just 'Action' is not clear enough.  Some options:
 - Change the interface name to `ExtensionRestAction`
   - Closer to the implementation naming but possibly still confusing
 - Change the Interface name to `ExtensionRestHandler`
   - I kind of prefer this one, and also would rename `getExtensionResponse()` (corresponding to `prepareRequest()` in existing Rest*Actions) to `handleRequest` to align (mostly) with the `RestHandler` interface.

saratvemulapalli:
> @owaiskazi19 and @saratvemulapalli, after reviewing #80 and #105 I can already see a need for name changes in my implementation.
> 
> What I have implemented in the `ExtensionAction` interface directly correlates to a `RestHandler` implementation, and the code enables a the equivalent of `getRestHandlers()`. I chose the `Action` name because of classes like `RestIndexAnomalyDetectorAction`. However, I see this is separate from `getActions()` which returns classes like `IndexAnomalyDetectorAction` <insert pun about "No Rest for the weary.">. Clearly just 'Action' is not clear enough. Some options:
> 
> * Change the interface name to `ExtensionRestAction`
>   
>   * Closer to the implementation naming but possibly still confusing
> * Change the Interface name to `ExtensionRestHandler`
>   
>   * I kind of prefer this one, and also would rename `getExtensionResponse()` (corresponding to `prepareRequest()` in existing Rest*Actions) to `handleRequest` to align (mostly) with the `RestHandler` interface.

Thanks @dbwiddis. I love these changes. Its a great start of helping extension authors write simple extensions. 

`Action` is used in multiple ways in OpenSearch but I like the second option `ExtensionRestHandler`. 
This keeps it simple and for all actions we can assume its always transport.

saratvemulapalli:
Changes look good to me. I'll do another pass once we merge https://github.com/opensearch-project/opensearch-sdk-java/pull/101 and https://github.com/opensearch-project/OpenSearch/pull/4282

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -18,6 +18,9 @@
 import org.opensearch.common.io.stream.NamedWriteableRegistry;
```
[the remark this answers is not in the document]
saratvemulapalli: A good problem to have :)

dbwiddis:
> Could you create an issue for this to be followed up on? One way we could manage this more securely is to deep copy then remove/unset all headers from the request before its passed along to the extension.

@peternied already created with #111 and comments on #37.

The challenge with a "deep copy" is we still have to send it across Transport, so we're stuck "disassembling" the `RestRequest` in any case.  The only question is whether to reassemble it on the Extension end before passing it to the Extension action, or to just pass the component pieces.

For now, passing the pieces seems easiest but we can always change it (up through version 0.9.99.999) as we iterate on new extensions and see what's needed.

> Might be worthwhile to ask, besides the method, path, body - do extensions need any other functionality to support a given REST request?

I think as we migrate plugins to extensions this will answer itself.

dbwiddis:
@saratvemulapalli @ryanbogan @joshpalis @owaiskazi19 

Have rebased this PR on main after merge of #101 which invalidated earlier reviews.  No changes, though, would like to get this merged today before I submit yet another PR building on this, addressing #112

joshpalis (approved):

saratvemulapalli (approved):

Issue #40, opened 2022-07-11
Create a permissions check API for extensions to us with custom resources


peternied:
OpenSearch should be responsible for inspecting the user, roles, resources to ensure standard practices are used.  A permissions check API should be designed and implemented to offload this work from extensions creators.

* [ ] Design an API for extensions to use to check users permissions on a resource
   * [ ] First examine and consider the model used by Anomaly Detection plugin
   * [ ] Examine other plugins for permissions additional model that might need to be supported.
   * [ ] Consider any other permissions implementations that should be covered.
* [ ] Implement in OpenSearch
* [ ] Implement in the SDK, allowing for drop in replacement of anomaly detection.

peternied:
For additional context checkout some anti-patterns, a pluginside permission check in AD, [link](https://github.com/opensearch-project/anomaly-detection/blob/875b03c1c7596cb34d74fea285c28d949cfb0d19/src/main/java/org/opensearch/ad/util/ParseUtils.java#L568).  Read up more at for how detectors are configured and expressed as user stories  https://opensearch.org/docs/latest/monitoring-plugins/ad/index/

peternied:
One of the APIs I was thinking of could look like this:
```
GET /security/checkResourcePermission
{
   // User is embedden in the requset
   "permissions": "NAME OF THE PERMISSION FOR THE RESOURCE",
   "resourceId": "The identifier of the resource",
   "additionalData": { "other information that can be used to process the request for allowing/disallowing"}
}

Issue #111, opened 2022-08-29
[FEATURE] Pass REST params to extensions


dbwiddis:
### Is your feature request related to a problem?

As part of #64, [OS #4282](https://github.com/opensearch-project/OpenSearch/pull/4282) established a basic way of forwarding REST requests to extensions, passing the HTTP method and uri fields.

To expand functionality, the `params()` of the `RestRequest` should also be forwarded.  This will be necessary for handling additional types of requests.

### What solution would you like?
Update the `RestExecuteOnExtensionRequest` class to include `params()`.  Update associated code in the `RestSendToExtensionAction`, `ExtensionsRunner`, `ExtensionRestHandler` interface, and the sample extension's `RestHelloAction`.

### What alternatives have you considered?
Sending the entire `RestRequest` rather than extracting parts of it. This doesn't work from a security perspective as sensitive information can be included in headers, and these are not easily removed.

### Do you have any additional context?
See first comment at #109.

mukulsethi22:
Hey, can I work on this issue ?

dbwiddis:
Hey, @mukulsethi22 thanks for your interest.  We'd love for you to work on this.   Some quick notes:
 - This is a must-have capability to support #24 and #100, which we are targeting in the next few weeks.  So we'd love if you can dive in promptly to work on this.  If not, we might end up doing it first. :-)
 - Note that this requires [OpenSearch PR #4282](https://github.com/opensearch-project/OpenSearch/pull/4282) to be merged for the work needed on the OpenSearch repo side. That part should hopefully happen soon.
   - It will be on the `feature/extensions` branch after merging
   - If you want to get started now you can use my branch from that PR
 - Note that it also requires #109 to be merged, which may take a little longer, and may have some name changes from the current state (and even this issue description).  
   - Working off my branch on this PR is probably a good idea, then you can cherry-pick the commits over to a fresh branch off main once it's merged.
 - Take a look at the `DEVELOPER_GUIDE` and `DESIGN` documents to orient yourself.

Let me know if you have any questions, happy to guide you along here.

dbwiddis:
Hi @mukulsethi22, just a quick update.  The PRs mentioned above have been merged, so the code is available for you to fork and start work if you're ready.
 - You'll want to fork the `OpenSearch` project and create your branch of the latest code in the `feature/extensions` branch.
 - You can fork this repo and branch off `main`.
[...]

PR #118, opened 2022-08-31
WIP on a prototype of authorization for extension actions

cwperks:
Signed-off-by: Craig Perkins <cwperx@amazon.com>

### Description

This is a WIP on creating a way for extensions to check user permissions before performing an action. I have created another sample extension called `CrudExtension` with 2 actions: `RestCreateAction` and `RestUpdateAction`. As of now these actions are `noops` and return `PUT /create successful` and `POST /update successful` respectively, but I plan to continue working on this and extend the `RestExecuteOnExtensionRequest` to send a body, parameters and headers with a request to an extension (Looks like @dbwiddis opened a draft PR to address this: https://github.com/opensearch-project/OpenSearch/pull/4356).

This work is enabled by @DarshitChanpura's work on passing Identity from core to extensions (https://github.com/opensearch-project/OpenSearch/pull/4299). I built on top of this and added `AuthorizationRequest` and `AuthorizationResponse` where we can add richer privilege evaluation in core. I have not opened a PR against core for this yet, but here are the changes: https://github.com/opensearch-project/OpenSearch/compare/feature/extensions...cwperks:authorize-request

In this prototype I have created an annotation called `RequiresPermission` which any action that extends `ExtensionRestHandler` can decorate their respective `handleRequest` with a list of permissions that will permit a user to perform an action. This model is overly simplified and will need to get richer to support more complex actions, I would appreciate any input to this effect on the best way of designing permissions checks for actions. 

If an action is decorated with `RequiresPermissions` then inside `ExtensionsRunner.handleRestExecuteOnExtensionRequest` this PR uses reflection to get the list of permissions and authors an `AuthorizationRequest` to see if the action is permitted. 

Questions:

 - This model is overly simplified and permits or denies a user from performing an action, but how should we extend this to protect actions on resources?
 - When an extension is registered is it given a namespace for permissions? How are the permissions registered?

### Issues Resolved

https://github.com/opensearch-project/opensearch-sdk-java/issues/40


src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -229,6 +239,40 @@ RestExecuteOnExtensionResponse handleRestExecuteOnExtensionRequest(RestExecuteOn
...
+            UUID principalUuid = request.getPrincipalUUID();
+
+            AuthorizationResponseHandler authorizationResponseHandler = new AuthorizationResponseHandler();
+            try {
+                AuthorizationResponse authResponse = this.extensionTransportService.submitRequest(
+                        opensearchNode,
+                        ExtensionsOrchestrator.REQUEST_EXTENSION_AUTHORIZE_ACTION,
+                        new AuthorizationRequest(principalUuid, Arrays.asList(permissionsRequired.permissions())),
+                        authorizationResponseHandler
+                ).get();
```
peternied, line 265: Seems like the response should have more information? How does an extension deal with permissions denial?

peternied (commented):
Thanks for getting this out @cwperks Interesting approach with the annotations, much easier than other approaches.  I would love to see this expanded on with more dynamic content, such as filtering resource by an identifier match.

dbwiddis (commented):
About to get on the road for several hours and will look at this later, but wanted to offer some quick feedback:
 - I see here that we're basically changing the "HelloWorld" to the Crud.   OK if we want to make this a new canonical or if this is just a demo PR, but we should answer the question earlier rather than later of "what sort of samples do we want to provide users" and give them an option to pick either or both to run in gradle with a swithc.
 - regarding passing more information to the handler, see #111 that may help with the params part.  There's an open question on headers and how they impact security; how can we strip out sensitive headers (e.g., basic auth) while preserving other info?   
 - See #114 that changes the Extension Rest Handler return type from String to RestResponse.  that may change your prototype when it is merged.
 - Agree with @peternied comment on avoiding reflection.  Current pattern is to explicitly register an instance of each handler in the `Extension` implementation.

dbwiddis:
> Looks like @dbwiddis opened a draft PR to address this: [opensearch-project/OpenSearch#4356](https://github.com/opensearch-project/OpenSearch/pull/4356))

For clarity, this PR is the other direction (response) rather than the request, and also is an actual PR rather than a "draft", that I'm hopeful will be merged this week.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -229,6 +239,40 @@ RestExecuteOnExtensionResponse handleRestExecuteOnExtensionRequest(RestExecuteOn
...
+            } catch (InterruptedException e) {
+                // Request should not be authorized if a definitive response cannot be received
+                throw new RuntimeException(e);
```
dbwiddis, line 268: not clear why we are throwing runtime exceptions rather than returning appropriate error codes to the user (timeout here, not authorized in the permissions case)

src/main/java/org/opensearch/sdk/sample/crud/rest/RestCreateAction.java
```diff
@@ -0,0 +1,46 @@
...
+    @RequiresPermissions(
+        permissions = { "extensions:sample/crud/create" }
+    )
+    @Override
+    public String handleRequest(Method method, String uri) {
+        // TODO modify RestExecuteOnExtensionRequest to get request body. (Should it get params and headers too?)
+        if (Method.PUT.equals(method) && "/crud/create".equals(uri)) {
+            return SUCCESSFUL;
+        }
+        return null;
```
dbwiddis, line 43: see PR #114 for new signature here

src/main/java/org/opensearch/sdk/sample/crud/rest/RestUpdateAction.java
```diff
@@ -0,0 +1,48 @@
...
+        return singletonList(new Route(POST, "/crud/update"));
+    }
+
+    // How should the extension list what permissions it wants to create?
+    // Will the permissions be part of the API spec of the extension?
+    @RequiresPermissions(
+            permissions = { "extensions:sample/crud/update" }
+    )
+
+    // TODO modify RestExecuteOnExtensionRequest to get request body. (Should it get params and headers too?)
```
dbwiddis, line 39: There's a PR open for the params, #111.

Headers may contain security information. Can we strip that part out and pass the rest?

build.gradle
```diff
@@ -27,8 +27,8 @@ apply plugin: 'application'
 apply from: 'gradle/formatting.gradle'
 apply plugin: 'maven-publish'
 
-mainClassName = 'org.opensearch.sdk.ExtensionsRunner'
-  
+mainClassName = 'org.opensearch.sdk.sample.helloworld.HelloWorldExtension'
```
dbwiddis, line 30: Per my latest changes this should stay as ExtensionsRunner.

src/main/java/org/opensearch/sdk/ExtensionRestHandler.java
```diff
@@ -19,12 +19,22 @@
  * This interface defines methods which an extension REST handler (action) must provide.
  * It is the Extension counterpart to core OpenSearch {@link RestHandler}.
  */
-public interface ExtensionRestHandler {
+public abstract class ExtensionRestHandler {
+
+    private String actionPrefix;
+
+    public ExtensionRestHandler(String extensionId) {
```
dbwiddis, line 26: Not clear on the motivation for abstract class vs. interface but I presume it's the presence of this variable?  If so should this be a `protected` constructor?   Also I'd prefer to keep interface but add an intermediate "BaseExtensionRestHandler" class, possibly to add functionality such as suggested in #128.

src/main/java/org/opensearch/sdk/ExtensionRestHandler.java
```diff
@@ -36,5 +46,5 @@ public interface ExtensionRestHandler {
      * @param uri The URI to handle.
      * @return An {@link ExtensionRestResponse} to the request.
      */
-    ExtensionRestResponse handleRequest(Method method, String uri);
+    public abstract ExtensionRestResponse handleRequest(Method method, String uri);
```
dbwiddis, line 49: FYI, I'm strongly leaning toward a `handleRequest(ExtensionRestRequest request)` method here, to keep a stable API for users as we add more params-over-transport (method+uri+params+access token+???)

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -237,6 +250,39 @@ RestExecuteOnExtensionResponse handleRestExecuteOnExtensionRequest(RestExecuteOn
                 "No handler for " + ExtensionRestPathRegistry.restPathToString(request.getMethod(), request.getUri())
             );
         }
+        // Action name here will be prefix (extensions:{extensionName} + route + method.
+        // For an extensions named `hw`, a call to /hello yields an action name of extensions:hw/hello/get
+        String actionName = restHandler.getActionPrefix() + request.getUri() + "/" + request.getMethod().toString().toLowerCase();
```
dbwiddis, line 255: we should be able to get the uniqueID for the extension to use as the prefix here.   Also suggest `String.join()` with the slashes.  Also I think `toLowerCase()` might run afoul of some OpenSearch conventions banning its use (for localization or something).  Shouldn't be an issue for UTF-8 compatible Method enum, though...

src/main/java/org/opensearch/sdk/sample/crud/rest/CrudRestHandler.java
```diff
@@ -0,0 +1,65 @@
...
+    // How should the extension list what permissions it wants to create?
+    // Will the permissions be part of the API spec of the extension?
+    @Override
+    public ExtensionRestResponse handleRequest(Method method, String uri) {
+        System.out.println("method: " + method);
+        System.out.println("URI: " + uri);
+        // TODO modify RestExecuteOnExtensionRequest to get request body. (Should it get params and headers too?)
+        List<String> consumedParams = new ArrayList<>();
+        if (Method.PUT.equals(method) && "/crud/create".equals(uri)) {
+            return new ExtensionRestResponse(OK, CREATE_SUCCESS, consumedParams);
```
dbwiddis, line 54: should probably let this constructor accept null here and assume empty list at that point.

src/main/resources/sample/crud-settings.yml
```diff
@@ -0,0 +1,3 @@
+extensionName: sample-crud
+hostAddress: 127.0.0.1
+hostPort: 7331
```
dbwiddis, line 3: 5h0u1dn'7 7h15 83 1337? 😁

src/main/java/org/opensearch/sdk/ExtensionRestHandler.java
cwperks: Has this been implemented yet? I was planning to rebase and rework when the change is available.

dbwiddis: Nope, just an idea in my head... well, and now this comment.  Feel free to implement it yourself as part of this if it's going to be merged.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
cwperks: This is really the crux of what I want to propose for this prototype of authorization. As far as I can see in core, each action is handled by its own file that extends `ActionType`. Currently in the SDK we `implement ExtensionRestHandler` and are responsible for figuring out how to handle each individual route that's registered inside `handleRequest`. The `handleRequest` function could end up being huge for an extension with many routes.

By concatenating the prefix (`extensions:{extensionName}`) + route + method this is creating a unique action name which can be used in an authorization request to let core determine ultimately whether the action request is granted or denied. This is the convention I have chosen for this iteration of the Draft PR.

The idea here is a simple AuthorizationRequest using the principalIdentifier, actionName, a list of resourceIds and a Map of additionalParams to keep it simple to start with and the response would have a status of either granted or denied. The logic of determining whether a request is granted or denied will be in a companion PR on the `feature/extensions` branch of core.

We may want to consider other ways extensions can determine what actions a user can perform on a resource as well.

I'm currently working backwards from the AD plugin to determine what is needed in the SDK to authorize actions for the extension. Admittedly, I'm having some writers block in designing a good generic way that extensions developers can use to protect actions and resources so any feedback to the design of the AuthorizationRequest is welcome.

dbwiddis: > The handleRequest function could end up being huge for an extension with many routes.

There can be multiple implementations of `ExtensionRestHandler`.  There's only one in this sample, but look here for example: https://github.com/opensearch-project/anomaly-detection/tree/main/src/main/java/org/opensearch/ad/rest

Each Action class is the same as our ExtensionRestHandler class.  Some have a few routes, some just have one.  I could have easily left the GET /hello in one action and created a new action for PUT.

> By concatenating the prefix (extensions:{extensionName}) + route + method this is creating a unique action name 

That (with unique ID) the same concatenation I'm using in the RestController.  But extension Name isn't guaranteed to be unique which is why I was suggesting uniqueId.

cwperks: Ok, I see my error now. I conflated `uniqueId` and `extensionName` because I didn't differentiate between the `extensions.yml` defined in the `extensions/extensions.yml` of core and the extensions specific settings (the sample configs in `main/resources/sample/) containing a name, hostAddress and port. 

There may be a problem with defining `uniqueId` in `extensions.yml` instead of the extension settings files. I see why its in `extensions.yml` so admins can namespace the extensions as they see fit (and its used for the base of the API), but plugin permissions usually use a shorthand for the plugin. i.e. `cluster:admin/opendistro/ad/detector/write` where `ad` is the shorthand for anomaly detector. We can keep `uniqueId` so admins can assign them as needed, but we also need static unique identifier for each extension.

dbwiddis: > We can keep uniqueId so admins can assign them as needed, but we may also need a static unique identifier for each extension.

The fully qualified class name should be unique, assuming reverse DNS package naming standard.

Docs/Extensions-Security.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Checking user permissions
+
+Before any action is performed on an extension, the action should be preceded with an `AuthorizationRequest` to 
+determine whether the user of an extension has the required privileges to perform the action. An authorization
+request consists of the name of a permission, a user identifier and a map of required parameters needed to check if a 
+user has the required permissions. AuthorizationRequest handling is performed in core and yet to be implemented.
+
+```java
+public class AuthorizationRequest extends TransportRequest {
```
peternied, line 105: @cwperks Please create a separate PR in OpenSearch for the AuthorizationRequest so we can discuss the design elements and see if we should merge it.

cwperks: @peternied The change I am making relies on @DarshitChanpura's unmerged changes: https://github.com/opensearch-project/OpenSearch/pull/4299/files#diff-a1079c2a0ec10aedc11808ac7129483612a01d759d34bf3499b76cdce5c98cfb

The diff is available here: https://github.com/DarshitChanpura/OpenSearch/compare/principal-identity-extensions...cwperks:OpenSearch:authorize-request

I am not able to create a PR based on the branch by @DarshitChanpura

cwperks: I'll change the requesterIdentifier from an object to a string in order to create a PR and it will be reworked with @DarshitChanpura's branch when that is ready.

cwperks: @peternied Here's the PR for adding `AuthorizationRequest` to `feature/extensions` of core: https://github.com/opensearch-project/OpenSearch/pull/4516

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -237,6 +246,50 @@ RestExecuteOnExtensionResponse handleRestExecuteOnExtensionRequest(RestExecuteOn
...
+
+        // TODO validate that caller has the permissions required here
+        PrincipalIdentifierToken requestIssuerIdentifier = request.getRequestIssuerIdentity();
+
+        AuthorizationResponseHandler authorizationResponseHandler = new AuthorizationResponseHandler();
+        try {
+            AuthorizationResponse authResponse = this.extensionTransportService.submitRequest(
+                    opensearchNode,
+                    ExtensionsOrchestrator.REQUEST_EXTENSION_AUTHORIZE_ACTION,
+                    new AuthorizationRequest(getUniqueId(), requestIssuerIdentifier, requiredPermission, params),
```
DarshitChanpura, line 271: `AuthorizationRequest` accepts String as inputType for requestIssueIdentifier (refer [here](https://github.com/opensearch-project/OpenSearch/pull/4516/files#diff-21e48b040c4cc6daff2bb4601378a730026ce338fdae278524e226894cc9aab1R32)), but we are passing in the PrincipalidentifierToken object which creates a mismatch. 

How should we proceed with this?

cwperks: The PR against core was opened to solicit feedback on the interface for submitting AuthorizationRequests from extensions. Passing a String in as a PrincipalIdentifierToken is a placeholder until the work to pass tokens to extensions is complete.

DarshitChanpura: okay..gotcha

src/main/java/org/opensearch/sdk/ExtensionRestHandler.java
```diff
@@ -37,4 +37,4 @@ public interface ExtensionRestHandler {
      * @return An {@link ExtensionRestResponse} to the request.
      */
     ExtensionRestResponse handleRequest(Method method, String uri);
-}
+}
```
dbwiddis, line 40: looks like you deleted the end of file newline.  Linelint will complain!

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -237,6 +246,50 @@ RestExecuteOnExtensionResponse handleRestExecuteOnExtensionRequest(RestExecuteOn
                 "No handler for " + ExtensionRestPathRegistry.restPathToString(request.getMethod(), request.getUri())
             );
         }
+
+        // TODO get this list of resource IDs and addition params from the RestExecuteOnExtentionRequest
```
dbwiddis, line 250: Tracked in https://github.com/opensearch-project/opensearch-sdk-java/issues/111 (should be completed by end of month)

src/main/java/org/opensearch/sdk/authz/ProtectedRoute.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public String getRequiredPermission() {
+        return this.requiredPermission;
+    }
+
+    public Pattern getRouteRegex() {
+        if (getPath() == null) {
+            return null;
+        }
+        String routeRegex = getPath().replaceAll("\\{.*?\\}", "(.*)");
```
dbwiddis, line 36: Regexes are sometimes mysterious.  This one isn't obvious what it's doing or why.  It looks like it's to conform to the expectation of `ROUTE_PARAM_REGEX` but I'm not sure why that is, either.

Can we please have a comment explaining this?  (Unless you switch to glob matching per the comment below...)

src/main/java/org/opensearch/sdk/authz/ProtectedRoute.java
```diff
@@ -0,0 +1,71 @@
...
+        Pattern routePattern = Pattern.compile(routeRegex);
+        return routePattern;
+    }
+
+    public Map<String, Object> getConsumedParams(String uri) {
+        // Extracting only route and URI params for now. Needs to extract params from the body when that is available
+        Map<String, Object> consumedParams = new HashMap<>();
+        Matcher nameMatcher = ROUTE_PARAM_REGEX.matcher(getPath());
```
dbwiddis, line 44: When messing around with replacing wildcard parameters in routes, which are Paths, I thought that glob matching would be an easier implementation than regex.  See https://docs.oracle.com/javase/8/docs/api/java/nio/file/PathMatcher.html

src/main/java/org/opensearch/sdk/sample/crud/rest/CrudRestHandler.java
```diff
@@ -0,0 +1,108 @@
...
+                NOT_FOUND,
+                "Extension REST action improperly configured to handle " + method.name() + " " + uri,
+                List.of()
+        );
+    }
+
+    public ProtectedRoute getMatchingRoute(Method method, String uri) {
+        for (Route r : routes()) {
+            ProtectedRoute protectedRoute = (ProtectedRoute) r;
+            if (protectedRoute.getMethod() == method && uri.matches(protectedRoute.getRouteRegex().pattern())) {
```
dbwiddis, line 102: this iteration is screaming for a `routes.stream()` with a filter on method and URI, returning an Optional. :)

dbwiddis (approved):

cwperks:
@dbwiddis Thank you for all your review comments as I was iterating through different ideas of defining actions and asking for permissions with extensions. I'm closing this PR for now as `feature/identity` is rapidly being developed on and will re-open when its in a more mature place. I plan to keep working on this to show examples of how extensions developers can define permissions in their extensions and how actions are protected in core.

PR #123, opened 2022-09-06
Handle named wildcards (REST path parameters)

### LINKED ISSUE #122: [FEATURE] Support Routes with named wildcards

> In addition to forwarding the params, should we also be able to support routes with parameters?
> 
> i.e. `new Route(POST, "/crud/update/{doc_id}")`
> 
> Our current process for locating `ExtensionRestHandlers` looks for an _exact_ match to the URI to find the respective handler (See https://github.com/opensearch-project/opensearch-sdk-java/blob/main/src/main/java/org/opensearch/sdk/ExtensionsRunner.java#L228).
> 
> Are route parameters functionality we want to support?

_Originally posted by @cwperks in https://github.com/opensearch-project/opensearch-sdk-java/issues/111#issuecomment-1234540303_

We need to change the matching in the `ExtensionsRunner` hashmap from exact, to parsing a request URI and allowing "wildcard" parameters.

Java's path glob matching is probably a great fit here. I'll dig through Rest Handler code to see if that's how it's done there.

dbwiddis:
Companion PR: https://github.com/opensearch-project/OpenSearch/pull/4415

### Description

Adds the ability to handle REST path parameters (named wildcards).  It is necessary to indicate which parameters were consumed.  If the user provides too many parameters, they receive an error.

In local/plugin operation, simply fetching the parameter via `param(key)` consumes it.  Since the parameters are being consumed remotely, this requires communicating the list of consumed parameters, which is done in the response header (normally unused and provided for purposes such as this!)

Primary changes:
 - (First commit) Replaces the `pathMap` in `ExtensionsRunner` with an `ExtensionRestPathRegistry`.
   - The internal storage of the REST paths uses a `PathTrie` so that the matching behavior matches that of the OpenSearch `RestController`
     - Exact match conflicts don't work
     - Conflicting wildcards don't work
     - An exact match takes precedence over a wildcard match when they overlap
 - (Fourth commit) Changes the return type from extension rest handlers to an `ExtensionRestResponse`.  This is a simple wrapper for a `BytesRestResponse` with added handling for a list of consumed parameters
   - The `RestResponse` class already provides a facility for custom headers, presently only used with the `OpenSearchException` class.  Adding a new key for consumed parameters doesn't conflict with this usage.
   - See the companion PR for how this is parsed out on the other side
   - There will be a need to more elegantly handle all the types of parameters or better document usage for developers; this detail is deferred until #111 is complete. 
 - (Third Commit) Updated the `HelloWorldExtension` to add another Route:
   - Using `PUT` and a `{name}` parameter allows changing the name of the world for future greetings. 
 - (Throughout) Added tests for the `ExtensionRestPathRegistry`, `ExtensionRestResponse`, and sample extension behavior, as well as updating other tests as required.

Other changes (mostly in second commit):
 - Changed the default `./gradlew run` to the test settings on `ExtensionsRunner` and added a new `task` for the Hello World extension: it can be executed with `./gradlew helloWorld`.
 - Moved the `HelloWorld` extension into a subpackage of the `sample` package in anticipation of more samples (such as the CRUD sample the security team is working on).
 - Added OpenAPI-compatible spec files (both YAML and JSON). We probably don't need both, but adding for demo/discussion until we settle on something. View by importing to https://editor.swagger.io/
 - Updated DESIGN and DEVELOPER_GUIDE with the latest changes

### Issues Resolved

Fixes #122


dbwiddis:
Rocking the code coverage 💪 

Also poor-man's integration test:
```
[~/git/OpenSearch] % curl -X GET localhost:9200/_extensions/_opensearch-sdk-1/hello
Hello, World!%

[~/git/OpenSearch] % curl -X PUT localhost:9200/_extensions/_opensearch-sdk-1/hello/OpenSearch+Build+Hamster
Updated the world's name to OpenSearch Build Hamster%

[~/git/OpenSearch] % curl -X GET localhost:9200/_extensions/_opensearch-sdk-1/hello
Hello, OpenSearch Build Hamster!% 
```

owaiskazi19:
@dbwiddis do you think we should have break this PR into smaller pieces?

dbwiddis:
> @dbwiddis do you think we should have break this PR into smaller pieces?

Possibly, but some of the pieces might break as they depend on each other.  The first commit is the main part of the implementation and is only 4 files (2 changes + 2 tests) and is most of what needs to be reviewed.  The rest is implementing/demoing/documenting/bugfixing that first commit.

For ease of reviewing, you can click on each commit separately:
 - ab2e0dc78eaa749fe3cde5d832e47b64efe6ac56 is the registry code using the PathTrie rather than a hashmap and is the main topic of the PR, with only 4 files to review
 - fea6ebc6fb334b117f685df4919647e172b690a0 is mostly just moving files into a subpackage, with the addition of some documentation updates and enabling gradle to run the extension as other than the default "run".  It shouldnt be too hard to review all of these changes quickly.
 - a2c127f285d12237c60f0495acbc99f0c44f36ae updates the sample extension (in its new location so the diff is easier to see) to take advantage of the new capability, and helps demo what the first commit enables.
 - 0b6a19bf176d847b3dfee427aac4c740daa0db1c and the companion PR add the list of consumed parameters to the response, a bug I didn't find until I thought I was done and tried manual integration testing.

dbwiddis:
@owaiskazi19 I added some notes to the outline in the first comment noting which commit they correspond to. I think that should help reviewers logically piece together all the changes.)

owaiskazi19:
Probably comments on specific commits are not visible here. Something new I learned :)

dbwiddis:
> Probably comments on specific commits are not visible here. 

Yeah, bummer. Click on the "Commits" tab up top :)

I did see and reply to your comments, though...

src/main/java/org/opensearch/sdk/sample/helloworld/spec/openapi.json
```diff
@@ -0,0 +1,89 @@
+{
```
saratvemulapalli, line 1: For my learning: Why do we need specs with `.json` and `.yml`? 
Is it mostly of different clients to generate APIs?

src/test/java/org/opensearch/sdk/TestExtensionRestPathRegistry.java
```diff
@@ -0,0 +1,99 @@
...
+        assertEquals(fooHandler, extensionRestPathRegistry.getHandler(Method.GET, "/foo"));
+        assertNull(extensionRestPathRegistry.getHandler(Method.PUT, "/foo"));
+
+        // Exact match and wildcard match can overlap, exact takes priority
+        assertEquals(barHandler, extensionRestPathRegistry.getHandler(Method.PUT, "/bar/mars"));
+        assertEquals(bazHandler, extensionRestPathRegistry.getHandler(Method.PUT, "/bar/baz"));
+        assertNull(extensionRestPathRegistry.getHandler(Method.PUT, "/bar/mars/bar"));
+
+        assertEquals(bazHandler, extensionRestPathRegistry.getHandler(Method.POST, "/baz/europa/qux"));
+        assertNull(extensionRestPathRegistry.getHandler(Method.POST, "/bar/europa"));
```
saratvemulapalli, line 83: I was like why doesn't this work and I realized the Method is POST. 
For few paths, folks usually want to support multiple Methods like GET, DELETE etc. 
Is this something you plan on to add ?

saratvemulapalli (dismissed):
Thanks @dbwiddis for the changes.

src/main/java/org/opensearch/sdk/sample/helloworld/spec/openapi.json
dbwiddis: We don't need both. It's probably an either-or.  I find the YAML one easier to read/edit, but the JSON one seems common for the OpenSearch spec.  Also there is discussion about using other tools to generate either of these from a simpler format.  This is mostly just a prompt/reminder for the future.  See #127

src/test/java/org/opensearch/sdk/TestExtensionRestPathRegistry.java
dbwiddis: The capability is there, you just add it in `routes()` and handle it appropriately in `handleRequest()`.   I'm trying not to overcomplicate hello world, though! :)

src/main/java/org/opensearch/sdk/sample/helloworld/spec/openapi.json
dbwiddis: (Another reason I prefer YAML: comments are supported.)

(But OpenSearch uses JSON.)

image

saratvemulapalli (approved):

src/main/java/org/opensearch/sdk/ExtensionRestPathRegistry.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * This class registers REST paths from extension Rest Handlers.
+ */
+public class ExtensionRestPathRegistry {
+
+    // PathTrie to match paths to handlers
+    private PathTrie<ExtensionRestHandler> pathTrie = new PathTrie<>(RestUtils.REST_DECODER);
+    // List to return registered handlers
+    private List<String> registeredPaths = new ArrayList<>();
```
peternied, line 25: This gives off a code smell where this class is additional functionality to the data structure, can we add `getKeys()` to the PathTrie implementation?

peternied (approved):
One nit, great forward progress.

src/main/java/org/opensearch/sdk/ExtensionRestPathRegistry.java
dbwiddis: @peternied I tried that route.  Unfortunately the `PathTrie` does not have functionality to list keys.  You can insert, but the retrieve functionality requires you to give it a path (and possibly parameters to match).  There is no way to generate a list of paths (short of a brute force dictionary attack of all possible strings to see which ones work).

This lack of functionality on the PathTrie is also a challenge in testing (no way to validate/confirm that a path has been inserted, other than an IT) and might be a reasonable feature request to OpenSearch.  But for now, for the purposes of deserializing the extension REST API to pass it over, this slightly smelly code is really the only reasonable option.

Issue #78, opened 2022-08-05
Add create component extension point support for AD


saratvemulapalli:
With AD detectors, lets add support for create components and integrate with the work already done with https://github.com/opensearch-project/opensearch-sdk/pull/44

joshpalis:
`createComponents` allows plugins to return a collection of objects to OpenSearch. PluginService iterates through all plugins and passes several services and registries, ie `clusterService`, `environment`, `client`, `NamedXContentRegistry` to help these plugins to configure and generate their components. These objects are then consolidated into a list, which is then bound to a Guice module during node initialization and then further used for dependency injection. In the case of extensibility, since these extensions are running an independent processes, we must add support to transport the necessary information from the services required to successfully create these components. In the case of Anomaly Detection, in order to support creating a detector, the required components are `AnomalyDetectionIndices`, `SearchFeatureDao`, and `ADTaskManager`.  

These components require several services from OpenSearch, and currently there is added support for the following : 

- Client : replaced by [SDKClient](https://github.com/opensearch-project/opensearch-sdk-java/blob/main/src/main/java/org/opensearch/sdk/SDKClient.java)

Some of the required services from OpenSearch are not necessary to add support for as they are not required for creating a detector : 

- NamedXContentRegistry : replaced with an empty collection

These objects require transport APIs to be created in order to access :

- ClusterService
- Settings

`ClusterService` is used by Anomaly Detection to access the `ClusterSettings` in order to add custom settings update consumers. These custom consumers are added within the object constructors of each component, so it is necessary to enable extensions to perform these actions as part of component creation. 

For example : 

- For `ADTaskManager` : https://github.com/joshpalis/anomaly-detection/blob/572fd8d4f08691326d92535f88f7d5f06ea4ea63/src/main/java/org/opensearch/ad/task/ADTaskManager.java#L233 
- For `AnomalyDetectionIndices` : https://github.com/joshpalis/anomaly-detection/blob/572fd8d4f08691326d92535f88f7d5f06ea4ea63/src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java#L201
- For `SearchFeatureDao` : https://github.com/joshpalis/anomaly-detection/blob/572fd8d4f08691326d92535f88f7d5f06ea4ea63/src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java#L117

To add support for extensions to add custom settings update consumers during create components : 

[...]

Issue #79, opened 2022-08-05
[Meta] [FEATURE] Add getSettings support for AD


owaiskazi19:
This issue is to register custom settings support for extensions. 

We can split this into 3 issues:

- [ ] #121
- [x] Add support for transporting Setting from Extension to OpenSearch.
- [x] Add an API to register settings from an extension.

saratvemulapalli:
Taking a stab at this. 
Looks like all the settings we are registering are not needed for create detector[1].
This issue can be split into 3 issues:
- Remove unused settings to unblock creating a detector for AD extension.
- Add support for transporting `Setting` from Extension to OpenSearch.
- Add an API to register settings from an extension. 

[1] https://github.com/opensearch-project/anomaly-detection/blob/feature/extensions/src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java#L797-L809

joshpalis:
getSettings support is required for creating a detector.

During `createComponents` , Anomaly Detection creates three components for use in creating a detector : `SearchFeatureDao`, `ADTaskManager`, and `AnomalyDetectionIndices`. During component construction, each component sets fields based on values retrieved from settings registered within OpenSearch. Additionally, these components also add setting update consumers, essentially listeners, for each setting it has registered, in order to update these fields whenever a setting has been changed. Therefore, it is necessary for these settings to be first registered within OpenSearch prior to creating these components, and also prior to creating a detector.

For [ADTaskManager](https://github.com/opensearch-project/anomaly-detection/blob/11d094aaa1ab5efc9ff80532884f714ff38680a3/src/main/java/org/opensearch/ad/task/ADTaskManager.java#L207), the required settings are : 

- MAX_OLD_AD_TASK_DOCS_PER_DETECTOR
- BATCH_TASK_PIECE_INTERVAL_SECONDS
- DELETE_AD_RESULT_WHEN_DELETE_DETECTOR
- MAX_BATCH_TASK_PER_NODE
- MAX_RUNNING_ENTITIES_PER_DETECTOR_FOR_HISTORICAL_ANALYSIS
- REQUEST_TIMEOUT

For [SearchFeatureDao](https://github.com/opensearch-project/anomaly-detection/blob/11d094aaa1ab5efc9ff80532884f714ff38680a3/src/main/java/org/opensearch/ad/feature/SearchFeatureDao.java#L99), the required settings are : 

- MAX_ENTITIES_FOR_PREVIEW
[...]

PR #138, opened 2022-09-14
[Feature/extensions] Adding support for Extension Create Components [environment settings and add setting update consumers]

joshpalis:
### Description
Adds support for extensions to request environment setting values from opensearch, to add setting update consumers to cluster settings, and to receive update setting requests from Opensearch whenever a setting is updated.

Companion PR : https://github.com/opensearch-project/OpenSearch/pull/4517

Note : Currently in a draft state until https://github.com/opensearch-project/opensearch-sdk-java/issues/79 is resolved

### Issues Resolved
https://github.com/opensearch-project/opensearch-sdk-java/issues/78


dbwiddis (dismissed):
LGTM!

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -232,6 +251,66 @@ ExtensionBooleanResponse handleIndicesModuleNameRequest(IndicesModuleRequest ind
...
+        logger.info("Registering UpdateSettingsRequest received from OpenSearch");
+
+        boolean settingUpdateStatus = true;
+
+        WriteableSetting.SettingType settingType = updateSettingsRequest.getSettingType();
+        Setting<?> componentSetting = updateSettingsRequest.getComponentSetting();
+        Object data = updateSettingsRequest.getData();
+
+        // Setting updater in OpenSearch performs setting change validation, only need to cast the consumer to the corresponding type and
+        // invoke the consumer
```
dbwiddis, line 271: Not a request to change anything in your PR but a note for awareness.  The original validator and parser are lost in the existing setup, so if there was some extra validation step (for example, minimum value integer) it might not properly validate on the OpenSearch side.   That's fine for now as it will eventually be handled by #154.

dbwiddis (approved):

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -181,6 +199,7 @@ InitializeExtensionsResponse handleExtensionInitRequest(InitializeExtensionsRequ
         } finally {
             // After sending successful response to initialization, send the REST API and Settings
             setOpensearchNode(opensearchNode);
+            setExtensionNode(extensionInitRequest.getExtension());
```
owaiskazi19, line 202: What's the reason to set the node here since the node is already present in the `extensionInitRequest`?

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -232,6 +251,66 @@ ExtensionBooleanResponse handleIndicesModuleNameRequest(IndicesModuleRequest ind
...
     }
 
+    /**
+     * Handles a request to update a setting from OpenSearch.  Extensions must register their setting keys and consumers within the settingUpdateConsumer map
+     *
+     * @param updateSettingsRequest  The request to handle.
+     * @return A response acknowledging the request.
+     */
+    @SuppressWarnings("unchecked")
+    ExtensionBooleanResponse handleUpdateSettingsRequest(UpdateSettingsRequest updateSettingsRequest) {
```
owaiskazi19, line 261: Separate handler class for this handle?

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -536,6 +624,74 @@ public void sendActionListenerOnFailureRequest(TransportService transportService
...
+        // Determine if setting update consumers have been registered
+        if (this.settingUpdateConsumers.isEmpty()) {
+            throw new Exception("There are no setting update consumers registered");
+        } else {
+
+            // Extract registered settings from setting update consumer map
+            List<Setting<?>> componentSettings = new ArrayList<>(this.settingUpdateConsumers.size());
+            componentSettings.addAll(this.settingUpdateConsumers.keySet());
+
+            AddSettingsUpdateConsumerResponseHandler addSettingsUpdateConsumerResponseHandler =
```
owaiskazi19, line 679: Can we use `ExtensionBoolerHandler` here?

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
joshpalis: We set the node here and pass the DiscoveryExtension object back to the SDK. This is then later passed to OpenSearch within the `AddSettingUpdateConsumerRequest` to enable the callback consumer to send a transport request to the extension node that registered the setting consumer

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
joshpalis: Sure, I'll replace this and delete the AddSettingsUpdateConsumerResponseHandler

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
owaiskazi19: Can't we directly pass `extensionInitRequest.getExtension()` to `AddSettingUpdateConsumerRequest`? In the future, we need to pass an entire list of extenionsNode so setting a specific node doesn't look good here.

joshpalis: The `InitializeExtensionsRequest` object is only available within the scope of `handleExtensionInitRequest` method, so in order to get the node when creating an `AddSettingUpdateConsumerRequest`, it needs to be available within the SDK. `AddSettingUpdateConsumerRequests` will not be initialized within `handleExtensionInitRequest`, so we cannot directly pass the extensionNode from the `extensionInitRequest` directly. Instead, this will be sent during the create-components workflow, after extensions have registered their setting consumers. All `setExtensionNode(extensionInitRequest.getExtension())` does is set the `extensionNode` in order to configure callback requests to this extension

dbwiddis (approved):

ryanbogan (approved):

Issue #136, opened 2022-09-13
[FEATURE] Create abstract Base classes for Extension and ExtensionRestHandler


dbwiddis:
### Is your feature request related to a problem?

Extension developers need a stable API (interface) to implement extensions.

SDK developers need a way to add additional functionality without forcing extension developers to change their code.

### What solution would you like?

I propose to create a 3-layer inheritance for `Extension` and `ExtensionRestHandler` implementations.
 - The `Extension` interface will include all the methods required by `ExtensionsRunner` including
   - settings
   - rest handlers
   - settings (to be added)
   - components (to be added)
 - A new `BaseExtension` abstract class will provide default methods so users don't need to implement everything in the interface but have the ability to override
   - can provide an empty settings list so extension authors with no custom settings don't need to do anything
   - can be used for default components but allow for the extension implementation to override
   - can move the read settings from yaml convenience code out of interface and simplify the awkward user implementation with all the exception handling
 - Similar layering can be done for the `ExtensionRestHandler` with a `BaseExtensionRestHandler`
   - This will create a place for an abstract class to provide optional syntax for handling REST requests as envisioned by #128 

Also we should create a new `ExtensionRestRequest` class to be the sole argument to `ExtensionRestHandler.handleRequest()`.
 - This will keep the API stable as we add new functionality such as params (#111) and security principal information
 - Goal should be to at least match `RestRequest` method names _exactly_ to ease porting extension code

### What alternatives have you considered?

An interface-only model requires using default methods and exposes a lot of implementation details that aren't really needed.

An abstract-class-only model would work but from an "API" or "SDK" perspective I prefer the cleanliness of an interface.

### Do you have any additional context?
@owaiskazi19 please provide some input on what the API method for creating components should look like (what arguments, etc.)
@cwperks this relates to [this comment](https://github.com/opensearch-project/opensearch-sdk-java/pull/118#discussion_r967268396) on your draft PR.
@mloufra the create components piece may touch your code after refactoring ExtensionsRunner in #116 so will try to coordinate my implementation with yours in #116.  If you can let me know which branch on your local fork you're pushing code to it will be helpful.

mloufra:
> @mloufra the create components piece may touch your code after refactoring ExtensionsRunner in #116 so will try to coordinate my implementation with yours in #116. If you can let me know which branch on your local fork you're pushing code to it will be helpful.

[...]

PR #140, opened 2022-09-16
[Feature/Extensions] Adds Principal-Identity Token to extensions rest request

DarshitChanpura:
### Description
Adds a principal identity token by accepting it from request coming in from OpenSearch core and adds it to the outgoing request to extensions

### Issues Resolved
Related to : https://github.com/opensearch-project/OpenSearch/pull/4299

### DIFF
--- a/src/main/java/org/opensearch/sdk/ExtensionRestHandler.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionRestHandler.java
@@ -31,7 +31,7 @@ public interface ExtensionRestHandler {
      * This method corresponds to the {@link BaseRestHandler#prepareRequest} method.
      * As in that method, consumed parameters must be tracked and returned in the response.
      *
-     * @param restRequest a rest request object for a request to be forwarded to extensions
+     * @param restRequest a REST request object for a request to be forwarded to extensions
      * @return An {@link ExtensionRestResponse} to the request.
      */
     ExtensionRestResponse handleRequest(ExtensionRestRequest restRequest);
--- a/src/main/java/org/opensearch/sdk/ExtensionRestHandler.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionRestHandler.java
@@ -13,7 +13,6 @@ import org.opensearch.rest.BaseRestHandler;
 import org.opensearch.rest.RestHandler;
 import org.opensearch.rest.RestHandler.Route;
 import org.opensearch.rest.RestRequest;
-import org.opensearch.rest.RestRequest.Method;
 
 /**
  * This interface defines methods which an extension REST handler (action) must provide.
@@ -32,9 +31,8 @@ public interface ExtensionRestHandler {
      * This method corresponds to the {@link BaseRestHandler#prepareRequest} method.
      * As in that method, consumed parameters must be tracked and returned in the response.
      *
-     * @param method A REST method.
-     * @param uri The URI to handle.
+     * @param restRequest a rest request object for a request to be forwarded to extensions
      * @return An {@link ExtensionRestResponse} to the request.
      */
-    ExtensionRestResponse handleRequest(Method method, String uri);
+    ExtensionRestResponse handleRequest(ExtensionRestRequest restRequest);
 }
--- a/src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
@@ -78,14 +78,23 @@ public class ExtensionRestRequest extends TransportRequest {
         out.writeNamedWriteable(principalIdentifierToken);
     }
 
+    /**
+     * @return This REST request {@link Method} type
+     */
     public Method method() {
         return method;
     }
 
+    /**
+     * @return This REST request's uri
+     */
     public String uri() {
         return uri;
     }
 
+    /**
+     * @return This REST request issuer's identity token
+     */
     public PrincipalIdentifierToken getRequestIssuerIdentity() {
         return principalIdentifierToken;
     }
--- a/src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
@@ -17,25 +17,47 @@ import org.opensearch.transport.TransportRequest;
 import java.io.IOException;
 import java.util.Objects;
 
+/**
+ * A subclass of {@link TransportRequest} which contains request relevant information
+ * to be utilised in ExtensionRestHandler implementation
+ */
 public class ExtensionRestRequest extends TransportRequest {
-
     private Method method;
     private String uri;
+    /**
+     * The owner of this request object
+     */
     private PrincipalIdentifierToken principalIdentifierToken;
 
-    public ExtensionRestRequest(Method method, String uri, PrincipalIdentifierToken requesterIdentifier) {
+    /**
+     * This object can be instantiated given method, uri and identifier
+     * @param method of type {@link Method}
+     * @param uri url string
+     * @param principalIdentifier the owner of this request
+     */
+    public ExtensionRestRequest(Method method, String uri, PrincipalIdentifierToken principalIdentifier) {
         this.method = method;
         this.uri = uri;
-        this.principalIdentifierToken = requesterIdentifier;
+        this.principalIdentifierToken = principalIdentifier;
     }
 
-    // TODO Check if this constructor is more useful, If so what happens when request object is null
-    protected ExtensionRestRequest(RestExecuteOnExtensionRequest request) {
+    /**
+     * The object to be created from rest request object incoming from OpenSearch
+     * @param request incoming object from OpenSearch
+     * @throws IllegalArgumentException when request is null
+     */
+    protected ExtensionRestRequest(RestExecuteOnExtensionRequest request) throws IllegalArgumentException {
+        if (request == null) throw new IllegalArgumentException("Request object can't be null");
         this.method = request.getMethod();
         this.uri = request.getUri();
         this.principalIdentifierToken = request.getRequestIssuerIdentity();
     }
 
+    /**
+     * Object generated from input stream
+     * @param in Input stream
+     * @throws IOException
+     */
     public ExtensionRestRequest(StreamInput in) throws IOException {
         super(in);
         method = in.readEnum(Method.class);
@@ -43,6 +65,11 @@ public class ExtensionRestRequest extends TransportRequest {
         principalIdentifierToken = in.readNamedWriteable(PrincipalIdentifierToken.class);
     }
 
+    /**
+     * Write this object to output stream
+     * @param out the writeable output stream
+     * @throws IOException
+     */
     @Override
     public void writeTo(StreamOutput out) throws IOException {
         super.writeTo(out);
--- a/src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
@@ -1,0 +1,90 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+package org.opensearch.sdk;
+
+import org.opensearch.common.io.stream.StreamInput;
+import org.opensearch.common.io.stream.StreamOutput;
+import org.opensearch.extensions.rest.RestExecuteOnExtensionRequest;
+import org.opensearch.identity.PrincipalIdentifierToken;
+import org.opensearch.rest.RestRequest;
+import org.opensearch.rest.RestRequest.Method;
+import org.opensearch.transport.TransportRequest;
+
+import java.io.IOException;
+import java.util.Objects;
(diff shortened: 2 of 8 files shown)


DarshitChanpura:
Gradle check should pass after https://github.com/opensearch-project/OpenSearch/pull/4299 is merged

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
```diff
@@ -0,0 +1,90 @@
...
+    protected ExtensionRestRequest(RestExecuteOnExtensionRequest request) {
+        this.method = request.getMethod();
+        this.uri = request.getUri();
+        this.principalIdentifierToken = request.getRequestIssuerIdentity();
+    }
+
+    public ExtensionRestRequest(StreamInput in) throws IOException {
+        super(in);
+        try {
+            method = RestRequest.Method.valueOf(in.readString());
```
dbwiddis, line 43: Thanks for creating this new class. I like it.

When I wrote this earlier but later discovered the StreamInput has enum-writing methods.   So this whole try/catch block could just be 
```java
method = in.readEnum(RestRequest.Method);`
```

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
```diff
@@ -0,0 +1,90 @@
...
+            throw new IOException(e);
+        }
+        uri = in.readString();
+        principalIdentifierToken = in.readNamedWriteable(PrincipalIdentifierToken.class);
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        super.writeTo(out);
+        out.writeString(method.name());
```
dbwiddis, line 54: This can be `out.writeEnum(method);` to match the input change

src/main/java/org/opensearch/sdk/ExtensionRestHandler.java
```diff
@@ -32,9 +31,8 @@ public interface ExtensionRestHandler {
      * This method corresponds to the {@link BaseRestHandler#prepareRequest} method.
      * As in that method, consumed parameters must be tracked and returned in the response.
      *
-     * @param method A REST method.
-     * @param uri The URI to handle.
+     * @param restRequest a rest request object for a request to be forwarded to extensions
```
dbwiddis, line 34: Nit: when by itself, REST is an acronym.

dbwiddis (dismissed):
LGTM, some suggestions inline that you can change or I can change next time I edit these files.

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
```diff
@@ -0,0 +1,90 @@
...
+        out.writeString(uri);
+        out.writeNamedWriteable(principalIdentifierToken);
+    }
+
+    public Method getMethod() {
```
dbwiddis, line 59: I had mentioned this earlier but forgot until I re-reviewed here.   

To ease migration of plugins to extensions, these getter method should parallel the similar methods on the `RestRequest` class.   In this case, `method()`, `uri()`, later `params()`.

peternied (dismissed):
Nicely done @DarshitChanpura, it looks like there are some gradle check failures, I after those have been resolved lets get this in!

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
DarshitChanpura: done

DarshitChanpura:
@peternied and @dbwiddis Thank you for quick reviews! Can you please re-review this PR.

Peter, the gradle check should resolve after changes for principal identity token are merged in core https://github.com/opensearch-project/opensearch-sdk-java/pull/140#issuecomment-1249502728

src/main/java/org/opensearch/sdk/sample/helloworld/rest/RestHelloAction.java
```diff
@@ -37,9 +38,13 @@ public List<Route> routes() {
     }
 
     @Override
-    public ExtensionRestResponse handleRequest(Method method, String uri) {
+    public ExtensionRestResponse handleRequest(ExtensionRestRequest request) {
         // We need to track which parameters are consumed to pass back to OpenSearch
         List<String> consumedParams = new ArrayList<>();
+        Method method = request.method();
+        String uri = request.uri();
+        // TODO: should we extract token and then add it to the list of consumed parameters?
```
dbwiddis, line 46: I think these are params from the client (named wildcards or REST parameters). You're adding the token separately.  So I think no.

dbwiddis (dismissed):
LGTM.  

You don't need to consume the token as it's not a parameter.  

This won't break anything by leaving it there for now (somewhere a map will return null and be ignored) unless by some crazy circumstance the user has a parameter exactly matching your token.  And we can remove it when `params()` gets implemented.  But if you have time, nix it.

src/main/java/org/opensearch/sdk/sample/helloworld/rest/RestHelloAction.java
DarshitChanpura: okay... i will remove it

DarshitChanpura:
@owaiskazi19 and @saratvemulapalli Can you please also review this PR? I'm unable to add you as reviewers directly and hence this comment.

dbwiddis (approved):

dbwiddis:
Added @owaiskazi19 and @saratvemulapalli as reviewers.  This PR is a priority to get merged with API changes impacting #136.

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.opensearch.common.io.stream.StreamOutput;
+import org.opensearch.extensions.rest.RestExecuteOnExtensionRequest;
+import org.opensearch.identity.PrincipalIdentifierToken;
+import org.opensearch.rest.RestRequest.Method;
+import org.opensearch.transport.TransportRequest;
+
+import java.io.IOException;
+import java.util.Objects;
+
+public class ExtensionRestRequest extends TransportRequest {
```
owaiskazi19, line 20: Javadoc for the class

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        super.writeTo(out);
+        out.writeEnum(method);
+        out.writeString(uri);
+        out.writeNamedWriteable(principalIdentifierToken);
+    }
+
+    public Method method() {
```
owaiskazi19, line 54:
```suggestion
    public Method getMethod() {
```

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
```diff
@@ -0,0 +1,85 @@
...
+        return method;
+    }
+
+    public String uri() {
```
owaiskazi19, line 58:
```suggestion
    public String getUri() {
```

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -237,8 +237,15 @@ RestExecuteOnExtensionResponse handleRestExecuteOnExtensionRequest(RestExecuteOn
                 "No handler for " + ExtensionRestPathRegistry.restPathToString(request.getMethod(), request.getUri())
             );
         }
+        // ExtensionRestRequest restRequest = new ExtensionRestRequest(request);
```
owaiskazi19, line 240: Can we remove the commented code?

src/main/java/org/opensearch/sdk/ExtensionRestHandler.java
```diff
@@ -32,9 +31,8 @@ public interface ExtensionRestHandler {
      * This method corresponds to the {@link BaseRestHandler#prepareRequest} method.
      * As in that method, consumed parameters must be tracked and returned in the response.
      *
-     * @param method A REST method.
-     * @param uri The URI to handle.
+     * @param restRequest a REST request object for a request to be forwarded to extensions
      * @return An {@link ExtensionRestResponse} to the request.
      */
-    ExtensionRestResponse handleRequest(Method method, String uri);
+    ExtensionRestResponse handleRequest(ExtensionRestRequest restRequest);
```
owaiskazi19, line 37: Great to see method and uri part of the request now

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.IOException;
+import java.util.Objects;
+
+public class ExtensionRestRequest extends TransportRequest {
+
+    private Method method;
+    private String uri;
+    private PrincipalIdentifierToken principalIdentifierToken;
+
+    public ExtensionRestRequest(Method method, String uri, PrincipalIdentifierToken requesterIdentifier) {
```
owaiskazi19, line 26:
```suggestion
    public ExtensionRestRequest(Method method, String uri, PrincipalIdentifierToken principalIdentifier) {
```

owaiskazi19 (commented):
LGTM with some suggestions

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.opensearch.transport.TransportRequest;
+
+import java.io.IOException;
+import java.util.Objects;
+
+public class ExtensionRestRequest extends TransportRequest {
+
+    private Method method;
+    private String uri;
+    private PrincipalIdentifierToken principalIdentifierToken;
```
saratvemulapalli, line 24: Lets make them final? 
We dont want them to change for a given request.

saratvemulapalli (commented):
Overall the changes look good to me!
Thanks @DarshitChanpura !

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
dbwiddis: And for all the methods and constructors.

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public ExtensionRestRequest(Method method, String uri, PrincipalIdentifierToken requesterIdentifier) {
+        this.method = method;
+        this.uri = uri;
+        this.principalIdentifierToken = requesterIdentifier;
+    }
+
+    // TODO Check if this constructor is more useful, If so what happens when request object is null
```
dbwiddis, line 32: Just put in the javadoc that the argument can't be null, and throw illegalArgumentException if it is.

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
dbwiddis: I specifically requested this be changed to `method()` to align with the same method on `RestRequest`.

Relevant context: [JEP 395](https://openjdk.org/jeps/395).  Since Java 14, "Record" syntax uses the non-Get variant:
> For each component in the header, two members: a public accessor method with the same name and return type as the component, and a private final field with the same type as the component;

The intention is a signal to the coder that the value is "shallowly immutable" (`private final`).   That's the case here.

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
dbwiddis: Same here.  Plugin `prepareRequest()` methods already include `.method()` and `.uri()` calls, this will avoid having to change them.    Eventually we'll add `params()` without "get" for the same reason.

dbwiddis (dismissed):
LGTM other than final modifiers and javadocs as mentioned by other reviewers.

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
DarshitChanpura: Dan suggested we leave it to `method()`...https://github.com/opensearch-project/opensearch-sdk-java/pull/140#discussion_r975898724

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
DarshitChanpura: should we consider a possibility of passing the request object as is and extracting the relevant details during object instantiation in ExtensionRestRequest. Doesn't seem like a good idea to pass the whole request object, but wanted to confirm the same

dbwiddis: Here's the issue discussing it.   https://github.com/opensearch-project/OpenSearch/issues/4429

dbwiddis (dismissed):

dbwiddis:
> [Adds javadoc to fix gradle check](https://github.com/opensearch-project/opensearch-sdk-java/pull/140/commits/650a337d963f8aaf0cc45dc70ec61404571224d5)

Heh... looks like I need to file an issue with our javadoc checking dependency.

dbwiddis (approved):

owaiskazi19 (approved):

PR #146, opened 2022-09-16
Added createComponent to Extension interface

### LINKED ISSUE #136: [FEATURE] Create abstract Base classes for Extension and ExtensionRestHandler

ryanbogan:

### Description
Adds createComponent to Extension interface and createsBaseExtension abstract class

Note: This is dependent on getSettings, so it will remain a draft PR for now.

### Issues Resolved
#136 


owaiskazi19:
Will review this after createComponent workflow is merged and integrated.

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -0,0 +1,169 @@
...
+import org.opensearch.repositories.RepositoriesService;
+import org.opensearch.script.ScriptContext;
+import org.opensearch.script.ScriptEngine;
+import org.opensearch.script.ScriptService;
+import org.opensearch.threadpool.ExecutorBuilder;
+import org.opensearch.threadpool.ThreadPool;
+import org.opensearch.watcher.ResourceWatcherService;
+
+public abstract class BaseExtension implements Extension {
+    private Client client;
```
dbwiddis, line 44: See [this comment](https://github.com/opensearch-project/opensearch-sdk-java/issues/136#issuecomment-1251687393). 

The private field/public getter/protected setter is generally a good model in the case that other classes may need to access these settings.  However in our case, only the subclass will be accessing these.  It may make sense to just make them all `protected`.   Granted, this is called a 'code smell` by some static analyzers.  

So I'm definitely okay with leaving this as is, just wanted to call attention to other reviewers to that comment so they can discuss it!

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -0,0 +1,169 @@
...
+    private Environment environment;
+    private NodeEnvironment nodeEnvironment;
+    private NamedWriteableRegistry namedWriteableRegistry;
+    private IndexNameExpressionResolver indexNameExpressionResolver;
+    private Supplier<RepositoriesService> repositoriesServiceSupplier;
+
+    /**
+     * 
+     */
+    public BaseExtension() throws IOException {
```
dbwiddis, line 59: I'm really kind of iffy on just passing through an exception in a constructor.   Would prefer a nice and verbose error message to the user wrapping (caused by) the thrown exception.  And since this is going to be fatal to the Extension startup, I think an `Error` is more appropriate than an `Exception`.  This also allows us to remove the `throws` clause.

Recommend a try/catch around create components, and in the catch block throw an `IOError` with a message: "Extension failed to initialize due to an exception creating a component: " + exception.getMessage(), and then linking to the exception so the stack trace provides more info.

Open to others' ideas on this as well.

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -0,0 +1,169 @@
...
+        createComponents();
+    }
+
+    private void createComponents() throws IOException {
+        //Settings.EMPTY will eventually be replaced with a getSettings method from the Extension interface
```
dbwiddis, line 64: There is now a `getSettings()` on the interface, but that's a `List<Setting<?>>` not a `Settings`.  This one should be fetchable from the `ExtensionsRunner`, so perhaps this method needs to be in the interface with at least this argument, and then overridden here (so the extension subclass doesn't need to implement it.)

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -0,0 +1,169 @@
...
+        Settings settings = Settings.builder().put("node.name", Property.NodeScope).build();
+        this.threadPool = new ThreadPool(settings, new AtomicReference<>(), Collections.emptyList().toArray(new ExecutorBuilder[0]));
+        this.client = new NodeClient(settings, threadPool);
```
dbwiddis, line 67: This may be the SDKClient from the `ExtensionsRunner`.  Another param to add to the arguments, seems @owaiskazi19 [comment here](https://github.com/opensearch-project/opensearch-sdk-java/issues/136#issuecomment-1251444056) was probably right despite my argument against. :)

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -0,0 +1,169 @@
...
+        this.clusterService = new ClusterService(settings, new ClusterSettings(settings, new HashSet<Setting<?>>()), threadPool);
```
dbwiddis, line 68: This also comes from ExtensionsRunner.  As do more of the below, I'll stop adding comments.

src/main/java/org/opensearch/sdk/sample/helloworld/HelloWorldExtension.java
```diff
@@ -40,7 +41,8 @@ public class HelloWorldExtension implements Extension {
     /**
      * Instantiate this extension, initializing the connection settings and REST actions.
      */
-    public HelloWorldExtension() {
+    public HelloWorldExtension() throws IOException {
+        super();
         try {
             this.settings = initializeSettings();
```
dbwiddis, line 47: Can we try to move all this awkward exception handling for the settings yaml (or objecct) up into the Base Extension and just have a very simple method (or methods) in the Extension itself?

src/test/java/org/opensearch/sdk/sample/helloworld/TestHelloWorldExtension.java
```diff
@@ -46,4 +75,34 @@ public void testExtensionRestHandlers() {
...
+        ResourceWatcherService resourceWatcherService = new ResourceWatcherService(settings, threadPool);
+        ScriptService scriptService = new ScriptService(settings, new HashMap<String, ScriptEngine>(), new HashMap<String, ScriptContext<?>>());
+        NamedXContentRegistry xContentRegistry = new NamedXContentRegistry(new ArrayList<NamedXContentRegistry.Entry>());
+        Environment environment = new Environment(settings, Paths.get(""));
+        NodeEnvironment nodeEnvironment = new NodeEnvironment(settings, environment);
+        NamedWriteableRegistry namedWriteableRegistry = new NamedWriteableRegistryAPI().getRegistry();
+        IndexNameExpressionResolver indexNameExpressionResolver = new IndexNameExpressionResolver(threadPool.getThreadContext());
+        final SetOnce<RepositoriesService> repositoriesServiceReference = new SetOnce<>();
+        Supplier<RepositoriesService> repositoriesServiceSupplier = repositoriesServiceReference::get;
+        assertEquals(threadPool, extension.getThreadPool());
```
dbwiddis, line 94: I assume some instantiation will eventually live here.  A line break between the two sections would help readability.

build.gradle
```diff
@@ -46,7 +46,7 @@ publishing {
 repositories {
     mavenLocal()
     // Remove the commented code below once TransportService is published to maven
```
dbwiddis, line 48: since we're removing the commented code, can we remove the comment to remove the commented code?  (Or add a comment above the comment to remove the comment saying we should remove the comment...? Er, no...)

dbwiddis (commented):
Great start!

As you can see from my comments, you need to add the createComponents() to the interface and then implement (override) it here.  You may not need all the same arguments as the plugins, but it's a reasonable start.

Then ExtensionsRunner will call this method during initialization, passing its own objects here.

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Empty constructor to fulfill abstract class requirements
+     */
+    public BaseExtension() {
+
+    }
+
+    @Override
+    public Collection<Object> createComponents(
+        SDKClient client,
```
dbwiddis, line 34: I think this should be `OpenSearchClient` to allow the user to override with their own preferred client.  The default initialization can store the result of `initializeClient()` from an internal `SDKClient` object.

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -0,0 +1,48 @@
...
+    protected SDKClient client;
+    protected ClusterService clusterService;
+    protected ThreadPool threadPool;
+    protected NamedXContentRegistry xContentRegistry;
+    protected Environment environment;
+
+    /**
+     * Empty constructor to fulfill abstract class requirements
+     */
+    public BaseExtension() {
```
dbwiddis, line 28: Should probably be `protected`.

src/main/java/org/opensearch/sdk/Extension.java
```diff
@@ -48,6 +62,27 @@ default List<Setting<?>> getSettings() {
         return Collections.emptyList();
     }
 
+    /**
+     * Returns components added by this plugin.
+     *
+     * Any components returned that implement {@link LifecycleComponent} will have their lifecycle managed.
+     * Note: To aid in the migration away from guice, all objects returned as components will be bound in guice
```
dbwiddis, line 69: I don't think this guice comment is relevant here.

src/main/java/org/opensearch/sdk/Extension.java
```diff
@@ -48,6 +62,27 @@ default List<Setting<?>> getSettings() {
         return Collections.emptyList();
     }
 
+    /**
+     * Returns components added by this plugin.
```
dbwiddis, line 66: s/plugin/extension/g

src/main/java/org/opensearch/sdk/BaseExtension.java
dbwiddis: After looking more at this, SDKClient doesn't inherit from anything, and internally contains both an `OpenSearchClient` and a `RestClient` (which I think wraps the `OpenSearchClient`).  In the long term I think I'd like this to be a `RestClient`.  In the near term to get this PR completed, let's just leave this as is.  

FYI @owaiskazi19

build.gradle
dbwiddis: ^^^

src/main/java/org/opensearch/sdk/sample/helloworld/HelloWorldExtension.java
dbwiddis: Can we either do this or create an issue to do it later?

dbwiddis (dismissed):
LGTM but needs some followup in new issues.

src/main/java/org/opensearch/sdk/sample/helloworld/HelloWorldExtension.java
dbwiddis: #182

build.gradle
dbwiddis: ^^^ can we please delete this comment?

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -0,0 +1,57 @@
...
+import org.opensearch.threadpool.ThreadPool;
+
+/**
+ * An abstract class that provides sample methods required by extensions
+ */
+public abstract class BaseExtension implements Extension {
+    /**
+     * A client to make requests to the system
+     */
+    protected SDKClient client;
```
dbwiddis, line 23: This will probably change, but OK for this PR.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -132,7 +132,10 @@ protected ExtensionsRunner(Extension extension) throws IOException {
         // save custom transport actions
         this.transportActions = new TransportActions(extension.getActions());
         // initialize the transport service
-        nettyTransport.initializeExtensionTransportService(this.getSettings(), this);
+        ThreadPool threadPool = new ThreadPool(this.getSettings());
+        nettyTransport.initializeExtensionTransportService(this.getSettings(), threadPool, this);
+        // create components
+        extension.createComponents(new SDKClient(), null, threadPool);
```
dbwiddis, line 138: Can we instantiate the client in this class as an instance variable and just pass a reference to the object here?  With this approach if the ExtensionsRunner wanted access to the client it'd not be able to.

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param clusterService A service to allow watching and updating cluster state
+     * @param threadPool A service to allow retrieving an executor to run an async action
+     * @return A collection of objects
+     */
+    public Collection<Object> createComponents(SDKClient client, ClusterService clusterService, ThreadPool threadPool) {
+        this.client = client;
+        this.clusterService = clusterService;
+        this.threadPool = threadPool;
+
+        return Collections.emptyList();
```
dbwiddis, line 55: I'm curious what we ever return from this.

dbwiddis (changes requested):

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Returns components added by this extension.
+     *
+     * @param client A client to make requests to the system
+     * @param clusterService A service to allow watching and updating cluster state
+     * @param threadPool A service to allow retrieving an executor to run an async action
+     * @return A collection of objects
+     */
+    public Collection<Object> createComponents(SDKClient client, ClusterService clusterService, ThreadPool threadPool) {
```
dbwiddis, line 50: Can we also pass the XContentRegistry object?

src/main/java/org/opensearch/sdk/BaseExtension.java
ryanbogan: AD returns this: `return ImmutableList.of(searchFeatureDao, anomalyDetectionIndices, adTaskManager);`

src/main/java/org/opensearch/sdk/BaseExtension.java
ryanbogan: Sure, but I thought you said to take that out since we would only be adding it later.

dbwiddis (approved):

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * A service to allow watching and updating cluster state
+     */
+    protected ClusterService clusterService;
+
+    /**
+     * A service to allow retrieving an executor to run an async action
+     */
+    protected ThreadPool threadPool;
```
owaiskazi19, line 33: Do we need a threadPool in the SDK?

ryanbogan: We need it in the BaseExtension class for createComponents, and this way the same threadpool can be stored for each extension

owaiskazi19: I had an offline discussion with @dbwiddis top keep threadpool either in SDK or create a new in extension. For now we can keep it in SDK but we might want to revisit this and discuss for the right place.

ryanbogan: Sure

ryanbogan: https://github.com/opensearch-project/opensearch-sdk-java/issues/183

owaiskazi19 (approved):

Issue #116, opened 2022-08-31
[FEATURE] Refactor ExtensionsRunner class


dbwiddis:
### Is your feature request related to a problem?

> ExtensionsRunner is getting big, could we compartmentalize this functionality?  Maybe there is a pattern like configuration/registration/request handling?

_Originally posted by @peternied in https://github.com/opensearch-project/opensearch-sdk-java/pull/109#discussion_r958898763_

### What solution would you like?

There are 513 lines.
 - Up to line 156 is headers, constructors, initializers, getters, setters, etc.
 - Line 157-244 is handler methods that are registered in lines 328-390
 - Line 246-390 is initializing the transport service and registering the above request handlers
 - Line 392-471 is sending transport requests
 - Line 473-513 is back to setters and startup methods

We can definitely separate out lines 157-471 into at least one other class, possibly two or three.
 - each of the handlers in lines 157-244 could probably have its own class.
 - ~the `transportService.registerRequestHandler()` calls are all the same except for two lines, and could benefit from a helper method for the 80-90% of that code that is common.~ _(these are all single method calls but take up several lines, probably ok to leave)._
 - The Netty4 and initialize transport service can probably be moved to their own class. (Speaking of netty4, there are some old third party license files that can be removed).

### What alternatives have you considered?

Doing nothing. It's tempting.

### Do you have any additional context?

Like most code, this starts small and grows to the point where the problem is obvious. We're there.

mloufra:
After I did same research, refactoring class is to delete duplicate code and split the class has many method or feature into several classes have single method or feature. In that way, we can make code clean and clear and easily for everyone to maintain and add new feature. And what I am understanding for this issue is to create a new class and move the code which from line 168 `InitializeExtensionsResponse handleExtensionInitRequest(InitializeExtensionsRequest extensionInitRequest) ` to line 250 `RestExecuteOnExtensionResponse handleRestExecuteOnExtensionRequest(RestExecuteOnExtensionRequest request)` into this new class and another new class for `public Netty4Transport getNetty4Transport(Settings settings, ThreadPool threadPool)` method. (the code keeps adding, so just for making sure)  If my understanding is correct, is there any recommendation name for these two class.

saratvemulapalli:
>What alternatives have you considered?
Doing nothing. It's tempting.

😆

mloufra:
> We can definitely separate out lines 157-471 into at least one other class, possibly two or three.
> 
[...]

PR #158, opened 2022-09-26
Refactor handler method

mloufra:
### Description
Create the following five new classes to handle each handler method for refactoring the class in `ExtensionsRunner.java`.

- `ExtensionsInitRequestHandler` handle method `handleExtensionInitRequest`
- `OpensearchRequestHandler` handle method `handleOpenSearchRequest`
- `ExtensionsIndicesModuleRequestHandler` handle method `handleIndicesModuleRequest`
- `ExtensionsIndicesModuleNameRequestHandler` handle method `handleIndicesModuleNameRequest`
- `ExtenionsRestRequestHandler` handle method `handleRestExecuteOnExtensionRequest`

### Issues Resolved


src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -90,14 +86,14 @@ public class ExtensionsRunner {
     private static final Logger logger = LogManager.getLogger(ExtensionsRunner.class);
     private static final String NODE_NAME_SETTING = "node.name";
 
-    private String uniqueId;
-    private DiscoveryNode opensearchNode;
+    private static String uniqueId;
```
dbwiddis, line 89: First of many `static` throughout this PR, and I don't think most of these should be static.  While there's a chance this may not make a difference (if each JVM only has one ExtensionsRunner) I really don't think it's needed.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -125,7 +121,7 @@ public ExtensionsRunner() throws IOException {
             .put(TransportSettings.BIND_HOST.getKey(), extensionSettings.getHostAddress())
             .put(TransportSettings.PORT.getKey(), extensionSettings.getHostPort())
             .build();
-        this.customSettings = Collections.emptyList();
+        ExtensionsRunner.customSettings = Collections.emptyList();
```
dbwiddis, line 124: It is almost always an error to set a static variable in a constructor.  If you instantiate this class multiple times they will all try to set the same value.

src/main/java/org/opensearch/sdk/handlers/ExtensionsRestRequestHandler.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+        // ExtensionRestRequest restRequest = new ExtensionRestRequest(request);
+        ExtensionRestRequest restRequest = new ExtensionRestRequest(
+            request.getMethod(),
+            request.getUri(),
+            request.getRequestIssuerIdentity()
+        );
+
+        // Get response from extension
+        RestResponse response = restHandler.handleRequest(restRequest);
```
dbwiddis, line 53: Should be an ExtensionRestResponse?  (Not sure here as this may contain all we need.)

src/main/java/org/opensearch/sdk/handlers/OpensearchRequestHandler.java
```diff
@@ -0,0 +1,40 @@
...
+     * @throws Exception if the corresponding handler for the request is not present
+     */
+    public static TransportResponse handleOpenSearchRequest(OpenSearchRequest request) throws Exception {
+        // Read enum
+        switch (request.getRequestType()) {
+            case REQUEST_OPENSEARCH_NAMED_WRITEABLE_REGISTRY:
+                return namedWriteableRegistryApi.handleNamedWriteableRegistryRequest(request);
+            // Add additional request handlers here
+            default:
+                throw new Exception("Handler not present for the provided request");
```
dbwiddis, line 36: One of my pet peeves is broad Exceptions.  Pick a subclass.   IllegalArgumentException is probably a good choice.  Also add what was wrong, e.g., show the type that isn't handled.

dbwiddis (changes requested):
I'm not sure why so many variables were changed to `static`, and suspect there's a better way to handle whatever challenge that was meant to address.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
mloufra: The reason that I use many static is to call the method from the new class which I create. I find another way to call these method, and I will delete these static.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
mloufra: Same answer as first comment, I will delete the static in my next commit.

src/main/java/org/opensearch/sdk/handlers/ExtensionsRestRequestHandler.java
mloufra: It's rebase code from #140. Not sure I can answer this question.

src/main/java/org/opensearch/sdk/handlers/OpensearchRequestHandler.java
mloufra: I will change that in next commit.

src/main/java/org/opensearch/sdk/handlers/ExtensionsRestRequestHandler.java
mloufra: I will change that in next commit.

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
```diff
@@ -56,7 +56,7 @@ protected ExtensionRestRequest(RestExecuteOnExtensionRequest request) throws Ill
     /**
      * Object generated from input stream
      * @param in Input stream
-     * @throws IOException
+     * @throws IOException if there a error generating object from input stream
```
owaiskazi19, line 59:
```suggestion
     * @throws IOException if there's an error in generating object from input stream
```

src/main/java/org/opensearch/sdk/ExtensionRestRequest.java
```diff
@@ -68,7 +68,7 @@ public ExtensionRestRequest(StreamInput in) throws IOException {
     /**
      * Write this object to output stream
      * @param out the writeable output stream
-     * @throws IOException
+     * @throws IOException if there is an error writing object to output stream
```
owaiskazi19, line 71: Same as above

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -381,7 +308,7 @@ public void startTransportService(TransportService transportService) {
             false,
             false,
             InitializeExtensionsRequest::new,
-            (request, channel, task) -> channel.sendResponse(handleExtensionInitRequest(request))
+            (request, channel, task) -> channel.sendResponse(extensionsInitRequestHandler.handleExtensionInitRequest(request, this))
```
owaiskazi19, line 311: Why do we need to pass `this` here?

src/main/java/org/opensearch/sdk/handlers/OpensearchRequestHandler.java
```diff
@@ -0,0 +1,40 @@
...
+import org.opensearch.sdk.NamedWriteableRegistryAPI;
+import org.opensearch.sdk.ExtensionsRunner;
+import org.opensearch.transport.TransportResponse;
+
+/**
+ * This class handles the request from OpenSearch to a {@link ExtensionsRunner#startTransportService(TransportService transportService)} call.
+ */
+
+public class OpensearchRequestHandler {
+    private NamedWriteableRegistryAPI namedWriteableRegistryApi = new NamedWriteableRegistryAPI();
```
owaiskazi19, line 20: Let's rename it to `NamedWriteableRegistryAPI` -> `NamedWriteableRegistry`

src/main/java/org/opensearch/sdk/handlers/OpensearchRequestHandler.java
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Handles a request from OpenSearch and invokes the extension point API corresponding with the request type
+     *
+     * @param request  The request to handle.
+     * @return A response to OpenSearch for the corresponding API
+     * @throws Exception if the corresponding handler for the request is not present
+     */
+    public TransportResponse handleOpenSearchRequest(OpenSearchRequest request) throws Exception {
```
owaiskazi19, line 29: This should be a class for `NamedWriteableRegistry` and not `OpenSearchRequest`. Also, we don't need a switch case here. If/else can do the job.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -91,23 +88,46 @@ public class ExtensionsRunner {
     private static final String NODE_NAME_SETTING = "node.name";
 
     private String uniqueId;
-    private DiscoveryNode opensearchNode;
+    /**
+     * This method is call from {@link ExtensionsInitRequestHandler}.
```
owaiskazi19, line 92: Not a method. To better frame it like: `Initializes OpenSearch Node`

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -91,23 +88,46 @@ public class ExtensionsRunner {
...
+     */
+    public DiscoveryNode opensearchNode;
     private DiscoveryExtension extensionNode;
-    private TransportService extensionTransportService = null;
+    /**
+     * This method is call from {@link ExtensionsInitRequestHandler}.
```
owaiskazi19, line 97: Same as above

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -91,23 +88,46 @@ public class ExtensionsRunner {
...
+     */
+    public TransportService extensionTransportService = null;
     // The routes and classes which handle the REST requests
     private final ExtensionRestPathRegistry extensionRestPathRegistry = new ExtensionRestPathRegistry();
     // Custom settings from the extension's getSettings
+    /**
+     * This method is call from {@link ExtensionsInitRequestHandler}.
```
owaiskazi19, line 104: Same as above

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -91,23 +88,46 @@ public class ExtensionsRunner {
...
+     */
     private final List<Setting<?>> customSettings;
     // Node name, host, and port
-    private final Settings settings;
+    /**
+     * This method is call from {@link ExtensionsInitRequestHandler}.
```
owaiskazi19, line 109: Same as above

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -91,23 +88,46 @@ public class ExtensionsRunner {
...
+        new ExtensionsIndicesModuleNameRequestHandler();
+    private ExtensionsRestRequestHandler extensionsRestRequestHandler = new ExtensionsRestRequestHandler();
+
     /*
      * TODO: expose an interface for extension to register actions
      * https://github.com/opensearch-project/opensearch-sdk-java/issues/119
      */
-    private TransportActions transportActions = new TransportActions(new HashMap<>());
+    /**
+     * This method is call from {@link ExtensionsInitRequestHandler}.
```
owaiskazi19, line 127: Not a method. Instantiates a new transportActions will be better here.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -161,131 +181,38 @@ private static ExtensionSettings readExtensionSettings() throws IOException {
...
+     * @param extensionTransportService assign value for extensionTransportService
+     */
     void setExtensionTransportService(TransportService extensionTransportService) {
         this.extensionTransportService = extensionTransportService;
     }
 
-    private void setUniqueId(String id) {
+    /**
+     * This method is call from {@link ExtensionsInitRequestHandler}.
```
owaiskazi19, line 193: Is it really a call because we are defining a setter here?

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
dbwiddis: Well, the field doesn't initialize, it gets initialized.  So I think a mix of what the two of you have here.  "This field is initialized by a call from ...."

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
dbwiddis: Ugh, this is where we have some competing desires.  I don't particularly like making the setter public, but it has to be for the init handler (in another package) to set it.  I do think it's fine to indicate where it's intended to be called from, although adding on that it "sets the Unique ID, used in REST requests to uniquely identify this extension" is a good thing.

dbwiddis (commented):
Looking good with the documentation comments noted!

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
mloufra: Thank you Dan @dbwiddis to answer this question for me.

src/main/java/org/opensearch/sdk/handlers/OpensearchRequestHandler.java
mloufra: Do you means rename the `NamedWriteableRegistryAPI.java` to `NamedWriteableRegistry.java`

owaiskazi19: All the places where we have `NamedWriteableRegistryAPI` -> `NamedWriteableRegistry`.

mloufra: There is another class called `NamedWriteableRegistry` get imported in `NamedWriteableRegistryAPI.java`. If we rename `NamedWriteableRegistryAPI` ->`NamedWriteableRegistry`, it will cause error. Can we use another name to avoid this error.

owaiskazi19: @dbwiddis I'm pretty sure you can suggest a better name here. 
actual-programming-vs-debating-30-minutes-on-how-to-name-a-variable-drake-meme

dbwiddis: How about `ExtensionNamedWriteableRegistry`? :-)

image

src/main/java/org/opensearch/sdk/handlers/ExtensionsRestRequestHandler.java
```diff
@@ -0,0 +1,63 @@
...
+    public RestExecuteOnExtensionResponse handleRestExecuteOnExtensionRequest(RestExecuteOnExtensionRequest request) {
+
+        ExtensionRestHandler restHandler = extensionRestPathRegistry.getHandler(request.getMethod(), request.getUri());
+        if (restHandler == null) {
+            return new RestExecuteOnExtensionResponse(
+                RestStatus.NOT_FOUND,
+                "No handler for " + ExtensionRestPathRegistry.restPathToString(request.getMethod(), request.getUri())
+            );
+        }
+        // ExtensionRestRequest restRequest = new ExtensionRestRequest(request);
```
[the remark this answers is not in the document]
dbwiddis: Don't worry about deleting this. I'm refactoring this as part of #163 (after you merge this) and I'll handle it then.

src/main/java/org/opensearch/sdk/handlers/OpensearchRequestHandler.java
mloufra: I think we can keep this switch case here, it will easier for others adding more case in the future.

src/main/java/org/opensearch/sdk/handlers/ExtensionsRestRequestHandler.java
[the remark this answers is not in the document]
mloufra: Thanks

dbwiddis (approved):
LGTM!   Any additional requests please put on the linked issue for the next round, so we can merge this. :)

owaiskazi19 (approved):

Issue #132, opened 2022-09-09
Meta: [FEATURE/Extensions] Interaction of AD Extension with OpenSearch


owaiskazi19:
### Is your feature request related to a problem?
After integrating SDK with AD Extension in #24, the next steps would be for AD Extension to interact with OpenSearch for creating a detector. Below are the steps of the workflow:
1. Initialize the extension and initialize the interfaces of the extension points.
2. Registered the REST APIs in OpenSearch
3. Send a Create Detector request to extension which will be redirected to the handler from `handleRequest`. If it fails exception will thrown on AD Extension and to throw the same exception on OpenSearch `listener.onFailure` will be invoked.
4. The handler will take care of calling Create Index API to create AD index on OpenSearch using the SDKClient.

![sequence](https://user-images.githubusercontent.com/16099877/189458766-029142d6-ff4e-4128-92c2-6d9af7195f07.png)

There will be bugs/issues after integrating all the SDK features with AD Extension and making it to interact with OpenSearch. We will turn this into a meta issue then and split the work to fix those.

Steps/ Sub issues:

- [x]  [getSettings](https://github.com/opensearch-project/opensearch-sdk-java/issues/159) - Register custom AD Extension settings in OpenSearch https://github.com/opensearch-project/anomaly-detection/pull/686

- [x] [getRestHandler](https://github.com/opensearch-project/opensearch-sdk-java/issues/58) - Register POST route for create detector in OpenSearch
- [x] [createComponent](https://github.com/opensearch-project/opensearch-sdk-java/issues/160) Initialize `searchFeatureDao`, `anomalyDetectionIndices`, `adTaskManager` objects using the Transport APIs created [here](https://github.com/opensearch-project/opensearch-sdk-java/issues/78). - not needed for creating a detector but will be needed for next set of features
- [x] [getActions](https://github.com/opensearch-project/opensearch-sdk-java/issues/161) - Register custom AD actions in OpenSearch - Since AD create detector is working on one extension node. Keeping this on hold as extension to extension communication will be needed for the AD feature which requires details from all the AD extension nodes.
- [x] [Finally](https://github.com/opensearch-project/opensearch-sdk-java/issues/58), send a request to create a detector to OpenSearch, should be redirected to AD Extension and later a request will be send back to OpenSearch to create the detector.

Issue #58, opened 2022-07-21
[Feature/extensions] Implement createDetector request in AD Extension


ryanbogan:
Create a Transport API for creating Index for AD extension

- [x] Initial discussion of the deliverables 
- [x] Figure out if an API already exist in clients/OpenSearch
- [x] Create the API if not
- [x] Register custom POST route for create detector
- [x] Integrate the createIndex API with AD plugin to test if SDKClient is working as expected
- [ ] Integrate the createDetector workflow

owaiskazi19:
As we are looking into creating a SDKClient and call the REST APIs associated with it to create the AD index
Listing out the occurrences of client for creating a detector:
1. https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/transport/IndexAnomalyDetectorTransportAction.java#L95
2. https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/transport/IndexAnomalyDetectorTransportAction.java#L190
3. https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/rest/handler/AbstractAnomalyDetectorActionHandler.java#L279
4. https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java#L538

dbwiddis:
1 and 3 will be handled as part of #22 (or are already handled in #23).  

2 looks like a dependency on the "user" stored with the detector which in our case will already be "null" so probably requires no action but might need some documentation.  

4 is the AdminClient that I think we get for free with the AbstractClient inheritance.

dbwiddis:
Opened up #63 to track item 2 above.  The rest should be no problem.

owaiskazi19:
Keeping this on hold until #24 is finished.

owaiskazi19:
Hi @dbwiddis! To replicate [prepareRequest](https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/rest/RestIndexAnomalyDetectorAction.java#L67) for creating a detector. This issue requires `seqNo`, `primaryTerm`, `refreshPolicy`, `detectorId` which in turns comes from `paramAsLong`, `hasParam`, `param` and `contentParser`. I see we have issue already for the same https://github.com/opensearch-project/opensearch-sdk-java/issues/111.

[...]

PR #163, opened 2022-09-29
Pass REST params and content to extensions

### LINKED ISSUE #111: [FEATURE] Pass REST params to extensions

dbwiddis:
Companion PR on OpenSearch: [#4633](https://github.com/opensearch-project/OpenSearch/pull/4633)

- Gradle checks will fail until companion PR is merged.  They pass locally after `publishToMavenLocal`.
- OpenSearch PR will break SDK build due to a class rename, merge these together.

### Description

Passes the `params()` from the original `RestRequest` object to extensions.  Internally tracks the consumed params (similar to the `RestRequest` class) when they are consumed with the `param(key)` method (similar to existing plugin usage).

During the process, realized that the `RestExecuteOnExtensionRequest` class (in OpenSearch) completely duplicated the `ExtensionRestHandler` class (here), so removed the SDK duplicate and renamed the "Execute" class as `ExtensionRestHandler`.

Additionally added `xContentType` and `content` as well as getters for the params that will enable AD extension functionality.

Remaining TODO to be handled as part of #132 is to instantiate the parser on the SDK side.  All that's needed from the REST request is the xContentType.

### Issues Resolved
Fixes #111


dbwiddis:
Flipping this into draft to handle all the pieces @owaiskazi19 referenced in https://github.com/opensearch-project/opensearch-sdk-java/issues/58#issuecomment-1261599707

src/main/java/org/opensearch/sdk/ExtensionRestResponse.java
```diff
@@ -23,71 +24,76 @@ public class ExtensionRestResponse extends BytesRestResponse {
...
      */
-    public ExtensionRestResponse(RestStatus status, String contentType, BytesReference content, List<String> consumedParams) {
+    public ExtensionRestResponse(ExtensionRestRequest request, RestStatus status, String contentType, BytesReference content) {
         super(status, contentType, content);
-        addConsumedParamHeader(consumedParams);
+        addConsumedHeaders(request.consumedParams(), request.isContentConsumed());
     }
 
-    private void addConsumedParamHeader(List<String> consumedParams) {
+    private void addConsumedHeaders(List<String> consumedParams, boolean contentConusmed) {
```
peternied, line 95: This name is too literal and kind of tripping me up, what about breaking this functionality apart into something like `markParameter(s)Consumed(String([]) param)`, `markContentConsumed()`, and finally `record/reportConsumedItems()`

dbwiddis: I'm not sure "mark...Consumed..." is right, because that's the behavior already happening inside the RestRequest class.  The purpose here is to add header keys and values to pass these already-marked items with the Response to leverage the convenience of the superclass, rather than creating a whole new transport action for them.

I was already thinking of moving this class in #165.  Can we move the rename discussion there so as not to hold up this PR?

DarshitChanpura (approved):
Gradle check won't pass until https://github.com/opensearch-project/OpenSearch/pull/4633 is merged.

peternied (approved):

src/main/java/org/opensearch/sdk/sample/helloworld/rest/RestHelloAction.java
```diff
@@ -39,31 +38,32 @@ public List<Route> routes() {
...
+    private ExtensionRestResponse handlePutRequest(ExtensionRestRequest request) {
+        String name = request.param("name");
+        try {
+            worldName = URLDecoder.decode(name, StandardCharsets.UTF_8);
+        } catch (IllegalArgumentException e) {
+            return new ExtensionRestResponse(request, BAD_REQUEST, e.getMessage());
+        }
+        return new ExtensionRestResponse(request, OK, "Updated the world's name to " + worldName);
+    }
+
```
owaiskazi19, line 64: Can we have a sample POST request too here?

src/main/java/org/opensearch/sdk/ExtensionRestHandler.java
```diff
@@ -9,6 +9,7 @@
 
 import java.util.List;
 
+import org.opensearch.extensions.rest.ExtensionRestRequest;
```
owaiskazi19, line 12: Thinking out loud. Why are we keeping `ExtensionRestRequest` in OpenSearch and not in SDK?

src/main/java/org/opensearch/sdk/ExtensionRestPathRegistry.java
```diff
@@ -28,11 +28,11 @@ public class ExtensionRestPathRegistry {
      * Register a REST handler to handle a method and route in this extension's path registry.
      *
      * @param method  The method to register.
-     * @param uri  The URI to register. May include named wildcards.
+     * @param path  The path to register. May include named wildcards.
      * @param extensionRestHandler  The RestHandler to handle this route
      */
-    public void registerHandler(Method method, String uri, ExtensionRestHandler extensionRestHandler) {
-        String restPath = restPathToString(method, uri);
+    public void registerHandler(Method method, String path, ExtensionRestHandler extensionRestHandler) {
```
owaiskazi19, line 34: I know this change has been done in the companion PR but can we keep it as `uri` instead of `path`?

src/main/java/org/opensearch/sdk/sample/helloworld/rest/RestHelloAction.java
dbwiddis: Can I do that as part of #165?

src/main/java/org/opensearch/sdk/ExtensionRestHandler.java
dbwiddis: Because the previous `RestSendToExtensionRequest` entirely duplicated it.  Literally all the lines were the same except for a few naming differences.

dbwiddis: Also it makes sense (see #165) to move the ExtensionRestResponse there as well since it's really the same thing as the BytesRestResponse with some tweaks.

src/main/java/org/opensearch/sdk/sample/helloworld/rest/RestHelloAction.java
owaiskazi19: Sure @dbwiddis! Not blocking this PR :)

src/main/java/org/opensearch/sdk/ExtensionRestHandler.java
owaiskazi19: This clears my questions. Just one thing to add can we keep the name as `RestSendToExtensionRequest` instead of `ExtensionRestRequest` for better understanding. Can be address in the next PR

src/main/java/org/opensearch/sdk/ExtensionRestPathRegistry.java
dbwiddis: I did a lot of research before changing this.
 - This is a parallel of `RestController`'s registerHandler method which takes `path`.
https://github.com/opensearch-project/OpenSearch/blob/2bfe8b31af354e5d00e4ac67d544ce664f083934/server/src/main/java/org/opensearch/rest/RestController.java#L203

 - The `Route` class takes a `path` which is really what is being both registered _and_ handled
https://github.com/opensearch-project/OpenSearch/blob/3aef125d0dbf2f29d9bf00b84cb52892a27a5328/server/src/main/java/org/opensearch/rest/RestHandler.java#L185-L190

 - `uri` includes the query (e.g. the `?foo=bar&baz=qux` part) and that query part is not being registered
https://github.com/opensearch-project/OpenSearch/blob/4d3da00c1235fbceb65239336a027936ec781dc2/server/src/main/java/org/opensearch/http/HttpRequest.java#L65-L70

 - `path` and `uri` are the same if there is no query.  The query part gets parsed out to params:
https://github.com/opensearch-project/OpenSearch/blob/4d3da00c1235fbceb65239336a027936ec781dc2/server/src/main/java/org/opensearch/rest/RestRequest.java#L178-L180

So in theory if you *want* the URI including the query string in the Rest Handler, we could add it. But I think it duplicates the path + params combination.  More importantly , "path" is actually what is being registered and looked up here, and using the "uri" name is potentially confusing.

owaiskazi19 (approved):

src/main/java/org/opensearch/sdk/ExtensionRestHandler.java
dbwiddis: Let’s discuss on #165, because both classes are part of the API we expect extension authors to use.  The ExtensionRestRequest is the argument type for the handleRequest method and it closely parallels the same method on plugins and I’d really like similar names.

PR #166, opened 2022-09-30
Refactor Netty4 method

mloufra:
### Description
Create a new class `GetNetty4Transport.java` to handle `getNetty4Transport` and `initializeExtensionTransportService` method for refactoring the class in ExtensionsRunner.java.

### Issues Resolved


src/main/java/org/opensearch/sdk/GetNetty4Transport.java
```diff
@@ -0,0 +1,122 @@
...
+import org.opensearch.transport.netty4.Netty4Transport;
+
+import static java.util.Collections.emptySet;
+import static org.opensearch.common.UUIDs.randomBase64UUID;
+
+/**
+ * This class initializes a Netty4Transport object and control communication between the extension and OpenSearch.
+ */
+
+public class GetNetty4Transport {
```
owaiskazi19, line 42: Let's rename it here and everywhere to `NettyTransport`

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -109,7 +92,10 @@ public class ExtensionsRunner {
      * This field is initialized by a call from {@link ExtensionsInitRequestHandler}.
      */
     public final Settings settings;
-    private final TransportInterceptor NOOP_TRANSPORT_INTERCEPTOR = new TransportInterceptor() {
+    /**
+     * This field is initialized by a call from {@link GetNetty4Transport}.
+     */
+    public final TransportInterceptor NOOP_TRANSPORT_INTERCEPTOR = new TransportInterceptor() {
```
owaiskazi19, line 98: You can move this to the newly created `Netty4Transport`class and make it `private`.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -85,7 +65,10 @@
 public class ExtensionsRunner {
 
     private static final Logger logger = LogManager.getLogger(ExtensionsRunner.class);
-    private static final String NODE_NAME_SETTING = "node.name";
+    /**
+     * This String get a call from {@link GetNetty4Transport}.
+     */
+    public static final String NODE_NAME_SETTING = "node.name";
```
owaiskazi19, line 71: You can move this to the newly created Netty4Transportclass and make it private.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -170,7 +157,7 @@ private ExtensionsRunner(Extension extension) throws IOException {
         // save custom settings
         this.customSettings = extension.getSettings();
         // initialize the transport service
-        this.initializeExtensionTransportService(this.getSettings());
+        this.getNetty4Transport.initializeExtensionTransportService(this.getSettings(), null);
```
owaiskazi19, line 160: This method can be called  like
```
getNetty4Transport.initializeExtensionTransportService(....

Comment (owaiskazi19) [reply src/main/java/org/opensearch/sdk/ExtensionsRunner.java:160]: Why are we passing `null` here? We should avoid it.

File: src/main/java/org/opensearch/sdk/ExtensionsRunner.java [hunk 5]
```diff
@@ -591,7 +501,7 @@ public static void main(String[] args) throws IOException {
         ExtensionsRunner extensionsRunner = new ExtensionsRunner();
 
         // initialize the transport service
-        extensionsRunner.initializeExtensionTransportService(extensionsRunner.getSettings());
+        extensionsRunner.getNetty4Transport.initializeExtensionTransportService(extensionsRunner.getSettings(), extensionsRunner);
```
owaiskazi19, line 504: We should avoid passing the object of `extensionRunner` here directly. Instead, all of the required methods, variables should be moved to `Netty4Transport` file.

src/test/java/org/opensearch/sdk/TestNetty4Transport.java
```diff
@@ -24,12 +24,12 @@
...
 public class TestNetty4Transport extends OpenSearchTestCase {
 
-    private ExtensionsRunner extensionsRunner;
     private ThreadPool threadPool;
+    private GetNetty4Transport getNetty4Transport = new GetNetty4Transport();
 
     @BeforeEach
     public void setUp() throws IOException {
-        this.extensionsRunner = new ExtensionsRunner();
+        new ExtensionsRunner();
```
owaiskazi19, line 32: What are we doing here exactly?

src/main/java/org/opensearch/sdk/GetNetty4Transport.java
```diff
@@ -0,0 +1,122 @@
...
+        NetworkService networkService = new NetworkService(Collections.emptyList());
+        PageCacheRecycler pageCacheRecycler = new PageCacheRecycler(settings);
+        IndicesModule indicesModule = new IndicesModule(Collections.emptyList());
+        SearchModule searchModule = new SearchModule(settings, Collections.emptyList());
+
+        List<NamedWriteableRegistry.Entry> namedWriteables = Stream.of(
+            NetworkModule.getNamedWriteables().stream(),
+            indicesModule.getNamedWriteables().stream(),
+            searchModule.getNamedWriteables().stream(),
+            null,
```
dbwiddis, line 61: No need to include a null in the stream.  Just omit it.  If there's a reason we are excluding something (I think this is where plugins module goes?) just add a comment explaining why it doesn't look like the other calls.

src/main/java/org/opensearch/sdk/GetNetty4Transport.java
```diff
@@ -0,0 +1,122 @@
...
+     */
+    public TransportService initializeExtensionTransportService(Settings settings, ExtensionsRunner extensionsRunner) {
+
+        ThreadPool threadPool = new ThreadPool(settings);
+
+        Netty4Transport transport = getNetty4Transport(settings, threadPool);
+
+        final ConnectionManager connectionManager = new ClusterConnectionManager(settings, transport);
+
+        // Stop any existing transport service
```
dbwiddis, line 98: I added this code block while it was in the ExtensionsRunner class.  Now that it's in its own class we have much better control over its initialization so you can remove these lines.

src/main/java/org/opensearch/sdk/GetNetty4Transport.java
```diff
@@ -0,0 +1,122 @@
...
+        if (extensionsRunner.extensionTransportService != null) {
+            extensionsRunner.extensionTransportService.stop();
+        }
+
+        // create transport service
+        extensionsRunner.extensionTransportService = new TransportService(
```
dbwiddis, line 104: Let's not set a field in another class from here.  This method returns the initialized transport service, so just return it and in the ExtensionsRunner where this method is called, just set the value there.

I realize the method in the old class was awkward setting a variable we already had access to, but I think that was for tests. Moving it to this class should make everything much more self-contained.

dbwiddis (commented):
As @owaiskazi19 said, we don't need to pass an ExtensionsRunner object to this class, as the only thing we really do with it is set the value of the transport service -- which is returned from the method anyway.   I realize the code you're refactoring was already somewhat redundant so it's a bit confusing but this is a good opportunity to clean it up.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
mloufra: > Why are we passing `null` here? We should avoid it.

The `initializeExtensionTransportService` method need to call method `extensionTransportService` from `ExtensionsRunner.java`, so we should have a reference to the object. And it will be `this` instead of `null`.

src/test/java/org/opensearch/sdk/TestNetty4Transport.java
mloufra: Before we start refactor `nettyTransport` method, we call the `nettyTransport` method from `ExtensionsRunner.java`. Since the `getNetty4Transport` moved to a new class, we do not need to call this method from `ExtensionsRunner.java`. So I use IDE delete this line, but the IDE do not help delete whole line, half line has been left. I will delete that in next commit.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -109,15 +88,14 @@ public class ExtensionsRunner {
...
-    private final TransportInterceptor NOOP_TRANSPORT_INTERCEPTOR = new TransportInterceptor() {
-    };
     private ExtensionNamedWriteableRegistry namedWriteableRegistryApi = new ExtensionNamedWriteableRegistry();
     private ExtensionsInitRequestHandler extensionsInitRequestHandler = new ExtensionsInitRequestHandler();
     private OpensearchRequestHandler opensearchRequestHandler = new OpensearchRequestHandler();
     private ExtensionsIndicesModuleRequestHandler extensionsIndicesModuleRequestHandler = new ExtensionsIndicesModuleRequestHandler();
     private ExtensionsIndicesModuleNameRequestHandler extensionsIndicesModuleNameRequestHandler =
         new ExtensionsIndicesModuleNameRequestHandler();
     private ExtensionsRestRequestHandler extensionsRestRequestHandler = new ExtensionsRestRequestHandler();
+    private NettyTransport nettyTransport = new NettyTransport();
```
dbwiddis, line 98: (Suggestion) rather than sending `this` in the initialize call, send `this` in the constructor here and save it as an instance field so you don't have to pass it later.

src/main/java/org/opensearch/sdk/NettyTransport.java
```diff
@@ -0,0 +1,120 @@
...
+            boundAddress -> DiscoveryNode.createLocal(
+                Settings.builder().put(NODE_NAME_SETTING, settings.get(NODE_NAME_SETTING)).build(),
+                boundAddress.publishAddress(),
+                randomBase64UUID()
+            ),
+            null,
+            emptySet()
+        );
+        extensionsRunner.startTransportService(extensionsRunner.extensionTransportService);
+        return extensionsRunner.extensionTransportService;
```
dbwiddis, line 117: (Suggestion) We're never using this return value, could just make this method return `void`.

dbwiddis (approved):
LGTM as a refactor.  Consider the suggestions but don't hold up the PR for them.

I still think it's really confusing having the transportService object being referenced from outside the class, but that's a more complicated discussion we can deal with at another time.

mloufra:
> LGTM as a refactor. Consider the suggestions but don't hold up the PR for them.
> 
> I still think it's really confusing having the transportService object being referenced from outside the class, but that's a more complicated discussion we can deal with at another time.

I will deal with these comment in my next PR.

saratvemulapalli (approved):

saratvemulapalli:
> LGTM as a refactor. Consider the suggestions but don't hold up the PR for them.
> 
> I still think it's really confusing having the transportService object being referenced from outside the class, but that's a more complicated discussion we can deal with at another time.

+1

PR #169, opened 2022-10-07
Return consumed params and content from extensions

### LINKED ISSUE #165: [FEATURE] Move ExtensionRestResponse class to same package as ExtensionRestRequest

As part of #163, the `ExtensionRestRequest` was moved to the OpenSearch project.   It makes sense to also relocate the `ExtensionRestResponse` object there as well for Reasons:
 - Put the request and response together in the same package
 - Enable the header keys to be made public constants and easily read by the `RestSendToExtensionAction` rather than duplicating constants (code smell, error prone).

Move the class and its tests to `org.opensearch.extensions.rest` package.

Leaving it as is and hoping nobody changes one of the constants without forgetting to change the other.

@peternied wants to rename some methods in the class [per this comment](https://github.com/opensearch-project/opensearch-sdk-java/pull/163#discussion_r984757672), we can handle that at the same time:
>  This name is too literal and kind of tripping me up, what about breaking this functionality apart into something like `markParameter(s)Consumed(String([]) param)`, `markContentConsumed()`, and finally `record/reportConsumedItems()`

dbwiddis:
Companion PR: https://github.com/opensearch-project/OpenSearch/pull/4705

Note: Gradle check will fail on this repo until the companion PR is merged due to these changes. Passes locally if companion PR changes are published locally.

### Description
Moves the `ExtensionRestResponse` class from the SDK to `o.o.extensions.rest`.
 - Updates the class to store the consumed parameters and content as fields
 - Updates the `ExtensionRestRequestHandler` to copy the new fields
 - To demo content, added `POST` and `DELETE` methods to the Hello World sample and updated openapi spec
 - Updated tests for the new methods
 - Fixed a few bugs I found along the way

### Issues Resolved
Fixes #165


joshpalis (approved):
LGTM

src/main/java/org/opensearch/sdk/sample/helloworld/rest/RestHelloAction.java
```diff
@@ -66,4 +125,21 @@ private ExtensionRestResponse handleBadRequest(ExtensionRestRequest request) {
         return new ExtensionRestResponse(request, NOT_FOUND, "Extension REST action improperly configured to handle " + request.toString());
     }
 
+    private String parseJsonAdjective(String json) {
+        // TODO: Once CreateComponents has an XContentRegistry available we can parse from there
```
owaiskazi19, line 129: Let's create a follow up issue for this. This will be needed.

owaiskazi19 (approved):
One minor comment

src/main/java/org/opensearch/sdk/sample/helloworld/rest/RestHelloAction.java
dbwiddis: I kind of figured you were already doing that as part of your AD detector and I'd just steal it here :)

dbwiddis: #178

PR #170, opened 2022-10-07
Transport service 

mloufra:
### Description
To avoid passing the object of `extensionRunner` in `ExtensionsRunner` main function, move all of the required methods and variables to `NettyTransport` file
Address comment from [previous PR](https://github.com/opensearch-project/opensearch-sdk-java/pull/166).

### Issues Resolved
https://github.com/opensearch-project/opensearch-sdk-java/pull/166

### DIFF
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -91,8 +91,13 @@ public class ExtensionsRunner {
     private ExtensionsIndicesModuleRequestHandler extensionsIndicesModuleRequestHandler = new ExtensionsIndicesModuleRequestHandler();
     private ExtensionsIndicesModuleNameRequestHandler extensionsIndicesModuleNameRequestHandler =
         new ExtensionsIndicesModuleNameRequestHandler();
+<<<<<<< HEAD
     private ExtensionsRestRequestHandler extensionsRestRequestHandler = new ExtensionsRestRequestHandler(extensionRestPathRegistry);
     private NettyTransport nettyTransport = new NettyTransport();
+=======
+    private ExtensionsRestRequestHandler extensionsRestRequestHandler = new ExtensionsRestRequestHandler();
+    private NettyTransport nettyTransport = new NettyTransport(this);
+>>>>>>> handle transportService object
 
     /*
      * TODO: expose an interface for extension to register actions
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -10,6 +10,7 @@ package org.opensearch.sdk;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.cluster.node.DiscoveryNode;
+import org.opensearch.common.Nullable;
 import org.opensearch.common.io.stream.NamedWriteableRegistryParseRequest;
 import org.opensearch.extensions.OpenSearchRequest;
 import org.opensearch.extensions.rest.ExtensionRestRequest;
@@ -72,6 +73,7 @@ public class ExtensionsRunner {
     /**
      * This field is initialized by a call from {@link ExtensionsInitRequestHandler}.
      */
+    @Nullable
     public TransportService extensionTransportService = null;
     // The routes and classes which handle the REST requests
     private final ExtensionRestPathRegistry extensionRestPathRegistry = new ExtensionRestPathRegistry();
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -139,7 +139,7 @@ public class ExtensionsRunner {
         // save custom transport actions
         this.transportActions = new TransportActions(extension.getActions());
         // initialize the transport service
-        nettyTransport.initializeExtensionTransportService(this.getSettings(), this);
+        nettyTransport.initializeExtensionTransportService(this.getSettings());
     }
 
     /**
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -92,7 +92,6 @@ public class ExtensionsRunner {
     private ExtensionsIndicesModuleNameRequestHandler extensionsIndicesModuleNameRequestHandler =
         new ExtensionsIndicesModuleNameRequestHandler();
     private ExtensionsRestRequestHandler extensionsRestRequestHandler = new ExtensionsRestRequestHandler(extensionRestPathRegistry);
-    private NettyTransport nettyTransport = new NettyTransport(this);
 
     /*
      * TODO: expose an interface for extension to register actions
@@ -131,8 +130,6 @@ public class ExtensionsRunner {
         this.customSettings = extension.getSettings();
         // save custom transport actions
         this.transportActions = new TransportActions(extension.getActions());
-        // initialize the transport service
-        nettyTransport.initializeExtensionTransportService(this.getSettings());
     }
 
     /**
@@ -449,8 +446,9 @@ public class ExtensionsRunner {
      */
     public static void run(Extension extension) throws IOException {
         logger.info("Starting extension " + extension.getExtensionSettings().getExtensionName());
-        @SuppressWarnings("unused")
         ExtensionsRunner runner = new ExtensionsRunner(extension);
+        // initialize the transport service
+        new NettyTransport(runner).initializeExtensionTransportService(runner.getSettings());
         runner.startActionListener(0);
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/sdk/NettyTransport.java
```diff
@@ -40,9 +40,23 @@
...
+     * @param extensionsRunner method to call
+     */
+    public NettyTransport(ExtensionsRunner extensionsRunner) {
+        this.extensionsRunner = extensionsRunner;
+    }
+
+    /**
+     * empty constructor for getNetty4Transport method call
+     */
+    public NettyTransport() {}
```
dbwiddis, line 58: Why is this needed?  Without initializing the `extensionsRunner` you'll end up with an NPE when you try to initialize.

src/main/java/org/opensearch/sdk/NettyTransport.java
```diff
@@ -40,9 +40,23 @@
 
 public class NettyTransport {
     private static final String NODE_NAME_SETTING = "node.name";
+    private ExtensionsRunner extensionsRunner;
```
dbwiddis, line 43: Suggest making this final. We want to guarantee that it's initialized in the constructor, and there will only ever be one runner set.

src/main/java/org/opensearch/sdk/NettyTransport.java
```diff
@@ -85,22 +99,18 @@ public Netty4Transport getNetty4Transport(Settings settings, ThreadPool threadPo
...
 
         ThreadPool threadPool = new ThreadPool(settings);
 
         Netty4Transport transport = getNetty4Transport(settings, threadPool);
 
-        // Stop any existing transport service
-        if (extensionsRunner.extensionTransportService != null) {
-            extensionsRunner.extensionTransportService.stop();
-        }
+        // delete stop
```
dbwiddis, line 110: This looks like a note-to-self to remove, so you can probably remove it.   

Note for other reviewers: I added the stop/start code myself at some point thinking it was needed, but it's not.

src/test/java/org/opensearch/sdk/TransportCommunicationIT.java
```diff
@@ -35,7 +35,9 @@ public class TransportCommunicationIT extends OpenSearchIntegTestCase {
     private final int port = 7777;
     private final String host = "127.0.0.1";
     private volatile String clientResult;
+    private ExtensionsRunner extensionsRunner;
     private NettyTransport nettyTransport = new NettyTransport();
+    private NettyTransport nettyTransport1 = new NettyTransport(extensionsRunner);
```
dbwiddis, line 40: You'll need to initialize the `extensionsRunner` object 2 lines up, otherwise this will fail.  And you should be deleting the no-arg constructor call so you don't need a copy with 1 appended.

dbwiddis (changes requested):

src/main/java/org/opensearch/sdk/NettyTransport.java
mloufra: Before I added this empty constructor in `NettyTransport.java`, The test class `TestNetty4Transport` and `TransportCommunicationIT` shows error message "The constructor NettyTransport() is undefined".

mloufra: The constructor problem is fixed.  Thank @vibrantvarun Varun help me on that.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -78,6 +79,7 @@ public class ExtensionsRunner {
     /**
      * This field is initialized by a call from {@link ExtensionsInitRequestHandler}.
      */
+    @Nullable
```
dbwiddis, line 82: I'm on the fence over whether to use this annotation here.  This field _can_ be null until it's initialized but we don't want it to be.  So it's a good warning that we need to null-check this if we ever read it, but that should primarily be localized to this class.  Interested in other reviewers' opinions here.

I don't think it hurts to have it, I just think it's unnecessary.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -98,7 +100,7 @@ public class ExtensionsRunner {
     private ExtensionsIndicesModuleNameRequestHandler extensionsIndicesModuleNameRequestHandler =
         new ExtensionsIndicesModuleNameRequestHandler();
     private ExtensionsRestRequestHandler extensionsRestRequestHandler = new ExtensionsRestRequestHandler();
-    private NettyTransport nettyTransport = new NettyTransport();
+    private NettyTransport nettyTransport = new NettyTransport(this);
```
dbwiddis, line 103: I really like this pattern for a lot of the refactorings which need to know what the ExtensionsRunner object is.

src/main/java/org/opensearch/sdk/NettyTransport.java
```diff
@@ -40,9 +40,18 @@
 
 public class NettyTransport {
     private static final String NODE_NAME_SETTING = "node.name";
+    private final ExtensionsRunner extensionsRunner;
```
dbwiddis, line 43: Good use of `final` here!

src/main/java/org/opensearch/sdk/NettyTransport.java
```diff
@@ -40,9 +40,18 @@
...
     private final TransportInterceptor NOOP_TRANSPORT_INTERCEPTOR = new TransportInterceptor() {
     };
 
+    /**
+     *
```
dbwiddis, line 48: Nit: whitespace.  

Non-nit: this javadoc isn't very explanatory.  Something like "Instantiate this object with a reference to the ExtensionsRunner."   Also the `@param` tag description doesn't describe the ExtensionsRunner.

src/test/java/org/opensearch/sdk/TestNetty4Transport.java
```diff
@@ -25,11 +25,14 @@
...
 
     private ThreadPool threadPool;
-    private NettyTransport nettyTransport = new NettyTransport();
+    private ExtensionsRunner extensionsRunner;
+    private NettyTransport nettyTransport;
 
     @BeforeEach
     public void setUp() throws IOException {
         this.threadPool = new TestThreadPool("test");
+        this.extensionsRunner = new ExtensionsRunner();
```
dbwiddis, line 34: (Not for this PR) I wonder if there's value in creating a "Test" version of the ExtensionsRunner.  The no-arg constructor here is used exclusively for testing and I'd like to see it removed and replaced with some TestExtensionsRunner class in the test suite.

dbwiddis (changes requested):
Code changes look good.  Just need some doc fixes on the constructor javadoc.

src/test/java/org/opensearch/sdk/TestNetty4Transport.java
dbwiddis: Created #171 for this.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
owaiskazi19: Let's remove this @mloufra!

src/test/java/org/opensearch/sdk/TestNetty4Transport.java
```diff
@@ -25,11 +25,13 @@
 public class TestNetty4Transport extends OpenSearchTestCase {
 
     private ThreadPool threadPool;
-    private NettyTransport nettyTransport = new NettyTransport();
+    private ExtensionsRunner extensionsRunner;
```
owaiskazi19, line 28: Can we use `TestExtensionsRunner` here @dbwiddis?

dbwiddis: Yes, we can, our problem here is synchronizing all the tests so they aren't all using the transport service associated with it.  I've suggested to @mloufra to use [`@ResourceLock`](https://junit.org/junit5/docs/5.3.0/api/org/junit/jupiter/api/parallel/ResourceLock.html).

mloufra: > Yes, we can, our problem here is synchronizing all the tests so they aren't all using the transport service associated with it. I've suggested to @mloufra to use [`@ResourceLock`](https://junit.org/junit5/docs/5.3.0/api/org/junit/jupiter/api/parallel/ResourceLock.html).

Since the error message is from `TransportCommunicationIT`, `TestNetty4Transport` and `TestExtensionsRunner`, I added `@ResourceLock("transportService")` in each test class, but still cause same error.

dbwiddis (dismissed):
Fixed the bind errors by moving the netty initialization out of the constructor in the Extensions Runner and backed out the "hack" solution from intervening PR.  The rest is what was ready to go before this weekend and is GTG now! :)

ryanbogan:
That method should take a threadPool parameter too

ryanbogan:
I can't access the branch with the changes for some reason.  The only branch I see on @mloufra 's fork is main.

mloufra:
> I can't access the branch with the changes for some reason. The only branch I see on @mloufra 's fork is main.

I fork our sdk repo again this morning. That's why only main branch in my fork repo.

owaiskazi19:
> I fork our sdk repo again this morning. That's why only main branch in my fork repo.
@mloufra you should close this PR and raise again with the new fork repo as you won't be able to access `transportService` branch anymore on your fork.

mloufra:
> > I fork our sdk repo again this morning. That's why only main branch in my fork repo.
> 
> @mloufra you should close this PR and raise again with the new fork repo as you won't be able to access `transportService` branch anymore on your fork.

Thanks.

PR #172, opened 2022-10-09
Refactor ExtensionsRunner test code out of production source tree

### LINKED ISSUE #171: [FEATURE] Refactor ExtensionsRunner test code out of production source tree

We shouldn't have exclusively test-related code in the main source tree, e.g.,
https://github.com/opensearch-project/opensearch-sdk-java/blob/8fbcc6eaf52e6b00f8bc93e3f4256ccf0d8e794e/src/main/java/org/opensearch/sdk/ExtensionsRunner.java#L117-L123

which calls here, only by tests:
https://github.com/opensearch-project/opensearch-sdk-java/blob/8fbcc6eaf52e6b00f8bc93e3f4256ccf0d8e794e/src/main/java/org/opensearch/sdk/ExtensionsRunner.java#L159-L163

which uses this constant pointing to the test source tree:
https://github.com/opensearch-project/opensearch-sdk-java/blob/8fbcc6eaf52e6b00f8bc93e3f4256ccf0d8e794e/src/main/java/org/opensearch/sdk/ExtensionSettings.java#L19-L22

1. Remove the no-arg constructor above from `ExtensionsRunner`
2. Make the constructor with an `Extension` argument either package-private (probably) or protected (maybe) so it's accessible by tests
3. Create a new subclass of `ExtensionsRunner` in the `src/test/java` tree to be used as an `ExtensionsRunner` object when needed for tests, such as the linked comment in the context section below.   Move the above code snippets there with modifications as noted below.
4. Update all tests which use the existing no-arg constructor to use the new class.

One would think `TestExtensionsRunner` would be a good name but it's already in use for our test class.  Picking a new name will probably be the hardest part.  It can extend `ExtensionsRunner` like so:
```java
public class ExtensionsRunnerForTesting extends ExtensionsRunner {

    public ExtensionsRunnerForTesting() throws IOException {
        super(new Extension () {

            @Override
            public ExtensionSettings getExtensionSettings() {
                // Move the result of current readExtensionSettings() method of ExtensionsRunner here
                // except get rid of yaml file reading, and just instantiate the object with the values in the yaml
                return new ExtensionSettings( ... );
            }

            @Override
            public List<ExtensionRestHandler> getExtensionRestHandlers() {
            }
        });
    }
}
```

We could use the "sample" extensions for this purpose, or even create a test extension in the sample branch. This seems reasonable for an IT-based approach, but I think this approach is better for unit testing requiring an object with minimal settings.

> I wonder if there's value in creating a "Test" version of the ExtensionsRunner.  The no-arg constructor here is used exclusively for testing and I'd like to see it removed and replaced with some TestExtensionsRunner class in the test suite.
> 
> _Originally posted by @dbwiddis in https://github.com/opensearch-project/opensearch-sdk-java/pull/170#discussion_r990678239_

roulpriya:
### Description
Remove `ExtensionsRunner` constructor which was used only in tests

### Issues Resolved
Closes #171 

### DIFF
--- a/src/test/java/org/opensearch/sdk/ExtensionsRunnerForTest.java
+++ b/src/test/java/org/opensearch/sdk/ExtensionsRunnerForTest.java
@@ -1,0 +1,32 @@
+package org.opensearch.sdk;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * An Extension Runner for testing using test settings.
+ */
+public class ExtensionsRunnerForTest extends ExtensionsRunner {
+
+    /**
+     * Instantiates a new Extensions Runner using test settings.
+     *
+     * @throws IOException if the runner failed to read settings or API.
+     */
+    public ExtensionsRunnerForTest() throws IOException {
+        super(new Extension() {
+            @Override
+            public ExtensionSettings getExtensionSettings() {
+                return new ExtensionSettings("sample-extension", "127.0.0.1", "4532");
+            }
+
+            @Override
+            public List<ExtensionRestHandler> getExtensionRestHandlers() {
+                return Collections.emptyList();
+            }
+        });
+
+    }
+
+}
--- a/src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
+++ b/src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
@@ -31,6 +31,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.opensearch.Version;
@@ -77,11 +78,14 @@ public class TestExtensionsRunner extends OpenSearchTestCase {
     private ExtensionsRunner extensionsRunner;
     private TransportService transportService;
 
+    private TransportService initialTransportService;
+
     @Override
     @BeforeEach
     public void setUp() throws Exception {
         super.setUp();
-        this.extensionsRunner = new ExtensionsRunner();
+        this.extensionsRunner = new ExtensionsRunnerForTest();
+        this.initialTransportService = extensionsRunner.extensionTransportService;
         this.transportService = spy(
             new TransportService(
                 Settings.EMPTY,
@@ -93,6 +97,17 @@ public class TestExtensionsRunner extends OpenSearchTestCase {
                 Collections.emptySet()
             )
         );
+    }
+
+    @Override
+    @AfterEach
+    public void tearDown() throws Exception {
+        super.tearDown();
+        if (initialTransportService != null) {
+            this.initialTransportService.stop();
+            this.initialTransportService.close();
+            Thread.sleep(1000);
+        }
     }
 
     // test manager method invokes start on transport service
--- a/src/test/java/org/opensearch/sdk/TransportCommunicationIT.java
+++ b/src/test/java/org/opensearch/sdk/TransportCommunicationIT.java
@@ -145,7 +145,7 @@ public class TransportCommunicationIT extends OpenSearchIntegTestCase {
     private void startTransportandClient(Settings settings, Thread client) throws IOException, InterruptedException {
 
         // retrieve transport service
-        ExtensionsRunner extensionsRunner = new ExtensionsRunner();
+        ExtensionsRunner extensionsRunner = new ExtensionsRunnerForTest();
         // start transport service
         TransportService transportService = nettyTransport.initializeExtensionTransportService(settings, extensionsRunner);
 
--- a/src/main/java/org/opensearch/sdk/ExtensionSettings.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionSettings.java
@@ -17,11 +17,6 @@ public class ExtensionSettings {
     private String hostPort;
 
     /**
-     * Path to extension.yml file for testing. Internal use only.
-     */
-    static final String EXTENSION_DESCRIPTOR = "src/test/resources/extension.yml";
-
-    /**
      * Jackson requires a default constructor.
      */
     @SuppressWarnings("unused")
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -7,9 +7,6 @@
  */
 package org.opensearch.sdk;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
-
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.cluster.node.DiscoveryNode;
@@ -50,10 +47,8 @@ import org.opensearch.transport.TransportResponseHandler;
 import org.opensearch.transport.TransportService;
 import org.opensearch.transport.TransportSettings;
 
-import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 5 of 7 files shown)


src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -81,7 +81,7 @@ public class TestExtensionsRunner extends OpenSearchTestCase {
     @BeforeEach
     public void setUp() throws Exception {
         super.setUp();
-        this.extensionsRunner = new ExtensionsRunner();
+        this.extensionsRunner = new ExtensionsRunnerForTest();
```
roulpriya, line 84: At this point the test blocks
On digging in `ExtensionsRunner`, it tries to start `ActionListener` with timeout 0 which further has a `Thread.join()` that blocks the calling thread. 
https://github.com/opensearch-project/opensearch-sdk-java/blob/8fbcc6eaf52e6b00f8bc93e3f4256ccf0d8e794e/src/main/java/org/opensearch/sdk/ExtensionsRunner.java#L156

https://github.com/opensearch-project/opensearch-sdk-java/blob/8fbcc6eaf52e6b00f8bc93e3f4256ccf0d8e794e/src/main/java/org/opensearch/sdk/ActionListener.java#L79

dbwiddis: > At this point the test blocks

Yeah, that's the one that just starts it and leaves it running for testing/debug purposes.  Put a nonzero timeout in there for tests.

build.gradle
```diff
@@ -48,7 +48,7 @@ publishing {
 repositories {
     mavenLocal()
     // Remove the commented code below once TransportService is published to maven
-    //maven { url "https://aws.oss.sonatype.org/content/repositories/snapshots" }
+    maven { url "https://aws.oss.sonatype.org/content/repositories/snapshots" }
```
dbwiddis, line 51: You should be able to back this out now that you are working off the feature/extensions published locally.

src/test/java/org/opensearch/sdk/ExtensionsRunnerForTest.java
```diff
@@ -0,0 +1,34 @@
...
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * An Extension Runner for testing using test settings.
+ */
+public class ExtensionsRunnerForTest extends ExtensionsRunner {
+
+    private static final String EXTENSION_DESCRIPTOR = "src/test/resources/extension.yml";
```
dbwiddis, line 12: We'll actually want to remove this. It's further down in the test extension settings.

dbwiddis (commented):
Looking good so far!

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
dbwiddis: Also the current `build.gradle` "run" will direct to the main method you just deleted.  Perhaps that "ExtensionsRunnerForTesting" should have a main method to be runnable there...

roulpriya: If I put a `Main` method in `ExtensionsRunnerForTests` we can't set it as `mainClassName` in gradle because it is part of test sources.

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -77,11 +78,14 @@ public class TestExtensionsRunner extends OpenSearchTestCase {
...
 
+    private TransportService initialTransportService;
+
     @Override
     @BeforeEach
     public void setUp() throws Exception {
         super.setUp();
-        this.extensionsRunner = new ExtensionsRunner();
+        this.extensionsRunner = new ExtensionsRunnerForTest();
+        this.initialTransportService = extensionsRunner.extensionTransportService;
```
roulpriya, line 88: The constructor of `ExtensionRunner` creates an instance of `TransportService` which listens a port in the background.

Tests further use `setExtensionTransportService` to replace the created instance with a mock to use in testing. The original instance becomes unreachable and our `tearDown` cannot stop it, which causes the rest of the tests to fail due to the port being already in use. Hence, this hack.

IMO, we should pass `TransportService` as a constructor parameter and not instantiate inside.

dbwiddis: At this point, let's "hack" as needed to get this merged.

I agree with you and have questioned how we've tracked the transport service from some of the initial PRs. I'm not sure putting it in the constructor is the best solution but we do need something better than what we have.

We do have a future requirement to gracefully handle disconnections/reconnections (hot plugin, etc.) and I think handling these will definitely need to be part of that functionality as well.

build.gradle
```diff
@@ -48,7 +48,7 @@ publishing {
 repositories {
     mavenLocal()
     // Remove the commented code below once TransportService is published to maven
-    //maven { url "https://aws.oss.sonatype.org/content/repositories/snapshots" }
+//    maven { url "https://aws.oss.sonatype.org/content/repositories/snapshots" }
```
dbwiddis, line 51: At this point all changes to this file are whitespace. Would prefer if you can just revert this to the state before your PR.

dbwiddis (changes requested):
Code changes look good! Thank you for your contribution and your patience in dealing with our early-stage code!

Please undo the whitespace changes in build.gradle and add the DCO signoff and I'll approve this.

build.gradle
roulpriya: Done

dbwiddis (dismissed):

joshpalis (approved):

src/test/java/org/opensearch/sdk/TransportCommunicationIT.java
```diff
@@ -145,7 +145,7 @@ public void run() {
     private void startTransportandClient(Settings settings, Thread client) throws IOException, InterruptedException {
 
         // retrieve transport service
-        ExtensionsRunner extensionsRunner = new ExtensionsRunner();
+        ExtensionsRunner extensionsRunner = new ExtensionsRunnerForTest();
```
owaiskazi19, line 148: Should we use `ExtensionsRunnerForTest` for other occurrences in tests as well?
Like:
https://github.com/opensearch-project/opensearch-sdk-java/blob/main/src/test/java/org/opensearch/sdk/TestNamedWriteableRegistryAPI.java#L138

dbwiddis: @owaiskazi19 that particular reference is to the class which remains valid as a superclass. And that particular test is just choosing a class that we know is not a named writeable.  So I think that one is fine. Are there others?  The rest of the PR should have broken any no-arg constructors.

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -95,6 +99,17 @@ public void setUp() throws Exception {
         );
     }
 
+    @Override
+    @AfterEach
+    public void tearDown() throws Exception {
+        super.tearDown();
```
owaiskazi19, line 105: Thanks for adding the `tearDown`. This will definitely avoid the bind connection error.

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -77,11 +78,14 @@ public class TestExtensionsRunner extends OpenSearchTestCase {
     private ExtensionsRunner extensionsRunner;
     private TransportService transportService;
 
+    private TransportService initialTransportService;
```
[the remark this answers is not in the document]
dbwiddis: I'm already drafting a Proposal for this that will cover more use cases. Stay tuned.

dbwiddis (approved):

joshpalis (approved):

src/test/java/org/opensearch/sdk/ExtensionsRunnerForTest.java
```diff
@@ -0,0 +1,32 @@
+package org.opensearch.sdk;
```
saratvemulapalli, line 1: This is great!
It sets up our initial framework for tests going forward.

saratvemulapalli (approved):
Thanks @roulpriya for the changes!

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
dbwiddis: Opened #175 to address this.

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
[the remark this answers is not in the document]
dbwiddis: See #174 to follow up on this.

PR #179, opened 2022-10-10
[Feature/extensions] Modified Environment settings support

joshpalis:
### Description
Environment Settings support for extensibility will now pass the entire Settings object from OpenSearch node environment to the SDK.

Companion PR : https://github.com/opensearch-project/OpenSearch/pull/4731

### Issues Resolved
https://github.com/opensearch-project/opensearch-sdk-java/issues/78

### DIFF
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -50,6 +50,8 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 
 /**
@@ -370,14 +372,27 @@ public class ExtensionsRunner {
      * Requests the environment settings from OpenSearch. The result will be handled by a {@link EnvironmentSettingsResponseHandler}.
      *
      * @param transportService  The TransportService defining the connection to OpenSearch.
+     * @return A Setting object from the OpenSearch Node environment
      */
-    public void sendEnvironmentSettingsRequest(TransportService transportService) {
-        sendGenericRequestWithExceptionHandling(
-            transportService,
-            ExtensionsOrchestrator.RequestType.REQUEST_EXTENSION_ENVIRONMENT_SETTINGS,
-            ExtensionsOrchestrator.REQUEST_EXTENSION_ENVIRONMENT_SETTINGS,
-            new EnvironmentSettingsResponseHandler()
-        );
+    public Settings sendEnvironmentSettingsRequest(TransportService transportService) {
+        final CountDownLatch inProgressLatch = new CountDownLatch(1);
+        logger.info("Sending Environment Settings request to OpenSearch");
+        EnvironmentSettingsResponseHandler environmentSettingsResponseHandler = new EnvironmentSettingsResponseHandler(inProgressLatch);
+        try {
+            transportService.sendRequest(
+                opensearchNode,
+                ExtensionsOrchestrator.REQUEST_EXTENSION_ENVIRONMENT_SETTINGS,
+                new ExtensionRequest(ExtensionsOrchestrator.RequestType.REQUEST_EXTENSION_ENVIRONMENT_SETTINGS),
+                environmentSettingsResponseHandler
+            );
+            // Wait on environment settings response
+            inProgressLatch.await(100, TimeUnit.SECONDS);
+        } catch (Exception e) {
+            logger.info("Failed to send Environment Settings request to OpenSearch", e);
+        }
+
+        // At this point, response handler has read in the environment settings
+        return environmentSettingsResponseHandler.getEnvironmentSettings();
     }
 
     /**
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -386,7 +386,9 @@ public class ExtensionsRunner {
                 environmentSettingsResponseHandler
             );
             // Wait on environment settings response
-            inProgressLatch.await(100, TimeUnit.SECONDS);
+            inProgressLatch.await(ExtensionsOrchestrator.EXTENSION_REQUEST_WAIT_TIMEOUT, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            logger.info("Failed to recieve Environment Settings response from OpenSearch", e);
         } catch (Exception e) {
             logger.info("Failed to send Environment Settings request to OpenSearch", e);
         }
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -20,7 +20,6 @@ import org.opensearch.common.settings.Settings;
 import org.opensearch.discovery.InitializeExtensionsRequest;
 import org.opensearch.extensions.ExtensionActionListenerOnFailureRequest;
 import org.opensearch.extensions.DiscoveryExtension;
-import org.opensearch.extensions.EnvironmentSettingsRequest;
 import org.opensearch.extensions.AddSettingsUpdateConsumerRequest;
 import org.opensearch.extensions.UpdateSettingsRequest;
 import org.opensearch.extensions.ExtensionsOrchestrator.RequestType;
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -368,24 +369,30 @@ public void sendActionListenerOnFailureRequest(TransportService transportService
...
     /**
-     * Requests the environment setting values from OpenSearch for the corresponding component settings. The result will be handled by a {@link EnvironmentSettingsResponseHandler}.
+     * Requests the environment settings from OpenSearch. The result will be handled by a {@link EnvironmentSettingsResponseHandler}.
      *
-     * @param componentSettings The component setting that correspond to the values provided by the environment settings
      * @param transportService  The TransportService defining the connection to OpenSearch.
+     * @return A Setting object from the OpenSearch Node environment
      */
-    public void sendEnvironmentSettingsRequest(TransportService transportService, List<Setting<?>> componentSettings) {
+    public Settings sendEnvironmentSettingsRequest(TransportService transportService) {
```
dbwiddis, line 377: When is this called? If we're starting up the extension before OpenSearch, there are no settings to fetch.  Should this be triggered as part of initialization?

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -368,24 +369,30 @@ public void sendActionListenerOnFailureRequest(TransportService transportService
...
         try {
             transportService.sendRequest(
                 opensearchNode,
                 ExtensionsOrchestrator.REQUEST_EXTENSION_ENVIRONMENT_SETTINGS,
-                new EnvironmentSettingsRequest(componentSettings),
+                new ExtensionRequest(ExtensionsOrchestrator.RequestType.REQUEST_EXTENSION_ENVIRONMENT_SETTINGS),
                 environmentSettingsResponseHandler
             );
+            // Wait on environment settings response
+            inProgressLatch.await(100, TimeUnit.SECONDS);
```
dbwiddis, line 389: This seems like a really long time to wait.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -368,24 +369,30 @@ public void sendActionListenerOnFailureRequest(TransportService transportService
...
         } catch (Exception e) {
             logger.info("Failed to send Environment Settings request to OpenSearch", e);
         }
```
dbwiddis, line 392: Since you're adding an `await()` you should explicitly handle the `InterruptedException`.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -368,24 +369,30 @@ public void sendActionListenerOnFailureRequest(TransportService transportService
...
+
+        // At this point, response handler has read in the environment settings
+        return environmentSettingsResponseHandler.getEnvironmentSettings();
```
dbwiddis, line 395: What is the value of this on failure or interrupt?

src/main/java/org/opensearch/sdk/handlers/EnvironmentSettingsResponseHandler.java
```diff
@@ -11,27 +11,43 @@
...
+import java.util.concurrent.CountDownLatch;
 
 /**
  * This class handles the response from OpenSearch to a {@link ExtensionsRunner#sendEnvironmentSettingsRequest} call.
  */
 public class EnvironmentSettingsResponseHandler implements TransportResponseHandler<EnvironmentSettingsResponse> {
     private static final Logger logger = LogManager.getLogger(EnvironmentSettingsResponseHandler.class);
 
+    private CountDownLatch inProgressLatch;
+    private Settings environmentSettings;
```
dbwiddis, line 30: This should be initialized to something (probably `Settings.EMPTY`) to return in case you get an exception or we are interrupted.

src/main/java/org/opensearch/sdk/handlers/EnvironmentSettingsResponseHandler.java
```diff
@@ -11,27 +11,43 @@
...
 import java.io.IOException;
+import java.util.concurrent.CountDownLatch;
 
 /**
  * This class handles the response from OpenSearch to a {@link ExtensionsRunner#sendEnvironmentSettingsRequest} call.
  */
 public class EnvironmentSettingsResponseHandler implements TransportResponseHandler<EnvironmentSettingsResponse> {
     private static final Logger logger = LogManager.getLogger(EnvironmentSettingsResponseHandler.class);
 
+    private CountDownLatch inProgressLatch;
```
dbwiddis, line 29: Make this `final`.

src/main/java/org/opensearch/sdk/handlers/EnvironmentSettingsResponseHandler.java
```diff
@@ -11,27 +11,43 @@
...
+    private Settings environmentSettings;
+
+
+    public EnvironmentSettingsResponseHandler(CountDownLatch inProgressLatch) {
+        this.inProgressLatch = inProgressLatch;
+        this.environmentSettings = null;
```
dbwiddis, line 35: Initializing things to null in a constructor just seems wrong to me.  If you want to initialize without reference to a param in the constructor, just initialize the field directly.   But I don't think this should be null, it should be an empty settings object.

src/main/java/org/opensearch/sdk/handlers/EnvironmentSettingsResponseHandler.java
```diff
@@ -43,4 +59,8 @@ public String executor() {
     public EnvironmentSettingsResponse read(StreamInput in) throws IOException {
         return new EnvironmentSettingsResponse(in);
     }
+
+    public Settings getEnvironmentSettings() {
```
dbwiddis, line 63: This feels really weird having a handler with a getter, but I don't see any other easy way to do it at the moment.

dbwiddis (changes requested):
Need better handling of non-happy-path.

src/main/java/org/opensearch/sdk/handlers/EnvironmentSettingsResponseHandler.java
joshpalis: Sure, Ill make this change

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
joshpalis: I've modified this timeout value to use the `EXTENSION_REQUEST_WAIT_TIMEOUT` value defined in ExtensionOrchestrator

src/main/java/org/opensearch/sdk/handlers/EnvironmentSettingsResponseHandler.java
```diff
@@ -11,27 +11,42 @@
...
  * This class handles the response from OpenSearch to a {@link ExtensionsRunner#sendEnvironmentSettingsRequest} call.
  */
 public class EnvironmentSettingsResponseHandler implements TransportResponseHandler<EnvironmentSettingsResponse> {
     private static final Logger logger = LogManager.getLogger(EnvironmentSettingsResponseHandler.class);
 
+    private final CountDownLatch inProgressLatch;
+    private Settings environmentSettings;
+
+    public EnvironmentSettingsResponseHandler(CountDownLatch inProgressLatch) {
+        this.inProgressLatch = inProgressLatch;
```
owaiskazi19, line 33: We should not pass latch as a part of constructor. It should be use where we are initializing the object for this class

joshpalis: Sure ill make the modification

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
joshpalis: With my latest modifications the value of this on failure would just be an empty settings object

src/main/java/org/opensearch/sdk/handlers/EnvironmentSettingsResponseHandler.java
dbwiddis: @owaiskazi19 do you mean pass it in the request?

owaiskazi19: I meant outside this class where we are creating the object of this class or in the class itself to make it independent

dbwiddis (dismissed):
LGTM!

ryanbogan (approved):

owaiskazi19 (approved):

owaiskazi19 (approved):

ryanbogan (approved):

dbwiddis:
Hey @joshpalis I just noticed the following in some gradle checks:

> /home/runner/work/opensearch-sdk-java/opensearch-sdk-java/src/main/java/org/opensearch/sdk/handlers/EnvironmentSettingsResponseHandler.java:70: warning: no `@throws` for java.lang.InterruptedException

Can you please add that `@throws` in one of your next PRs?

joshpalis:
@dbwiddis Sure thanks for catching that, I'll make sure to add this

Issue #110, opened 2022-08-29
[FEATURE] Configure spotless to handle newline linting in lieu of linelint


dbwiddis:
### Is your feature request related to a problem?

Issue #46 requested to add `linelint` to this repo matching [OS #2857](https://github.com/opensearch-project/OpenSearch/issues/2857).  I commented that it could have been done with spotless, but there was no further discussion and `linelint` was implemented as a GitHub Action.

This presents a frustrating situation for users, as `linelint` does not automatically fix errors. (The actual program does but this is not helpful from a GHA.). There is no real way to proactively lint files from the command line, or even check them, prior to submitting a PR, seeing a failing lint check, and then manually fixing the affected files.

I would like to re-raise the issue of using a single linter (spotless) which **not only detects but auto-fixes these issues**.  

Contrary to the comments on [OS #2857](https://github.com/opensearch-project/OpenSearch/issues/2857), it can handle all sorts of files and is not limited to just `.java` files.  [See for example here](https://github.com/diffplug/spotless/blob/main/plugin-gradle/README.md#quickstart) where `.gradle` and `.md` (and other) files are included:
```gradle
  format 'misc', {
    // define the files to apply `misc` to
    target '*.gradle', '*.md', '.gitignore'

    // define the steps to apply to those files
    trimTrailingWhitespace()
    indentWithTabs() // or spaces. Takes an integer argument if you don't like 4
    endWithNewline()
  }
```

### What solution would you like?
Remove linelint and configure spotless as above (but with spaces, not tabs).

Consider making the same change in OpenSearch for the same reasons and with a much wider impact.

### What alternatives have you considered?
Building linelint on my local machine so I can take advantage of its auto-fix features.  I already tried to install with `homebrew` but it's not set up there and the [effort vs. time saved](https://xkcd.com/1205/) isn't worth building it myself on multiple machines; or even if it was, that would only help me.

### Do you have any additional context?
I plan on adding more `.svg` files documenting designs, and they don't include newlines by default.  I don't want to manually edit `.svg` files.  I don't want to upload `.svg` files and get yelled at by failed GHA. I don't want to keep adding exclusions to `linelint.yml`.   I just want to run `./gradlew spotlessApply` which I already do, and not have to think about end of file newlines ever again.

owaiskazi19:
@peternied do we have any advantage of using the current line linter as compared to Spotless `linelint`? If not, then we are good to go with Spotless :)

[...]

PR #193, opened 2022-10-13
Configure spotless to cleanup misc files

dbwiddis:

### Description
It was faster for me to configure spotless to add newlines to the openapi files I keep exporting than to figure out which one needed me to open the editor and add a newline to.

### Issues Resolved
This addresses half of #110 and makes the linelint check redundant.  

Intentionally committing with one failing openapi file to test whether gradle check fails on spotless; if so, we can remove the linelint check entirely.


.github/workflows/code-hygiene.yml
```diff
@@ -1,14 +1,13 @@
...
-    name: Check if all files end in newline
+    name: Spotless Check
     steps:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Linelint
-        uses: fernandrone/linelint@0.0.4
+      - name: Spotless Check
+        run: ./gradlew spotlessCheck
```
owaiskazi19, line 13: Spotless is run with precommit. We can remove this file

dbwiddis: We don’t have a precommit workflow on SDK.

owaiskazi19: Ah. I meant gradle check. Also, get confused with sdk and OpenSearch repo. My bad

dbwiddis: The point of this setup is to do a “fail fast” on spotless before calling gradle check which takes several minutes before getting that far.

.github/workflows/build.yml
```diff
@@ -1,9 +1,14 @@
 name: Gradle Build
-on: [push, pull_request]
+on:
+  workflow_run:
```
owaiskazi19, line 3: Trying to understand here. Shouldn't this workflow run on `push,pull`? Did we test that

.github/workflows/code-hygiene.yml
owaiskazi19: Makes sense!

.github/workflows/build.yml
dbwiddis: Probably need to give a bit more context here.  First, take a glance at issue [#4174](https://github.com/opensearch-project/OpenSearch/issues/4174) at the main repo.  I'm trying to replicate the same thing here.  My first comment on that issue references [these GitHub docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run).

Background: Jenkins gradle check is slow and takes a lot of resources.  It's not as slow as on main project (yet) but we are just getting started and once we add ITs it will probably get slower.  We are already pulling down the `feature/extensions` branch and publishing it on every run.  It's a waste to do that if the check is going to fail on spotless!

Goal:
 - Run one "fast" workflow that checks some of the obvious failures first (spotless, possibly add javadoc?)
 - Only if that check passes, run the full gradle check on Jenkins

Now to this specific change.  The `on` section indicates what triggers the workflow.  Previously it was either a `push` to main or a pull request.  The way that change is written, it will no longer run on every push or pull request it will only run on a `completed` `workflow_run` of the workflow named "Code Hygiene".   Later on it will also check the result of that workflow.

The Code Hygiene workflow is set to run on `push` and `pull_request` so it will trigger first, and when it completes, trigger this one.

It would be nice if you could see that workflow work on this PR, but for security reasons, the workflow_run only triggers on the default branch, so this will have to be merged for you to watch what happens.

We could get a little fancier and put `push` back on this workflow, and add an "or" conditional on line 11 below to run if it's a push workflow or if the workflow_run was a success.

Or we could be simple and always run gradle check and try to order the checks to fail fast (e.g., run linters BEFORE all the building activity, but I'm not sure they work without compiling so that still ends up taking a while.)

Or we could just always run the gradle check no matter how long it takes.

owaiskazi19: Thanks for answering the question here. I see the design you are following here and I'm totally in line with the same. Just a small ask which you mentioned before already can we have `javadocs` with spotless because we have seen a couple of instance where gradle check fails because of `javadocs`. So quoting you again "fail fast". 😁

owaiskazi19: Also, we should remove the linelint from OpenSearch repo too.

owaiskazi19 (dismissed):
Thanks for making our CI faster @dbwiddis!

saratvemulapalli (approved):

joshpalis (approved):

dbwiddis:
@saratvemulapalli  @owaiskazi19 @joshpalis 
Demo:  after merge, only running these CI:

Screen Shot 2022-10-14 at 11 54 11 AM

Following completion of Code Hygeine, kicked off Gradle Build

Screen Shot 2022-10-14 at 11 57 11 AM

Screen Shot 2022-10-14 at 11 57 49 AM

Issue #130, opened 2022-09-08
Clean up compiler warnings


dbwiddis:
Some code on `main` is producing compiler warnings and should be cleaned up so I don't see so many ⚠️s in my IDE.
 - [x] TransportActions.java `Class` on line 29 needs type arguments. (`Class<? extends TransportAction<? extends ActionRequest, ? extends ActionResponse>>>`).  This will need a simultaneous change to OpenSearch project `feature/extensions` branch `RegisterTransportActionsRequest` constructor (and associated instance variable).
 - [x] TransportAction.java `HashMap` on line 38 needs a diamond operator (`<>`)
 - [ ] NamedWriteableRegistryApi.java lines 84 and 107 need generic types for the `Class` (probably `<? extends NamedWritable>`). This will need a simultaneous change to OpenSearch project `feature/extensions` branch `NamedWriteable RegistryParseRequest` to change the type of `categoryClass` to add the same generics.
 - [x] SDKClient.java has an unused `logger`. Either log something or remove it.
 - [x] TestNamedWriteable.java has an unused inner class constructor `Example(String message)`.  I suspect this means there's missing tests somewhere.
 - [x] Unused field `extensionsRunner` in `NettyTransport` (probably the result of a merge conflict resolution)

dbwiddis:
FYI, just noticed this in gradle check logs:
> Note: /home/runner/work/opensearch-sdk-java/opensearch-sdk-java/src/main/java/org/opensearch/sdk/ExtensionNamedWriteableRegistry.java uses unchecked or unsafe operations.

I believe that fixing this issue will resolve this warning.

mloufra:
> * [ ]  TransportActions.java `Class` on line 29 needs type arguments. (`Class<? extends TransportAction<? extends ActionRequest, ? extends ActionResponse>>>`).  This will need a simultaneous change to OpenSearch project `feature/extensions` branch `RegisterTransportActionsRequest` constructor (and associated instance variable).

Can we use `@SuppressWarnings(“rawtypes”)` to ignore compiler warning for clean up purpose?

owaiskazi19:
> Can we use `@SuppressWarnings(“rawtypes”)` to ignore compiler warning for clean up purpose?

Hey @mloufra! SuppressWarnings is used to ignore the warnings. We don't have to ignore them but to resolve them. We need to add a type argument by making the change in OpenSearch repo's `feature/extensions` branch `RegisterTransportActionsRequest` constructor.

[...]

PR #195, opened 2022-10-14
Compiler warning

mloufra:

### Description
Add Add unbounded wildcard for `transportActions` in `TransportActions.java` to clean up compiler warning.
Equivalent PR on OpenSearch:  https://github.com/opensearch-project/OpenSearch/pull/4796
### Issues Resolved
https://github.com/opensearch-project/opensearch-sdk-java/issues/130

### DIFF
--- a/src/main/java/org/opensearch/sdk/TransportActions.java
+++ b/src/main/java/org/opensearch/sdk/TransportActions.java
@@ -26,7 +26,7 @@ import java.util.Map;
  */
 public class TransportActions {
     private final Logger logger = LogManager.getLogger(TransportActions.class);
-    private Map<String, Class> transportActions;
+    private Map<String, Class<?>> transportActions;
 
     /**
      * Constructor for TransportActions. Creates a map of transportActions for this extension.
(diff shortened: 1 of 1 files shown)


src/main/java/org/opensearch/sdk/TransportActions.java
```diff
@@ -26,7 +26,7 @@
  */
 public class TransportActions {
     private final Logger logger = LogManager.getLogger(TransportActions.class);
-    private Map<String, Class> transportActions;
+    private Map<String, Class<?>> transportActions;
```
dbwiddis, line 29: While `<?>` makes the warning go away, it still means "any class".  

However, we have more information about the class type that will be assigned.   See the constructor below:
```
Map<String, Class<? extends TransportAction<Request, Response>>> transportActions
```
So that is the generic we should use.

Also note in the constructor those `Request` and `Response` are further bounded:
```
<Request extends ActionRequest, Response extends ActionResponse>
```
So this definition here should be:
```java
Class<? extends TransportAction<? extends ActionRequest, ? extends ActionResponse>>
```

dbwiddis (changes requested):

src/main/java/org/opensearch/sdk/TransportActions.java
mloufra: Thanks @dbwiddis

owaiskazi19:
@mloufra can you rebase your branch with latest main and push the changes?

dbwiddis (approved):

saratvemulapalli (approved):

dbwiddis:
Awaiting merging of companion PR

PR #202, opened 2022-10-20
Modified local node request to return discovery node directly to caller

joshpalis:
Signed-off-by: Joshua Palis <jpalis@amazon.com>

### Description
Modified local node request to return discovery node directly to caller

Edit : Separate local Node Requests are not necessary to provide support for since the local node is already transported to the SDK during node initialization. There is already a getter for the opensearch node (source node) used for testing, and this can be modified in order to access this within an extension

Companion PR : https://github.com/opensearch-project/OpenSearch/pull/4862


src/main/java/org/opensearch/sdk/handlers/LocalNodeResponseHandler.java
```diff
@@ -12,29 +12,48 @@
...
     private static final Logger logger = LogManager.getLogger(LocalNodeResponseHandler.class);
+    private final CountDownLatch inProgressLatch;
+    private DiscoveryNode localNode;
+
+    /**
+     * Instantiates a new LocalNodeResponseHandler with a count down latch
+     */
+    public LocalNodeResponseHandler() {
+        this.inProgressLatch = new CountDownLatch(1);
+        this.localNode = null;
```
owaiskazi19, line 41: Since we are setting it to `null`? Do we need it in constructor? We can just have it as a class variable

src/main/java/org/opensearch/sdk/handlers/LocalNodeResponseHandler.java
```diff
@@ -46,4 +65,15 @@ public String executor() {
...
     }
+
+    /**
+     * Invokes await on the LocalNodeResponseHandler count down latch
+     */
+    public void awaitResponse() throws InterruptedException {
+        inProgressLatch.await(ExtensionsOrchestrator.EXTENSION_REQUEST_WAIT_TIMEOUT, TimeUnit.SECONDS);
+    }
+
+    public DiscoveryNode getLocalNode() {
```
owaiskazi19, line 76: Small test to verify this getter?

src/main/java/org/opensearch/sdk/handlers/LocalNodeResponseHandler.java
joshpalis: We'll need it in the constructor because we'll update this upon receiving a response

dbwiddis: From a style perspective I'd really like to avoid setting to null in the constructor.  It's basically a no-op.  You can just add an `= null` on line 34 and leave it off here.

src/main/java/org/opensearch/sdk/handlers/LocalNodeResponseHandler.java
dbwiddis: Since with the given implementation it is allowed to be null I'm not sure how it can be tested.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -359,13 +359,26 @@ public void sendClusterSettingsRequest(TransportService transportService) {
...
+            // Wait on local node response
+            localNodeResponseHandler.awaitResponse();
+        } catch (InterruptedException e) {
+            logger.info("Failed to recieve Local Node response from OpenSearch", e);
+        } catch (Exception e) {
+            logger.info("Failed to send Local Node request to OpenSearch", e);
+        }
+
+        // At this point, response handler has read in the local node
+        return localNodeResponseHandler.getLocalNode();
```
dbwiddis, line 381: The comment above this line isn't really true unless we follow the "happy path".  I'd prefer returning null (if that's our choice) from inside the catch blocks above so this method actually does indicate success.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -359,13 +359,26 @@ public void sendClusterSettingsRequest(TransportService transportService) {
...
+                opensearchNode,
+                ExtensionsOrchestrator.REQUEST_EXTENSION_LOCAL_NODE,
+                new ExtensionRequest(ExtensionsOrchestrator.RequestType.REQUEST_EXTENSION_LOCAL_NODE),
+                localNodeResponseHandler
+            );
+            // Wait on local node response
+            localNodeResponseHandler.awaitResponse();
+        } catch (InterruptedException e) {
+            logger.info("Failed to recieve Local Node response from OpenSearch", e);
+        } catch (Exception e) {
```
dbwiddis, line 376: Do we know what types of exceptions may be thrown here?  I really don't like catching all exceptions here.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -359,13 +359,26 @@ public void sendClusterSettingsRequest(TransportService transportService) {
...
+            logger.info("Failed to send Local Node request to OpenSearch", e);
```
dbwiddis, line 377: Should these failure log messages be higher than "info" level?

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -359,13 +359,26 @@ public void sendClusterSettingsRequest(TransportService transportService) {
...
      * @param transportService  The TransportService defining the connection to OpenSearch.
      */
-    public void sendLocalNodeRequest(TransportService transportService) {
-        sendGenericRequestWithExceptionHandling(
-            transportService,
-            ExtensionsOrchestrator.RequestType.REQUEST_EXTENSION_LOCAL_NODE,
-            ExtensionsOrchestrator.REQUEST_EXTENSION_LOCAL_NODE,
-            new LocalNodeResponseHandler()
-        );
+    public DiscoveryNode sendLocalNodeRequest(TransportService transportService) {
```
dbwiddis, line 362: How are we handling the possible null return from this method?

dbwiddis (changes requested):

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
joshpalis: Sure ill make this catch block more specific

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
joshpalis: Ill modify this so that the log messages are of level ERROR

dbwiddis (approved):

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -170,8 +169,8 @@ public void setExtensionNode(DiscoveryExtension extensionNode) {
         this.extensionNode = extensionNode;
     }
 
-    DiscoveryNode getOpensearchNode() {
-        return opensearchNode;
+    public DiscoveryNode getLocalNode() {
```
dbwiddis, line 172: Our exceptions for javadocs on getters require the return value match the field name, so the build is failing on lack of javadoc. Which brings up the question, why rename this?  "LocalNode" implies to me that it's the local (to the extension) node, not the node of OpenSearch/ExtensionsOrchestrator.

joshpalis: Ill keep the original name

dbwiddis (approved):

owaiskazi19 (approved):

PR #203, opened 2022-10-25
Enforce type safety for ExtensionNamedWriteableRegistry

mloufra:

### Description
Add generic type for `categoryClass` in `ExtensionNamedWriteableRegistry` and `TestNamedWriteableRegistryAPI`
Equivalent PR on OpenSearch: https://github.com/opensearch-project/OpenSearch/pull/4923

### Issues Resolved
https://github.com/opensearch-project/opensearch-sdk-java/issues/130

### DIFF
--- a/src/main/java/org/opensearch/sdk/ExtensionNamedWriteableRegistry.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionNamedWriteableRegistry.java
@@ -100,7 +100,6 @@ public class ExtensionNamedWriteableRegistry {
      * @throws IOException if InputStream generated from the byte array is unsuccessfully closed
      * @return A response acknowledging the request to parse has executed successfully
      */
-    @SuppressWarnings("unchecked")
     public ExtensionBooleanResponse handleNamedWriteableRegistryParseRequest(NamedWriteableRegistryParseRequest request)
         throws IOException {
 
@@ -108,6 +107,7 @@ public class ExtensionNamedWriteableRegistry {
         boolean status = false;
 
         // Extract data from request and procress fully qualified category class name into class instance
+        @SuppressWarnings("unchecked")
         Class<? extends NamedWriteable> categoryClass = request.getCategoryClass();
         byte[] context = request.getContext();
 
--- a/src/main/java/org/opensearch/sdk/ExtensionNamedWriteableRegistry.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionNamedWriteableRegistry.java
@@ -107,7 +107,6 @@ public class ExtensionNamedWriteableRegistry {
         boolean status = false;
 
         // Extract data from request and procress fully qualified category class name into class instance
-        @SuppressWarnings("unchecked")
         Class<? extends NamedWriteable> categoryClass = request.getCategoryClass();
         byte[] context = request.getContext();
 
--- a/src/main/java/org/opensearch/sdk/ExtensionNamedWriteableRegistry.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionNamedWriteableRegistry.java
@@ -20,6 +20,7 @@ import java.util.Map;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.common.io.stream.InputStreamStreamInput;
+import org.opensearch.common.io.stream.NamedWriteable;
 import org.opensearch.common.io.stream.NamedWriteableAwareStreamInput;
 import org.opensearch.common.io.stream.NamedWriteableRegistry;
 import org.opensearch.common.io.stream.NamedWriteableRegistryParseRequest;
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/sdk/ExtensionNamedWriteableRegistry.java
```diff
@@ -79,12 +80,13 @@ private List<NamedWriteableRegistry.Entry> getNamedWriteables() {
...
      */
+    @SuppressWarnings("unchecked")
     public NamedWriteableRegistryResponse handleNamedWriteableRegistryRequest(OpenSearchRequest request) {
         logger.info("Registering Named Writeable Registry Request recieved from OpenSearch.");
         // Iterate through Extensions's named writeables and add to extension entries
-        Map<String, Class> extensionEntries = new HashMap<>();
+        Map<String, Class<? extends NamedWriteable>> extensionEntries = new HashMap<>();
         for (NamedWriteableRegistry.Entry entry : this.namedWriteables) {
-            extensionEntries.put(entry.name, entry.categoryClass);
+            extensionEntries.put(entry.name, (Class<? extends NamedWriteable>) entry.categoryClass);
```
dbwiddis, line 89: Note to other reviewers:  I blinked when I saw this direct field access and cast instead of a getter, but that's apparently the design, and the field is just a `<?>` despite being set with a `<T extends NamedWriteable>` in the constructor.  So this, surprisingly, is correct. 👀

src/main/java/org/opensearch/sdk/ExtensionNamedWriteableRegistry.java
```diff
@@ -98,14 +100,15 @@ public NamedWriteableRegistryResponse handleNamedWriteableRegistryRequest(OpenSe
      * @throws IOException if InputStream generated from the byte array is unsuccessfully closed
      * @return A response acknowledging the request to parse has executed successfully
      */
+    @SuppressWarnings("unchecked")
```
dbwiddis, line 103: Similar to the comment on the companion PR, move this down to the local variable at line 111.

src/test/java/org/opensearch/sdk/TestNamedWriteableRegistryAPI.java
```diff
@@ -135,12 +135,12 @@ public void testInvalidCategoryClass() throws UnknownHostException, IOException
         StreamInput in = new InputStreamStreamInput(input);
 
         // Category Class ExtensionRunner is not registered
-        NamedWriteableRegistryParseRequest request = new NamedWriteableRegistryParseRequest(ExtensionsRunner.class, in);
+        NamedWriteableRegistryParseRequest request = new NamedWriteableRegistryParseRequest(NamedWriteable.class, in);
```
dbwiddis, line 138: (Opinion, ok but think about changing) I think the intention of this test is to have an invalid class that will never be a NamedWriteable.  This probably works as that's an interface, but I'd prefer to use a class we have more control over on the SDK side. that we know will never be a registered namedWritable (per the comment above on line 137).

dbwiddis (commented):
Looking good! One minor nit.

src/main/java/org/opensearch/sdk/ExtensionNamedWriteableRegistry.java
mloufra: For this `@SuppressWarnings`, if I move this down to the local variable at line 111, it will cause error like `extensionEntries cannot be resolved to a type`. That's the reason that I move `@SuppressWarnings` to the top of method.

src/main/java/org/opensearch/sdk/ExtensionNamedWriteableRegistry.java
mloufra: @dbwiddis you mean the right way to add generic type is `(Class<T extends NamedWriteable>)` instead of `(Class<? extends NamedWriteable>)`?

src/test/java/org/opensearch/sdk/TestNamedWriteableRegistryAPI.java
mloufra: I have a question on it. Since we added generic type for `NamedWriteableRegistryParseRequest`, if we use an invalid class which will never be a NamedWriteable instead of NameWriteable, the object `request` will not able to be generated. It will cause error like `incompatible types`.

dbwiddis: Ah, that makes sense because that is the whole reason for you adding the type safety.

But we know we will not ever register the interface in the registry, so this is OK as you have it!

src/main/java/org/opensearch/sdk/ExtensionNamedWriteableRegistry.java
dbwiddis: No, you have it fine for this use, the `T` is in the `NamedWriteableRegistry.Entry` definition.

src/main/java/org/opensearch/sdk/ExtensionNamedWriteableRegistry.java
dbwiddis: I think that's because of the change on the OpenSearch side, if you publish to maven local it should work.

src/main/java/org/opensearch/sdk/ExtensionNamedWriteableRegistry.java
```diff
@@ -105,7 +107,8 @@ public ExtensionBooleanResponse handleNamedWriteableRegistryParseRequest(NamedWr
         boolean status = false;
 
         // Extract data from request and procress fully qualified category class name into class instance
-        Class categoryClass = request.getCategoryClass();
+        @SuppressWarnings("unchecked")
```
dbwiddis, line 110: Huh, so I made the comment earlier thinking this needed it, but the return type of `request.getCategoryClass()` is already  `Class<? extends NamedWriteable>`, so this shouldn't need to be suppressed at all.

dbwiddis (changes requested):
You only need to suppress an unchecked cast when you actually cast.  After updating to the maven local you didn't need to cast any more, so you don't need to suppress any more either!

dbwiddis (approved):

owaiskazi19 (approved):

PR #205, opened 2022-10-28
Move extension settings reading into the BaseExtension

### LINKED ISSUE #182: [FEATURE] Move extension settings reading into the BaseExtension

#146 added a BaseExtension superclass of Extension implementations to hide a lot of the repetitive implementation details and simplify the actual extensions.

There's some awkward exception handling and initialization in the HelloWorldExtension that could be moved to the BaseExtension.

The `extensionSettings` field and the initialization in the constructor could be moved to the BaseExtension.  
 - The implementing class (HelloWorldExtension) would just pass an `ExtensionSettings` object to `super()`, either:
   - `new ExtensionSettings( ... )`
   - or `Extension.readSettingsFromYaml(EXTENSION_SETTINGS_PATH);`
 - All the exception handling would occur in the BaseExtension

Leaving it as is so extension authors can handle it themselves.

DRY

mloufra:

### Description
Move the `extensionSettings` field and the initialization in the constructor and all the exception handling in `HelloWorldExtension` to `BaseExtension`.

### Issues Resolved
Fixes #182

### DIFF
--- a/src/main/java/org/opensearch/sdk/BaseExtension.java
+++ b/src/main/java/org/opensearch/sdk/BaseExtension.java
@@ -56,7 +56,7 @@ public abstract class BaseExtension implements Extension {
             if (settings == null || settings.getHostAddress() == null || settings.getHostPort() == null) {
                 throw new IOException("Failed to initialize Extension settings. No port bound.");
             }
-            this.settings = Extension.readSettingsFromYaml(path);   
+            this.settings = Extension.readSettingsFromYaml(path);
         } catch (IOException e) {
             throw new ExceptionInInitializerError(e);
         }
--- a/src/main/java/org/opensearch/sdk/BaseExtension.java
+++ b/src/main/java/org/opensearch/sdk/BaseExtension.java
@@ -9,6 +9,7 @@
 
 package org.opensearch.sdk;
 
+import java.io.IOException;
 import java.util.Collection;
 import java.util.Collections;
 
@@ -35,10 +36,40 @@ public abstract class BaseExtension implements Extension {
     protected ThreadPool threadPool;
 
     /**
-     * Empty constructor to fulfill abstract class requirements
+     * Optional classpath-relative path to a yml file containing extension settings.
+     */
+    protected static final String EXTENSION_SETTINGS_PATH = "/sample/helloworld-settings.yml";
+
+    /**
+     * The extension settings include a name, host address, and port.
+     */
+    protected ExtensionSettings settings;
+
+    /**
+     * Instantiate this extension, initializing the connection settings and REST actions.
      */
     protected BaseExtension() {
+        try {
+            this.settings = initializeSettings();
+        } catch (IOException e) {
+            throw new ExceptionInInitializerError(e);
+        }
+    }
 
+    /**
+     * The Extension must provide its settings to the ExtensionsRunner.
+     * These may be optionally read from a YAML file on the class path.
+     * Or you may directly instantiate with the ExtensionSettings constructor.
+     *
+     * @return This extension's settings.
+     * @throws IOException on failure to load settings.
+     */
+    protected static ExtensionSettings initializeSettings() throws IOException {
+        ExtensionSettings settings = Extension.readSettingsFromYaml(EXTENSION_SETTINGS_PATH);
+        if (settings == null || settings.getHostAddress() == null || settings.getHostPort() == null) {
+            throw new IOException("Failed to initialize Extension settings. No port bound.");
+        }
+        return settings;
     }
 
     /**
--- a/src/main/java/org/opensearch/sdk/BaseExtension.java
+++ b/src/main/java/org/opensearch/sdk/BaseExtension.java
@@ -38,20 +38,12 @@ public abstract class BaseExtension implements Extension {
     /**
      * The extension settings include a name, host address, and port.
      */
-    private ExtensionSettings settings;
-
-    /**
-     * Empty constructor to fulfill abastract class requirement
-     */
-    protected BaseExtension() {
-
-    }
+    private final ExtensionSettings settings;
 
     /**
      * Instantiate this extension, initializing the connection settings and REST actions.
-     * @throws IOException on failure to load settings.
      */
-    protected BaseExtension(String path) throws IOException {
+    protected BaseExtension(String path) {
         try {
             this.settings = Extension.readSettingsFromYaml(path);
             if (settings == null || settings.getHostAddress() == null || settings.getHostPort() == null) {
@@ -60,6 +52,13 @@ public abstract class BaseExtension implements Extension {
         } catch (IOException e) {
             throw new ExceptionInInitializerError(e);
         }
+    }
+
+    /**
+     * take an ExtensionSettings object and set it directly
+     */
+    protected BaseExtension(ExtensionSettings settings) {
+        this.settings = settings;
     }
 
     @Override
--- a/src/main/java/org/opensearch/sdk/BaseExtension.java
+++ b/src/main/java/org/opensearch/sdk/BaseExtension.java
@@ -36,40 +36,35 @@ public abstract class BaseExtension implements Extension {
     protected ThreadPool threadPool;
 
     /**
-     * Optional classpath-relative path to a yml file containing extension settings.
-     */
-    protected static final String EXTENSION_SETTINGS_PATH = "/sample/helloworld-settings.yml";
-
-    /**
      * The extension settings include a name, host address, and port.
      */
-    protected ExtensionSettings settings;
+    private ExtensionSettings settings;
+
+    /**
+     * Empty constructor to fulfill abastract class requirement
+     */
+    protected BaseExtension() {
+
+    }
 
     /**
      * Instantiate this extension, initializing the connection settings and REST actions.
+     * @throws IOException on failure to load settings.
      */
-    protected BaseExtension() {
+    protected BaseExtension(String path) throws IOException {
         try {
-            this.settings = initializeSettings();
+            this.settings = Extension.readSettingsFromYaml(path);
+            if (settings == null || settings.getHostAddress() == null || settings.getHostPort() == null) {
+                throw new IOException("Failed to initialize Extension settings. No port bound.");
+            }
         } catch (IOException e) {
             throw new ExceptionInInitializerError(e);
         }
     }
 
-    /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -35,10 +36,40 @@ public abstract class BaseExtension implements Extension {
     protected ThreadPool threadPool;
 
     /**
-     * Empty constructor to fulfill abstract class requirements
+     * Optional classpath-relative path to a yml file containing extension settings.
+     */
+    protected static final String EXTENSION_SETTINGS_PATH = "/sample/helloworld-settings.yml";
```
dbwiddis, line 41: This path constant (if used) needs to stay in the Extension subclass/implementation.  This `BaseExtension` class shouldn't have any extension-specific code.

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -35,10 +36,40 @@ public abstract class BaseExtension implements Extension {
...
+
+    /**
+     * The extension settings include a name, host address, and port.
+     */
+    protected ExtensionSettings settings;
+
+    /**
+     * Instantiate this extension, initializing the connection settings and REST actions.
      */
     protected BaseExtension() {
```
dbwiddis, line 51: The constructor here should take an argument that lets this class find the `ExtensionSettings`. This is how the subclass will pass its settings in.  You will need one or two constructors:
1. Definitely include one that allows the user to just instantiate extension settings without file reading, e.g., the constructor in the subclass would call `super(new ExtensionSettings("hello-world", "127.0.0.1", "4532"));`
2. Allow the YAML path one of these two ways:
   - Create a `String` constructor and just pass in the path and do the file reading here
   - Pass the resolved path to the constructor you created in bullet 1, like `super(Extension.readSettingsFromYaml(EXTENSION_SETTINGS_PATH));`.  This keeps it to only one constructor, but also introduces complexity for the user.  So I think I'd prefer just passing the String.

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -35,10 +36,40 @@ public abstract class BaseExtension implements Extension {
...
 
+    /**
+     * The Extension must provide its settings to the ExtensionsRunner.
+     * These may be optionally read from a YAML file on the class path.
+     * Or you may directly instantiate with the ExtensionSettings constructor.
+     *
+     * @return This extension's settings.
+     * @throws IOException on failure to load settings.
+     */
+    protected static ExtensionSettings initializeSettings() throws IOException {
```
dbwiddis, line 67: You can't inherit `static` methods so this should definitely not be `protected`. But I think this code is going to be moved into the constructor anyway (along with the javadoc outlining the two options).

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -35,10 +36,40 @@ public abstract class BaseExtension implements Extension {
...
     /**
-     * Empty constructor to fulfill abstract class requirements
+     * Optional classpath-relative path to a yml file containing extension settings.
+     */
+    protected static final String EXTENSION_SETTINGS_PATH = "/sample/helloworld-settings.yml";
+
+    /**
+     * The extension settings include a name, host address, and port.
+     */
+    protected ExtensionSettings settings;
```
dbwiddis, line 46: You'll be setting this in the constructor, so make it `private final`.  And add a getter (or move it up from the subclass) that overrides `getExtensionSettings()`.

dbwiddis (changes requested):

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -35,12 +36,37 @@ public abstract class BaseExtension implements Extension {
     protected ThreadPool threadPool;
 
     /**
-     * Empty constructor to fulfill abstract class requirements
+     * The extension settings include a name, host address, and port.
+     */
+    private ExtensionSettings settings;
```
dbwiddis, line 41: We are only ever setting this once, in the constructor, so it should be `final`.

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -35,12 +36,37 @@ public abstract class BaseExtension implements Extension {
...
+
+    /**
+     * Empty constructor to fulfill abastract class requirement
      */
     protected BaseExtension() {
```
dbwiddis, line 46: I don't think we need this constructor any more now that we have others, and we want to require subclasses to provide settings.

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -35,12 +36,37 @@ public abstract class BaseExtension implements Extension {
...
 
     }
 
+    /**
+     * Instantiate this extension, initializing the connection settings and REST actions.
+     * @throws IOException on failure to load settings.
+     */
+    protected BaseExtension(String path) throws IOException {
```
dbwiddis, line 54: This does not throw `IOException` as it is caught, and an `ExceptionInInitializerError` is thrown instead.  That doesn't need to be added to a throws block as it's a runtime exception.

src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -35,12 +36,37 @@ public abstract class BaseExtension implements Extension {
...
+        try {
+            this.settings = Extension.readSettingsFromYaml(path);
+            if (settings == null || settings.getHostAddress() == null || settings.getHostPort() == null) {
+                throw new IOException("Failed to initialize Extension settings. No port bound.");
+            }
+        } catch (IOException e) {
+            throw new ExceptionInInitializerError(e);
+        }
+    }
+
```
dbwiddis, line 64: Please also add a constructor that takes an `ExtensionSettings` object and sets it directly.

Optionally you can change the `this.settings = ` in the other constructor to call the new one you created like `this(Extension.readSettingsFromYaml(path));`.

src/main/java/org/opensearch/sdk/sample/helloworld/HelloWorldExtension.java
```diff
@@ -36,46 +36,20 @@ public class HelloWorldExtension extends BaseExtension {
     private static final String EXTENSION_SETTINGS_PATH = "/sample/helloworld-settings.yml";
 
     /**
-     * The extension settings include a name, host address, and port.
-     */
-    private ExtensionSettings settings;
-
-    /**
-     * Instantiate this extension, initializing the connection settings and REST actions.
```
dbwiddis, line 44: Can we keep this as the first line of the constructor javadoc?

src/main/java/org/opensearch/sdk/sample/helloworld/HelloWorldExtension.java
```diff
@@ -36,46 +36,20 @@ public class HelloWorldExtension extends BaseExtension {
...
-    }
-
-    /**
+     *
      * The Extension must provide its settings to the ExtensionsRunner.
      * These may be optionally read from a YAML file on the class path.
      * Or you may directly instantiate with the ExtensionSettings constructor.
      *
-     * @return This extension's settings.
      * @throws IOException on failure to load settings.
```
dbwiddis, line 44: Don't think we need this `@throws`.

src/main/java/org/opensearch/sdk/sample/helloworld/HelloWorldExtension.java
```diff
@@ -36,46 +36,20 @@ public class HelloWorldExtension extends BaseExtension {
...
      */
-    private static ExtensionSettings initializeSettings() throws IOException {
-        ExtensionSettings settings = Extension.readSettingsFromYaml(EXTENSION_SETTINGS_PATH);
-        if (settings == null || settings.getHostAddress() == null || settings.getHostPort() == null) {
-            throw new IOException("Failed to initialize Extension settings. No port bound.");
-        }
-        return settings;
+    public HelloWorldExtension() throws IOException {
```
dbwiddis, line 46: Don't think we need this `@throws`.

dbwiddis (changes requested):
Looking good!  A few more tweaks needed!

mloufra:
> Looking good! A few more tweaks needed!

Thank you for you comment @dbwiddis

src/main/java/org/opensearch/sdk/BaseExtension.java
mloufra: @dbwiddis I add new constructor in `BaseExtension` which takes ExtensionSetting object and sets it directly. 
`protected BaseExtension(ExtensionSettings settings) {
        this.settings = settings;
    }`
Since the test class `ExtensionsRunnerForTest()` was using this empty constructor from `BaseExtension` in its constructor,  the `ExtensionsRunnerForTest()` Constructor need to add parameter. No sure we really need to pass the variable to the Constructor in `ExtensionsRunnerForTest()`.

owaiskazi19 (dismissed):

dbwiddis (dismissed):
LGTM!  Great job, I think this will really simplify the experience for users!

@owaiskazi19 FYI, once this is merged you may want to simplify the AD Extension constructor as well.

owaiskazi19:
@dbwiddis can you resolve the comments if they look good to you? Also, let's hold merging this until we fix the build error

dbwiddis:
> @dbwiddis can you resolve the comments if they look good to you? Also, let's hold merging this until we fix the build error

Done.

src/main/java/org/opensearch/sdk/BaseExtension.java
dbwiddis: > @dbwiddis Why do we need add `final` for `ExtensionSettings settings`?

It's not so much a "need" as good programming practice if we know a value will never change.  It helps you:
1. Make sure if you add a new constructor that you set the value correctly
2. Avoid adding other code elsewhere that changes the value at a point later than instantiation (which could impact thread safety)

Adding `final` makes it a lot easier to not have to carefully review other accesses of the variable.

dbwiddis (approved):

owaiskazi19 (approved):

owaiskazi19:
@dbwiddis I'll let you merge this one based on the sequence to avoid merge conflicts

Issue #174, opened 2022-10-10
[PROPOSAL] Pair TransportService / ActionListener initialization to allow shutdown/restart hot plugging


dbwiddis:
### What are you proposing?

Establish the capability to:
 - Add a new extension without restarting OpenSearch
 - Remove an active extension (and its dependencies) without restarting OpenSearch
 - Auto-reboot extensions to handle transport failures (or upon request by a user)

### What users have asked for this feature?

In PR #172 (based on #171), we pulled test code out of the ExtensionsRunner. This exposed a flaw with our current "initialize and leave it running forever" setup and the need to include `TransportService.stop()` somewhere in the API.  

### What problems are you trying to solve?

While adding a stop/start method is easy enough, the bigger question is, when will that be called?  Testing as in #172 is only one use case.  But we can also integrate this with our longer term goal of hot plugging extensions.

As part of that effort we are dealing with the sequencing of initializing Extensions, and knowing when they are initialized.  See issues #65 and #94, and possible usage in #149 and #151.

During some of my REST handler testing, I've ended up with a bug causing Transport to fail, with the only way to fix it restarting the extension (and restarting OpenSearch).   And we want a future where we do not have to restart OpenSearch and can dynamically add or remove extensions on-the fly.

### What is the developer experience going to be?

 - To add a new extension on the fly, start it up and then make a REST request to connect it: `PUT /_extensions/add/uniqueID`
 - To remove a new extension on the fly, start it up and then make a REST request to connect it: `PUT /_extensions/remove/uniqueID`
 - To (attempt to) reconnect to a problematic extension which executes the above two steps and sends a signal to the extension to restart itself: `PUT /_extensions/restart/uniqueID`

#### Are there any security considerations? 

Extensions should require any shutdown/reboot requests come from the OpenSearch instance they already are (or have been) communicating with.

### What is the user experience going to be?

A user can spin up a new extension on a new node (or same node) and issue a single REST request to activate it (and its dependencies).

A user can remove an extension, or restart it to upgrade to a new software version.

### Why should it be built? Any reason not to?

There may be other ways to handle this functionality but connecting REST handlers to the transport service start/stop seems a great way to enable this functionality.

### What will it take to execute?

We'll need to create a design/sequence diagram that clearly identifies the complete set of initialization steps and what is required to "uninitialize".

### Any remaining open questions?

Dependencies are a big unknown here.  This proposal primarily deals with extensions without dependencies.

[...]

PR #244, opened 2022-11-12
Make NamedXContent available to extensions 

### LINKED ISSUE #94: [PROPOSAL] Align on meaning of a successfully initialized extension
### What are you proposing?

Initializing an extension requires several steps, including:
 - Establishing a transport connection between the nodes
 - Passing REST API from extension to orchestrator
 - Registering the REST API with the RestHandler
 - Passing NamedWriteables from extension to orchestrator
 - Registering the NamedWriteables 
 - Future security capabilities such as passing tokens

Each of these events involves network transporation and ample possibility (expectation) of occasional IOExceptions.  We need to align on:
 - what are the minimum steps to consider an extension initialized?
 - how should transport errors be handled? Should there be retries on failures?
 - when and how can an extension's API or NamedWriteables be updated on-the-fly and how will the success of this update be communicated

### How did you come up with this proposal?

Background: https://github.com/opensearch-project/opensearch-sdk-java/pull/74#discussion_r946011288

### What is the user experience going to be?

Ideally, everything just works for the user on initialization.  In the event things don't work, the user should have sufficient information in log files to quickly identify and correct any configuration errors.

If the user is updating API or NamedWriteables during operation, the success of the update should be promptly communicated to the user, with similar debugging information as above on failure.

### Why should it be built? Any reason not to?

Users need to understand what the log message "Extension initialized" implies.

### What will it take to execute?

We need a more detailed design document on extension lifecycle.  

Presently extensions are assumed to be already in operation when the orchestrator starts.  We need to handle extension node shutdown/restart or changing nodes, changing APIs, removing an extension mid-operation (and removing its registrations), etc.

### What are remaining open questions?

What other exceptional situations need to be handled?

### LINKED ISSUE #131: Determining what to do with sendClusterStateRequest, sendClusterSettingsRequest, sendLocalNodeRequest and sendRegisterRestActionsRequest
We need to figure out how the functionality of sendClusterStateRequest, sendClusterSettingsRequest, sendLocalNodeRequest and sendRegisterRestActionsRequest are going to work for the SDK.

### LINKED ISSUE #208: [FEATURE] Give Extensions access to core NamedXContent

Extensions need access to core NamedXContent, such as predefined range filters, cumulative sums, etc.  These will need to be part of the content parser used with the ExtensionRestRequest:

>  #163 is now merged, so in an extension when processing an `ExtensionRestRequest` we will have a `getXContentType()` getter, returning either null, or one of the `XContentType` enums.  The equivalent of `contentParser()` is:
> ```java
> request.getXContentType().xContent().createParser(xContentRegistry, LoggingDeprecationHandler.INSTANCE, content.streamInput())
> ```
> To create the `xContentRegistry` parameter in the above, note the AD Plugin has a `getNamedXContent()` method already: https://github.com/opensearch-project/anomaly-detection/blob/08fdbdde4c2f97c27f63d3a4ba31f8acd6dbf570/src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java#L938-L946
> 
> Add that to an empty registry as follows
> ```java
> NamedXContentRegistry xContentRegistry() {
>     SearchModule searchModule = new SearchModule(Settings.EMPTY, Collections.emptyList());
>     List<NamedXContentRegistry.Entry> entries = searchModule.getNamedXContents();
>     entries.addAll(getNamedXContent());
>     return new NamedXContentRegistry(entries);
> }
> ```
> 
> Creating the registry like the above will probably end up being a function of `createComponents()` as part of issue #160  with the `ExtensionsRunner` class reading the `getNamedXContent()` from the extension (default empty list).  cc: @owaiskazi19 @joshpalis @ryanbogan
> 
> _Originally posted by @dbwiddis in https://github.com/opensearch-project/opensearch-sdk-java/issues/58#issuecomment-1264255319_
>       

These should be initialized in the BaseExtension or as part of CreateComponents.

Generate the registry on the SDK side.  It's easy except for the Search Module, but we can instantiate one (such as in the above example) and figure out how to instantiate its standard methods already in core.

> > In order to generate this Xcontent parser on the SDK side, we would first need to transport the consolidated namedXcontentRegistry ...
>
> This is what I'm saying is ugly.
>
> I'm suggesting we generate the whole registry fresh on the SDK side.
>
> There are a fixed number of registries, see here:
> https://github.com/opensearch-project/OpenSearch/tree/main/libs/x-content/src/main/java/org/opensearch/common/xcontent
> 
> The registry is initialized in `Node.java` by iterating known getters:
> ```java
> NamedXContentRegistry xContentRegistry = new NamedXContentRegistry(
>     Stream.of(
>         NetworkModule.getNamedXContents().stream(),
>         IndicesModule.getNamedXContents().stream(),
>         searchModule.getNamedXContents().stream(),
>         pluginsService.filterPlugins(Plugin.class).stream().flatMap(p -> p.getNamedXContent().stream()),
>         ClusterModule.getNamedXWriteables().stream()
>     ).flatMap(Function.identity()).collect(toList())
> );
> ```
> The ones starting from uppercase class names (Network, Indices, Cluster) are static and can be called from the SDK side exactly as they are here.
> 
> _Originally posted by @dbwiddis in https://github.com/opensearch-project/opensearch-sdk-java/issues/58#issuecomment-1262927845_

Wasting a lot of bandwidth communicating the entire registry (and content) over transport.

See https://github.com/opensearch-project/anomaly-detection/pull/692 for some cherry-picked NamedXContent that would not be needed if we just had the whole registry available in a BaseExtension

dbwiddis:
### Description

Implements the `getNamedXContent()` extension point:
1. Adds `getNamedXContent()` to the `Extension` interface, so that the ExtensionsRunner can read it.  This will allow direct use of existing plugin method of this name.
2. Creates an `ExtensnionNamedXContentRegistry` which combines the Extension's custom content and the core XContent from OpenSearch.  ~Getters fetch either the combined registry (needed as an input for various parsers) or just the custom content (presently unused, but potentially available for transport request/response~ _(no need, functionality already on ExtensionsRunner)_
3. Exposes the `ExtensionNamedXContentRegistry` via a getter to extensions so they can easily integrate them into REST Handlers by:
   - Passing the `extensionsRunner` object from the Extension (inherited from `BaseExtension`) to the RestHandler
   - Using `extensionsRunner.getNamedXContentRegistry().getRegistry()` to return the (combined) registry to pass to parsers

See [AD #725](https://github.com/opensearch-project/anomaly-detection/pull/725) for an implementation leveraging these improvements!

### Issues Resolved
Fixes #208 
Fixes #131 
Fixes #94


src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -66,14 +71,12 @@ public ExtensionSettings getExtensionSettings() {
...
-    /**
-     * Returns components added by this extension.
-     *
-     * @param client A client to make requests to the system
-     * @param clusterService A service to allow watching and updating cluster state
-     * @param threadPool A service to allow retrieving an executor to run an async action
-     * @return A collection of objects
-     */
+    @Override
+    public void setExtensionsRunner(ExtensionsRunner extensionsRunner) {
```
owaiskazi19, line 75: Any use case/reason to have this setter? Shouldn't we pass the object directly [like](https://github.com/opensearch-project/anomaly-detection/blob/feature/extensions/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java#L112)

owaiskazi19:
@dbwiddis if we can add a design for getNamedXContent() in DESIGN.md would be great.

src/main/java/org/opensearch/sdk/ExtensionNamedXContentRegistry.java
```diff
@@ -0,0 +1,59 @@
...
+    public ExtensionNamedXContentRegistry(Settings settings, List<NamedXContentRegistry.Entry> extensionNamedXContent) {
+        this.namedXContentRegistry = new NamedXContentRegistry(
+            Stream.of(
+                extensionNamedXContent.stream(),
+                NetworkModule.getNamedXContents().stream(),
+                IndicesModule.getNamedXContents().stream(),
+                new SearchModule(settings, Collections.emptyList()).getNamedXContents().stream(),
+                ClusterModule.getNamedXWriteables().stream()
+            ).flatMap(Function.identity()).collect(toList())
+        );
```
owaiskazi19, line 48: Just what we wanted! Thanks for adding this 🙌

src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
```diff
@@ -60,6 +63,15 @@ public InitializeExtensionsResponse handleExtensionInitRequest(InitializeExtensi
                 extensionsRunner.opensearchNode,
                 extensionsRunner.getUniqueId()
             );
+            // Get OpenSearch Settings and set values on ExtensionsRunner
+            Settings settings = extensionsRunner.sendEnvironmentSettingsRequest(extensionTransportService);
+            extensionsRunner.setEnvironmentSettings(settings);
+            extensionsRunner.setNamedXContentRegistry(
+                new ExtensionNamedXContentRegistry(settings, extensionsRunner.getCustomNamedXContent())
+            );
+
```
owaiskazi19, line 72: Can we add a TODO here for other extension points like createComponent? Do we have to set getSettings, getRestHandlers too here? Confused a little

owaiskazi19:
#94 talks about retry for handling transport errors by doing a retry. Are we handling it in this PR?

src/main/java/org/opensearch/sdk/BaseExtension.java
dbwiddis: > Any use case/reason to have this setter? Shouldn't we pass the object directly [like](https://github.com/opensearch-project/anomaly-detection/blob/feature/extensions/src/main/java/org/opensearch/ad/AnomalyDetectorExtension.java#L112)

Note in that example we're in a static `main()` method.  So while you could receive an `ExtensionsRunner` instance as a return value from the `run()` call, we would still have to operate on that new Extension instance somehow to pass the runner into it.

There's no guarantee that `main()` method will always be in the same class as the `Extension`, so setting a static variable in that class doesn't work.  You could have a standalone class with only the `main()` method that launches several extensions in the same JVM.

dbwiddis: More specifically here's an example that would theoretically work:
```java
/**
   * Runs the specified extension.
   *
   * @param extension The extension to run.
   * @return the ExtensionsRunner instance running the extension.
   * @throws IOException on failure to bind ports.
   */
  public static ExtensionsRunner run(Extension extension) throws IOException {
    logger.info("Starting extension " + extension.getExtensionSettings().getExtensionName());
    ExtensionsRunner runner = new ExtensionsRunner(extension);
    NettyTransport nettyTransport = new NettyTransport(runner);
    Settings runnerSettings = runner.getSettings();
    ThreadPool threadPool = new ThreadPool(runnerSettings);
    runner.extensionTransportService = nettyTransport.initializeExtensionTransportService(runnerSettings,
        threadPool);
    runner.startActionListener(0);
    return runner;
  }
```
and
```java
class HelloWorldExtension extends BaseExtension {
    private static ExtensionsRunner extensionsRunner;

    // ...

    public static void main(String[] args) throws IOException {
        // Execute this extension by instantiating it and passing to ExtensionsRunner
        runner = ExtensionsRunner.run(new HelloWorldExtension());
    }
}
```
But it's super-code-smelly setting a static variable after the class has been instantiated.  And as above, `static` also assumes the class executing `main()` is the same as the extension class, which is not a requirement.

dbwiddis:
> @dbwiddis if we can add a design for getNamedXContent() in DESIGN.md would be great.

Sure, I'll provide one.  It's a good template for other similar extension points.

https://i.imgflip.com/70rp2r.jpg

dbwiddis:
> #94 talks about retry for handling transport errors by doing a retry. Are we handling it in this PR?

No, but #174 addresses this.

src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
dbwiddis: Don't think we need a TODO but I can definitely add more comments.  There's two categories here:
 - things the ExtensionRunner already knows about because they were on the Extension interface (getSettings, getRestHandlers) and just needs to send to OpenSearch
 - things we don't have enough information about yet (we need environment settings in order to create the NamedXContent registry)

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -142,15 +161,70 @@ protected ExtensionsRunner(Extension extension) throws IOException {
     }
 
     /**
-     * This method is call from {@link ExtensionsInitRequestHandler}.
+     * Marks the extension initialized.
+     */
+    public void setInitialized() {
+        this.initialized = true;
```
owaiskazi19, line 167: Do we really need this boolean value?

src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
owaiskazi19: That helps

src/main/java/org/opensearch/sdk/BaseExtension.java
owaiskazi19: Okay so `main` is in the HelloWorldExtension.java here we don't have to make any changes to Hello World Extension
https://github.com/opensearch-project/opensearch-sdk-java/blob/main/src/main/java/org/opensearch/sdk/sample/helloworld/HelloWorldExtension.java#L62

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
dbwiddis: I think it's of value for an extension to know if it's received all the data it needs.  For example, you need the NamedXContent in order to successfully create a detector.    You don't know you have that without checking this boolean.

src/main/java/org/opensearch/sdk/BaseExtension.java
dbwiddis: > Okay so `main` is in the HelloWorldExtension.java [here](https://github.com/opensearch-project/opensearch-sdk-java/blob/main/src/main/java/org/opensearch/sdk/sample/helloworld/HelloWorldExtension.java#L62) 

Yes but it doesn't have to be.

> we don't have to make any changes to Hello World Extension

We may need to pass the `runner` as an argument to a Rest Handler as in [this](https://github.com/opensearch-project/anomaly-detection/pull/725/files#diff-9d4fba791ae69d5fc024218799521d5a87e6c3cac867e7fbd2754a1d4abbb39f)

owaiskazi19: Should we change the HelloWorldExtension.java then?

dbwiddis: No, this is going to be one of many things available in the Base Extension that may need to be passed. SDK Client is another. CreateComponents will probably produce even more.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
owaiskazi19: A-OK with boolean here.

owaiskazi19 (approved):

DESIGN.md
```diff
@@ -116,9 +116,35 @@ The `ExtensionsOrchestrator` reads a list of extensions present in `extensions.y
 
 (27) The User receives the response.
 
+#### Extension Point Implementation Walk Through
+
+An example of a more complex extension point, `getNamedXConten()` is shown below. A similar pattern can be followed for most extension points.
```
ryanbogan, line 121: Nit: getNamedXContent()

ryanbogan (dismissed):

Docs/ExtensionRestActions.txt
```diff
@@ -0,0 +1,97 @@
+# File comapible with https://sequencediagram.org/
```
owaiskazi19, line 1: I'm not sure if we should commit the Docs file to the repo

dbwiddis: I figure it would help someone else edit the svg...

owaiskazi19 (dismissed):

owaiskazi19 (approved):

joshpalis (approved):

Issue #221, opened 2022-11-02
[FEATURE] Create a Rest Handler to add POST _validate/detector method


dbwiddis:
### Is your feature request related to a problem?
Similar to the Rest Handler created for Create Detector or Hello World Extension or GET method in #213, create an action implementing `ExtensionRestHandler` with routes for the validate detector API calls handled by appropriate methods with placeholder code:
```
POST /detectors/_validate
POST /detectors/_validate/{type}
```

Java code:
```
private ExtensionRestResponse handleValidateDetectorRequest(ExtensionRestRequest request) {
    // do things with request
    return new ExtensionRestResponse(request, OK, "placeholder");
}
```

Create a POST _validate Rest handler for validate detector similar to [this](https://github.com/opensearch-project/anomaly-detection/blob/ca905f15355594a1e354f4a1da7c07a893af2211/src/main/java/org/opensearch/ad/rest/RestValidateAnomalyDetectorAction.java#L105). The business logic will be later done in #222.

mloufra:
- [  ] Create `RestValidateDetectorAction.java` placed in AnomalyDetector repository feature/extensions branch
- [  ] Add `Route` method 
- [  ] Add `handleValidateDetectorRequest` method 
- [  ] Add Post_Validate action to the list of Rest Handlers in the `AnomalyDetectorExtension.java`

dbwiddis:
I like the checkbox status comment!

dbwiddis:
Completed in [AD #732](https://github.com/opensearch-project/anomaly-detection/pull/732)

Issue #213, opened 2022-11-01
[FEATURE] Create a Rest Handler to add GET method for the detector


owaiskazi19:
### Is your feature request related to a problem?
Similar to the Rest Handler created for Create Detector or Hello World Extension, create an action implementing `ExtensionRestHandler` with routes for the get detector API calls handled by appropriate methods with placeholder code:
```
GET detectors/{detectorId}
GET detectors/{detectorId}?job=true
```

Java code:
```
private ExtensionRestResponse handleGetRequest(ExtensionRestRequest request) {
    // do things with request
    return new ExtensionRestResponse(request, OK, "placeholder");
}}
```

Create a GET Rest handler for get detector similar to [this](https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/rest/RestGetAnomalyDetectorAction.java#L57). The business logic will be later done in #214

mloufra:
- [ ] Create `RestGetDetectorAction.java` placed in AnomalyDetector repository feature/extensions branch
- [ ] Add `Route` method 
- [ ] Add `handleGetRequest` method

PR #246, opened 2022-11-12
 Provide a higher level implementation for REST handlers 

### LINKED ISSUE #128: [FEATURE] Provide a higher level implementation for REST handlers

In https://github.com/opensearch-project/opensearch-sdk-java/blob/main/src/main/java/org/opensearch/sdk/sample/rest/RestHelloAction.java#L37 the developer has to implement a `handleRequest` that acts as a router that checks for methods, e.g.

```
if (Method.GET.equals(method) && "/hello".equals(uri)) {
            return new BytesRestResponse(OK, GREETING);
        }
```

This is unnecessarily verbose.

Provide a higher level implementation that implements `ExtensionRestHandler `. As a developer I would like to write something shorter:

```java
public class RestHelloAction extends GenericExtensionRestHandler {

    @Override
    public List<Route> routes() {
        return singletonList(
            new RouteHandler(GET, "/hello", this.handleHello),
            new RouteHandler(GET, "/world", this.handleWorld)
        );
    }

    public RestResponse handleHello(...) {
    }

    public RestResponse handleWorld(...) {
    }
}
```

### LINKED ISSUE #245: [FEATURE] Create boilerplate "not configured to handle" ExtensionRestResponse

This code is repeated in every Rest Handler:
```java
return new ExtensionRestResponse(
    request,
    NOT_FOUND,
    "Extension REST action improperly configured to handle " + request.toString()
);
```

Create a method somewhere (`ExtensionRestHandler` default method? Static method on `ExtensionRestResponse`?) that takes the `request` as an argument and returns the result of the above instantiation.

Copying and pasting the above in every Rest Handler implementation.

image

dbwiddis:

### Description

Creates a `RouteHandler` class extending `Route` which also includes a handler method

Creates a `BaseExtensionRestHandler` abstract class:
 - Matches the request's route to its handler
 - Moves the unhandled request boilerplate into the superclass

Note to reviewers: consider [this comment](https://github.com/opensearch-project/opensearch-sdk-java/issues/128#issuecomment-1312656871) for two possible implementations. Happy to switch back to the other one if you don't like the one I chose.  The actual implementation permits either choice; the question is more about what we want the HelloWorld sample to show.

### Issues Resolved

Fixes #128
Fixes #245

### DIFF
--- a/src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
+++ b/src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
@@ -1,0 +1,150 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.sdk;
+
+import static org.opensearch.rest.RestStatus.NOT_FOUND;
+
+import java.util.List;
+import java.util.Optional;
+
+import org.opensearch.common.bytes.BytesReference;
+import org.opensearch.common.xcontent.XContentBuilder;
+import org.opensearch.extensions.rest.ExtensionRestRequest;
+import org.opensearch.extensions.rest.ExtensionRestResponse;
+import org.opensearch.rest.RestHandler.Route;
+import org.opensearch.rest.RestStatus;
+
+/**
+ * Provides convenience methods to reduce boilerplate code in an {@link ExtensionRestHandler} implementation.
+ */
+public abstract class BaseExtensionRestHandler implements ExtensionRestHandler {
+
+    private ExtensionRestRequest request;
+
+    /**
+     * Defines a list of methods which handle each rest {@link Route}.
+     *
+     * @return a list of {@link RouteHandler} with corresponding methods to each route.
+     */
+    protected abstract List<RouteHandler> routeHandlers();
+
+    @Override
+    public List<Route> routes() {
+        return List.copyOf(routeHandlers());
+    }
+
+    @Override
+    public ExtensionRestResponse handleRequest(ExtensionRestRequest request) {
+        this.request = request;
+        Optional<RouteHandler> handler = routeHandlers().stream()
+            .filter(rh -> rh.getMethod().equals(request.method()))
+            .filter(rh -> restPathMatches(request.path(), rh.getPath()))
+            .findFirst();
+        return handler.isPresent() ? handler.get().getExtensionRestResponse() : unhandledRequest();
+    }
+
+    /**
+     * Determines if the request's path is a match for the configured handler path.
+     *
+     * @param requestPath The path from the {@link ExtensionRestRequest}
+     * @param handlerPath The path from the {@link RouteHandler}
+     * @return true if the request path matches the route
+     */
+    private boolean restPathMatches(String requestPath, String handlerPath) {
+        // Check exact match
+        if (handlerPath.equals(requestPath)) {
+            return true;
+        }
+        // Split path to evaluate named params
+        String[] handlerSplit = handlerPath.split("/");
+        String[] requestSplit = requestPath.split("/");
+        if (handlerSplit.length != requestSplit.length) {
+            return false;
+        }
+        for (int i = 0; i < handlerSplit.length; i++) {
+            if (!(handlerSplit[i].equals(requestSplit[i]) || (handlerSplit[i].startsWith("{") && handlerSplit[i].endsWith("}")))) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    /**
+     * Returns a default response when a request does not match the handled methods or paths. This can occur if a
+     * handler indicates routes that it handles but does not actually handle them.
+     *
+     * @return an ExtensionRestResponse identifying the unhandled request.
+     */
+    protected ExtensionRestResponse unhandledRequest() {
+        return createResponse(NOT_FOUND, "Extension REST action improperly configured to handle " + getRequest().toString());
+    }
+
+    /**
+     * Creates an {@link ExtensionRestResponse} with the given status and Xcontent.
+     *
+     * @param status The {@link RestStatus} for the response.
+     * @param builder An {@link XContentBuilder} for the response.
+     * @return the response.
+     */
+    protected ExtensionRestResponse createResponse(RestStatus status, XContentBuilder builder) {
+        return new ExtensionRestResponse(getRequest(), status, builder);
+    }
+
+    /**
+     * Creates an {@link ExtensionRestResponse} with the given status and text content.
+     *
+     * @param status The {@link RestStatus} for the response.
+     * @param content The content.
+     * @return the response.
+     */
+    protected ExtensionRestResponse createResponse(RestStatus status, String content) {
+        return new ExtensionRestResponse(getRequest(), status, content);
+    }
+
+    /**
+     * Creates an {@link ExtensionRestResponse} with the given status and content.
+     *
+     * @param status The {@link RestStatus} for the response.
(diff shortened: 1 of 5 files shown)


dbwiddis:
@mloufra FYI this can simplify your handlers in #213 and #221 if merged before they are (optional to include).

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Optional;
+
+import org.opensearch.extensions.rest.ExtensionRestRequest;
+import org.opensearch.extensions.rest.ExtensionRestResponse;
+import org.opensearch.rest.RestHandler.Route;
+
+/**
+ * Provides convenience methods to reduce boilerplate code in an {@link ExtensionRestHandler} implementation.
+ */
+public abstract class BaseExtensionRestHandler implements ExtensionRestHandler {
```
owaiskazi19, line 24: Just trying to understand why we need a separate `BaseExtensionRestHandler` class when `ExtensionRestHandler ` can handle the same? Can we move the code there?

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
```diff
@@ -0,0 +1,83 @@
...
+            return true;
+        }
+        // Split path to evaluate named params
+        String[] handlerSplit = handlerPath.split("/");
+        String[] requestSplit = requestPath.split("/");
+        if (handlerSplit.length != requestSplit.length) {
+            return false;
+        }
+        for (int i = 0; i < handlerSplit.length; i++) {
+            if (!(handlerSplit[i].equals(requestSplit[i]) || (handlerSplit[i].startsWith("{") && handlerSplit[i].endsWith("}")))) {
```
owaiskazi19, line 66: What's the significance of the 2nd case here?

src/main/java/org/opensearch/sdk/sample/helloworld/rest/RestHelloAction.java
```diff
@@ -47,34 +54,23 @@ public class RestHelloAction implements ExtensionRestHandler {
...
     @Override
-    public List<Route> routes() {
-        return List.of(new Route(GET, "/hello"), new Route(POST, "/hello"), new Route(PUT, "/hello/{name}"), new Route(DELETE, "/goodbye"));
+    public List<RouteHandler> routeHandlers() {
+        return List.of(
+            new RouteHandler(GET, "/hello", handleGetRequest),
+            new RouteHandler(POST, "/hello", handlePostRequest),
+            new RouteHandler(PUT, "/hello/{name}", handlePutRequest),
+            new RouteHandler(DELETE, "/goodbye", handleDeleteRequest)
+        );
```
owaiskazi19, line 63: Looks clean :)

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
dbwiddis: the two cases are:
 - handler string exactly matches request string
 - handler string is a named param, such as `{detector_id}`, which matches any string.

src/main/java/org/opensearch/sdk/sample/helloworld/rest/RestHelloAction.java
dbwiddis: Credit to @dblock for the suggestion!

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
dbwiddis: > Just trying to understand why we need a separate `BaseExtensionRestHandler` class when `ExtensionRestHandler ` can handle the same? Can we move the code there?

Good question.  With JDK8 and `default` methods on interfaces, this could be moved up.  

It's generally my personal style preference to make a pretty sandwich:
 - Clean interface with no more than the class consumer needs (routes and a single `handleRequest()` method.
 - Messy default implementation that no sane user ever wants to deal with.
 - Clean implementation class that's really simple for end users writing the code.

In core, we have a `RestHandler` interface, which is really a functional interface as the only method is `handleRequest()` (and we should probably mimic that part?).

But we also have a `BaseRestHandler` which does the ugly guts of handling the request and only requires the end user to override `prepareRequest()`.

Would really like @dblock opinion on this one.

mloufra:
> @mloufra FYI this can simplify your handlers in #213 and #221 if merged before they are.
> 
> A better example of the changes you might need is in https://github.com/opensearch-project/anomaly-detection/pull/726/files
> 
> However, it's a bit out of scope of what you're doing, so it's totally optional if you think it's too confusing.

Thanks @dbwiddis  for these information, I will take a look first.

mloufra:
> 



> > @mloufra FYI this can simplify your handlers in #213 and #221 if merged before they are.
> > A better example of the changes you might need is in https://github.com/opensearch-project/anomaly-detection/pull/726/files
> > However, it's a bit out of scope of what you're doing, so it's totally optional if you think it's too confusing.

I will push the code change after this PR merged

dbwiddis:
> I will push the code change after this PR merged

@mloufra this one may take a while, don't hold up your PR

mloufra:
> > I will push the code change after this PR merged
> 
> @mloufra this one may take a while, don't hold up your PR

got it

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
owaiskazi19: I'm fine with the new abstract class. @dblock if you anything to add

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
owaiskazi19: Will named param has curly braces around? I thought it's already removed when passing the request

owaiskazi19 (dismissed):

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
dbwiddis: The param is exactly what is in the `routes()` (or `routeHandlers()`) method, with the curly braces.   So for a route `GET` `/detectors/{detectorId}` the `handlerSplit` will be   `[ "detectors", "{detectorId}" ]`.  But the user submitting the request will have sent `GET` `/extensions/_anomaly_detection/detectors/ha7jrT78wK` so the `requestSplit` wlil be `[ "detectors", "ha7jrT78wK" ]`.

So we are splitting both strings by `/`, verifying they both have length 2, then checking each element. Index 0 is an exact string match, will match the first clause.  Index 1 is not an exact string match, but the handler starts and ends with curly braces so it passes, and the handler for this combination is filtered in.  

See the test here:
https://github.com/opensearch-project/opensearch-sdk-java/blob/69c8b392c0289285213067ed0d50954f67402a0e/src/test/java/org/opensearch/sdk/sample/helloworld/rest/TestRestHelloAction.java#L68

https://github.com/opensearch-project/opensearch-sdk-java/blob/69c8b392c0289285213067ed0d50954f67402a0e/src/test/java/org/opensearch/sdk/sample/helloworld/rest/TestRestHelloAction.java#L103


When the request is passed, it will have something

owaiskazi19: Thanks for the detailed information!

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
dblock: I am a Rubyist and thus generally avoid Base* classes because they are a Java thing. I think it's entirely a matter of preference.

owaiskazi19 (approved):

ryanbogan (approved):

PR #248, opened 2022-11-13
Add gradle diffCoverage report

### LINKED ISSUE #129: Improvement about code coverage report
### What are you proposing?

- Generate a readable code coverage report locally without raising PR in GitHub to check code-coverage rate.
- Can we get code coverage report on each push ?

dbwiddis:

### Description

Adds the [Diff coverage gradle plugin](https://github.com/form-com/diff-coverage-gradle).

Developers can run `./gradlew diffCoverage` which will generate a jacoco coverage report only on the diff from `main`.
 - Keep your github fork `main` synced with upstream for this to work well! You can do it on Github with one click.
 - Due to [a bug](https://github.com/form-com/diff-coverage-gradle/issues/67) you should already have run `./gradle test` or `./gradlew check` or `./gradlew jacocoReport`. This may be fixed by 1.0.0 eventually.
 - Full results are viewable at `build/reports/jacoco/diffCoverage/html/index.html`

See #250 for a discussion on thresholds and including this in CI.

### Issues Resolved
Fixes #129 

Note #129 also mentions updating the `codecov` report on pushes. This is actually already occurring with the default configuration: you can see "edited" on the original report and view changes over time if desired.  (It's possible to configure to get a new report each push, but I prefer the edit/update default.)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,13 +1,14 @@
-- [Contributing to OpenSearch-SDK-Java](#contributing-to-opensearch-sdk-java)
+- [Contributing to OpenSearch SDK for Java](#contributing-to-opensearch-sdk-for-java)
   - [First Things First](#first-things-first)
   - [Ways to Contribute](#ways-to-contribute)
     - [Bug Reports](#bug-reports)
     - [Feature Requests](#feature-requests)
     - [Contributing Code](#contributing-code)
   - [Developer Certificate of Origin](#developer-certificate-of-origin)
+  - [Code Coverage](#code-coverage)
   - [Review Process](#review-process)
 
-# Contributing to OpenSearch-SDK
+# Contributing to OpenSearch SDK for Java
 
 OpenSearch-SDK is a community project that is built and maintained by people just like **you**. We're glad you're interested in helping out. There are several different ways you can do it, but before we talk about that, let's talk about how to get started.
 
@@ -76,6 +77,12 @@ Each commit must include a DCO which looks like this
 Signed-off-by: Jane Smith <jane.smith@email.com>
 ```
 You may type this line on your own when writing your commit messages. However, if your user.name and user.email are set in your git configs, you can use `-s` or `--signoff` to add the `Signed-off-by` line to the end of the commit message.
+
+## Code Coverage
+
+Any new functionality requires testing. Your PR will trigger an automatic assessment of the code coverage of the lines you've added. You should add unit and/or integration tests to exercise as much of your new code as possible.
+
+If you'd like to preview your coverage before submitting your PR, to identify lines of code which are not tested, you may run `./gradlew diffCoverage` and review the report available in the project build directory at `build/reports/jacoco/diffCoverage/html/index.html`.
 
 ## Review Process
 
--- a/build.gradle
+++ b/build.gradle
@@ -168,8 +168,8 @@ diffCoverageReport {
     }
 
     violationRules {
-        minBranches = 0.9
-        minLines = 0.9
-        minInstructions = 0.9
+        minBranches = 0.60
+        minLines = 0.75
+        failOnViolation = true
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -13,6 +13,7 @@ plugins {
     id 'java'
     id "com.diffplug.spotless" version "6.11.0" apply false
     id 'jacoco'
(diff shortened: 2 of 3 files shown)


saratvemulapalli (dismissed):

build.gradle
```diff
@@ -147,3 +148,28 @@ jacocoTestReport {
...
+    reports {
+        html = true
+    }
+
+    violationRules {
+        minBranches = 0.9
+        minLines = 0.9
+        minInstructions = 0.9
+    }
+}
```
owaiskazi19, line 175: Should we add codecoverage dependency on gradle check here?
```
check.dependsOn jacocoTestCoverageVerification

File: build.gradle [hunk 2]
```diff
@@ -147,3 +148,28 @@ jacocoTestReport {
...
+        //   git diff refs/remotes/origin/main > build/tmp/git.diff
+        //   ./gradlew diffCoverage
+    }
+
+    // View report at build/reports/jacoco/diffCoverage/html/index.html
+    reports {
+        html = true
+    }
+
+    violationRules {
```
owaiskazi19, line 170: How about?
```
failOnViolation = true

File: build.gradle (hunk 1 above)
Comment (dbwiddis) [reply build.gradle:175]: Given the check has to occur to generate the coverage before this happens, that would create a circular dependency. :-)

We could add this check to the gradle check workflow if we wanted, but... I don't think we want that. There are some cases where it's impossible to test, and diff coverage can swing wildly.  If someone submitted a PR just to catch an exception, it might be impossible to add coverage and the check would always fail. 

This functionality duplicates what CodeCov is already doing on PRs, so we will have that information avilable when reviewing.  Coverage is a guideline and not a hard and fast rule.

File: build.gradle (hunk 2 above)
Comment (dbwiddis) [reply build.gradle:170]: Sure, which would be the default if I deleted that block.    If we do want to fail we should decide what the required threshold for failure is.

Given that this is just a tool for developers to locally check before submitting a PR, the actual result is less important.  I left it as-is because locally I've wrapped it in a script and the failure would require me to add an `|| true` to make sure my script finishes to report missing coverage lines.  (I may submit that script later!)

File: build.gradle (hunk 1 above)
Comment (dbwiddis) [reply build.gradle:175]: Also, I believe the requirement to have a gradle check (or at least test) run first is also [a bug]() so I'm not sure we will need dependencies going the other way.   Forcing this workflow to depend on gradle check would make it take longer if you've already been working to get passing tests.

Comment (owaiskazi19) [reply build.gradle:175]: Sure. That makes sense. I feel like we should have a strict dependency on code coverage the way AD is doing [here](https://github.com/opensearch-project/anomaly-detection/blob/main/build.gradle#L615) and [here](https://github.com/opensearch-project/anomaly-detection/blob/main/build.gradle#L644). This will help us to improve the coverage of the repo. Although, we have have code coverage I am not sure if we care about coverage a lot. This will help us to overcome that. We can definitely cover this in a new issue

Comment (dbwiddis) [reply build.gradle:175]: @owaiskazi19 sure, open a discuss issue with a link to those AD suggestions and we can easily update our config later.

Comment (dbwiddis) [reply build.gradle:175]: Added https://github.com/opensearch-project/opensearch-sdk-java/issues/250 to discuss. Updated the current test to fail based on AD thresholds.

Comment (owaiskazi19) [reply build.gradle:175]: Thanks @dbwiddis for opening up the issue!

**Comment by dbwiddis** [comment]
Worked around the uncommitted file issue, should be GTG now.

**Review by owaiskazi19 — APPROVED** [review]

File: CONTRIBUTING.md [hunk 3]
```diff
@@ -1,10 +1,11 @@
-- [Contributing to OpenSearch-SDK-Java](#contributing-to-opensearch-sdk-java)
+- [Contributing to OpenSearch-SDK-Java](#contributing-to-opensearch-sdk)
```
owaiskazi19, line 1: Should we keep the title below as `Contributing to OpenSearch-SDK-Java`?

owaiskazi19 (dismissed):
Thanks for adding this!

CONTRIBUTING.md
dbwiddis: Probably "for java"....

owaiskazi19 (approved):

ryanbogan (approved):

dbwiddis:
UGH..... so all that effort to get CI passing in a PR now breaks main.

PR #295, opened 2022-12-21
Adding more unit test coverage

saratvemulapalli:
### Description
Adding few more unit tests to improve code coverage for SDK.

### DIFF
--- a/src/main/java/org/opensearch/sdk/BaseExtension.java
+++ b/src/main/java/org/opensearch/sdk/BaseExtension.java
@@ -56,6 +56,8 @@ public abstract class BaseExtension implements Extension {
             }
         } catch (IOException e) {
             throw new ExceptionInInitializerError(e);
+        } catch (Exception e) {
+            throw new RuntimeException();
         }
     }
 
--- a/src/main/java/org/opensearch/sdk/BaseExtension.java
+++ b/src/main/java/org/opensearch/sdk/BaseExtension.java
@@ -47,6 +47,7 @@ public abstract class BaseExtension implements Extension {
 
     /**
      * Instantiate this extension, initializing the connection settings and REST actions.
+     * @param path to extensions configuration.
      */
     protected BaseExtension(String path) {
         try {
@@ -54,15 +55,14 @@ public abstract class BaseExtension implements Extension {
             if (settings == null || settings.getHostAddress() == null || settings.getHostPort() == null) {
                 throw new IOException("Failed to initialize Extension settings. No port bound.");
             }
-        } catch (IOException e) {
-            throw new ExceptionInInitializerError(e);
         } catch (Exception e) {
-            throw new RuntimeException();
+            throw new IllegalArgumentException(e);
         }
     }
 
     /**
-     * take an ExtensionSettings object and set it directly
+     * take an ExtensionSettings object and set it directly.
+     * @param settings defined by the extension.
      */
     protected BaseExtension(ExtensionSettings settings) {
         this.settings = settings;
--- a/src/test/java/org/opensearch/sdk/TestBaseExtension.java
+++ b/src/test/java/org/opensearch/sdk/TestBaseExtension.java
@@ -12,36 +12,33 @@ package org.opensearch.sdk;
 import org.junit.jupiter.api.Test;
 import org.opensearch.test.OpenSearchTestCase;
 
-import java.util.List;
-
+/*
+ * Most of the code in BaseExtension is tested by HelloWorld extension tests.
+ * This class tests code paths which are not tested.
+ */
 public class TestBaseExtension extends OpenSearchTestCase {
 
-    private static final String EXTENSION_DESCRIPTOR_CLASSPATH = "/bad-extension.yml";
-    private static final String EXTENSION_DESCRIPTOR_FILEPATH = "src/test/resources" + EXTENSION_DESCRIPTOR_CLASSPATH;
+    private static final String UNPARSEABLE_EXTENSION_CONFIG = "/bad-extension.yml";
+    private static final String EXTENSION_DESCRIPTOR_FILEPATH = "src/test/resources" + UNPARSEABLE_EXTENSION_CONFIG;
 
-    private class TestExtension extends BaseExtension {
+    public class TestExtension extends BaseExtension {
 
-        protected TestExtension(String path) {
+        public TestExtension(String path) {
             super(path);
-        }
-
-        @Override
-        public List<ExtensionRestHandler> getExtensionRestHandlers() {
-            return null;
         }
     }
 
     @Test
     public void testBaseExtensionWithNullPath() {
         // When a null path is passed, reading from YAML will fail.
-        assertThrows(RuntimeException.class, () -> { TestExtension testExtension = new TestExtension(null); });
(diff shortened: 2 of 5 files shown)


src/main/java/org/opensearch/sdk/BaseExtension.java
```diff
@@ -56,6 +56,8 @@ protected BaseExtension(String path) {
             }
         } catch (IOException e) {
             throw new ExceptionInInitializerError(e);
+        } catch (Exception e) {
+            throw new RuntimeException();
```
dbwiddis, line 60: I don't see the value in catching an exception and throwing it without any message.  Why not just make the previous catch block more general?

src/test/java/org/opensearch/sdk/TestExtensionRestHandler.java
```diff
@@ -0,0 +1,33 @@
...
+        @Override
+        public ExtensionRestResponse handleRequest(ExtensionRestRequest request) {
+            return null;
+        }
+    }
+
+    @Test
+    public void testHandlerDefaultRoutes() {
+        NoOpExtensionRestHandler handler = new NoOpExtensionRestHandler();
+        assertEquals(handler.routes(), Collections.emptyList());
```
dbwiddis, line 31:
```suggestion
        assertTrue(handler.routes().isEmpty());
```

src/test/resources/bad-extension.yml
```diff
@@ -0,0 +1 @@
+extensionName: hello-world
```
dbwiddis, line 1: This name doesn't match the name implied by the filename.  Can we have "bad" in the name? ;-)

src/test/java/org/opensearch/sdk/TestBaseExtension.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package org.opensearch.sdk;
+
+import org.junit.jupiter.api.Test;
+import org.opensearch.test.OpenSearchTestCase;
+
+import java.util.List;
+
+public class TestBaseExtension extends OpenSearchTestCase {
```
dbwiddis, line 17: No objection to the tests that are included here, but seems like there should be more tested?  

I know a lot of `BaseExtension` is tested in HelloWorld tests, perhaps maybe just a comment here indicating that?

src/test/java/org/opensearch/sdk/TestBaseExtension.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private static final String EXTENSION_DESCRIPTOR_CLASSPATH = "/bad-extension.yml";
+    private static final String EXTENSION_DESCRIPTOR_FILEPATH = "src/test/resources" + EXTENSION_DESCRIPTOR_CLASSPATH;
+
+    private class TestExtension extends BaseExtension {
+
+        protected TestExtension(String path) {
```
dbwiddis, line 24: Nit: seems silly to make this `protected` rather than public.

src/test/java/org/opensearch/sdk/TestBaseExtension.java
```diff
@@ -0,0 +1,48 @@
...
+            super(path);
+        }
+
+        @Override
+        public List<ExtensionRestHandler> getExtensionRestHandlers() {
+            return null;
```
dbwiddis, line 30: Doesn't `BaseExtension` give us a default empty list?  What value is there overriding here with null?

src/test/java/org/opensearch/sdk/TestBaseExtension.java
```diff
@@ -0,0 +1,48 @@
...
+package org.opensearch.sdk;
+
+import org.junit.jupiter.api.Test;
+import org.opensearch.test.OpenSearchTestCase;
+
+import java.util.List;
+
+public class TestBaseExtension extends OpenSearchTestCase {
+
+    private static final String EXTENSION_DESCRIPTOR_CLASSPATH = "/bad-extension.yml";
```
dbwiddis, line 19: Would prefer a constant name indicating this is good path to a bad extension, e.g., `UNPARSABLE_ ...`

src/test/java/org/opensearch/sdk/TestBaseExtension.java
```diff
@@ -0,0 +1,48 @@
...
+        @Override
+        public List<ExtensionRestHandler> getExtensionRestHandlers() {
+            return null;
+        }
+    }
+
+    @Test
+    public void testBaseExtensionWithNullPath() {
+        // When a null path is passed, reading from YAML will fail.
+        assertThrows(RuntimeException.class, () -> { TestExtension testExtension = new TestExtension(null); });
```
dbwiddis, line 37: Why a general RuntimeException here instead of illegal argument exception... or just exception in initializer in both cases, which will wrap the exception that caused it.

src/main/java/org/opensearch/sdk/BaseExtension.java
saratvemulapalli: I like that. Sure!

src/test/java/org/opensearch/sdk/TestExtensionRestHandler.java
saratvemulapalli: Even better :)

src/test/resources/bad-extension.yml
saratvemulapalli: haha sure!

src/test/java/org/opensearch/sdk/TestBaseExtension.java
saratvemulapalli: Yeah I was looking at codecov results, you are right most of the code is being tested and for the ones they are not I've added them. Sure i'll add a comment.

src/test/java/org/opensearch/sdk/TestBaseExtension.java
saratvemulapalli: You are right, I'll change it!

src/test/java/org/opensearch/sdk/TestBaseExtension.java
saratvemulapalli: Looks like it doesn't.

```
    /**
     * Gets a list of {@link ExtensionRestHandler} implementations this extension handles.
     *
     * @return a list of REST handlers (REST actions) this extension handles.
     */
    List<ExtensionRestHandler> getExtensionRestHandlers();
```
We could change it to have a default implementation in the `Extension.java` interface.

dbwiddis: Yes, the `ActionPlugin` interface we are mimicking does so (along with some extension points!)
```
    /**
     * Rest handlers added by this plugin.
     */
    default List<RestHandler> getRestHandlers(
        Settings settings,
        RestController restController,
        ClusterSettings clusterSettings,
        IndexScopedSettings indexScopedSettings,
        SettingsFilter settingsFilter,
        IndexNameExpressionResolver indexNameExpressionResolver,
        Supplier<DiscoveryNodes> nodesInCluster
    ) {
    }
```

dbwiddis (approved):

ryanbogan (approved):

joshpalis (approved):

Issue #302, opened 2022-12-29
[FEATURE] Create OpenSearchClient implementation in SDK Client and deprecate it


dbwiddis:
### Is your feature request related to a problem?

To speed migration work for AD Plugin (and other existing plugins) we will create a client that can use the existing (NodeClient) API calls from those plugins.  

### What solution would you like?

Create a new getter in SDKClient that returns an instance Implementing OpenSearchClient that can be used by REST Handlers.  Add deprecation annotations and javadocs clearly stating that it is provided for backwards compatibility and will eventually be removed.

### What alternatives have you considered?

See #294

### Do you have any additional context?

See https://github.com/opensearch-project/OpenSearch/issues/5424 regarding eventual deprecation.

dbwiddis:
Existing plugin implementations use this signature:
```java
protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) throws IOException { ... }
```
Proposed implementation:
1. Add another level of inheritance between `BaseExtensionRestHandler` and the individual implementations.
2. Include an abstract `prepareRequest()` method with the above signature.
3. Implement the necessary bits to translate from the `routes()` to handle the request/response handling similar to how is done in the `handleRequest()` method of `BaseRestHandler`.

dbwiddis:
> 3. So instead of [this](https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/rest/RestIndexAnomalyDetectorAction.java#L100), we will just return the REST response and the rest should be taken care by high level rest client

I don't think that's such a simple replacement.  `client.execute()` still has to do something to execute the transport; the arguments include 3 arguments: the action (a constant, we can deal with), the request (that the action needs), and an action listener for the response.  (In this case it's created by [a private method](https://github.com/opensearch-project/anomaly-detection/blob/20d2c5f94811f62a7077142ba505bed272a83549/src/main/java/org/opensearch/ad/rest/RestIndexAnomalyDetectorAction.java#L130-L153).)

So whatever client we create still has to pass the action identifier, and the request, to be executed by another class.

dbwiddis:
Looking further into this:
 - `RemoteClusterAwareClient` looks like it could work if we had designed Extensions from scratch to run as a cluster.  Since we didn't, it ends up being pretty heavyweight for this purpose.
[...]

Issue #294, opened 2022-12-21
[DISCUSS] Client implementation (Rest High LevelClient vs OpenSearch Java client) for Extensibility


owaiskazi19:
### What are you proposing?
As we are working towards converting AnomalyDetector plugin into AnomalyDetector(AD) Extension, we are using OpenSearch Java client to make REST API calls from AD Extension to OpenSearch. We have finished the first feature of AD i.e. Create Detector using the same.
The AD plugin uses High Level Rest client and the whole plugin is written based on a parser model (XContent) whereas Java client uses JSON model. Migrating this model to Java Client causes issue such as: https://github.com/opensearch-project/opensearch-java/issues/297 and https://github.com/opensearch-project/opensearch-java/issues/257. These issues are blocking us from proceeding with the next set of features of AD.

This issue will talk about the pros/cons of both the clients to move forward for Extensibility:

High Level Rest client (present in OpenSearch):
Pros:
1. Easier to use this for converting any plugin to extension, as currently the plugin presents are using the same client.
2. Less pain in migration or less code changes, which will eventually help any plugin owner to convert their plugin into extension.
3. Request/Response models are already present in OpenSearch for the use cases.

Cons:
1. Will bring OpenSearch dependency. (SDK also has OpenSearch dependency)
2. Available only in Java. For multi-language SDK, this can be an issue
3. Open proposal for deprecating REST Client: https://github.com/opensearch-project/OpenSearch/issues/5424

OpenSearch Java Client:
Pros:
1. Avoids dependency on OpenSearch and provides isolation.
2. Available in multi-language. Will help in multi-language SDK.
3. Eventually in 4.0 plugins owners will move to this client.

Cons:
1. Difficult for converting plugins into extension as the model is different parser vs builder respectively.
2. Pain in migration as the whole plugin needs to be written from scratch.
3. Request/Response model should be match with java client requirements.
4. Doesn't have a good support for all the request, builders right now which will impact Extensibility work.

If we move forward with Rest High level client, the migration would be much easier but eventually in the future we need to migrate to Java Client. Until then Java Client will be in the good shape to incorporate the above issues which is blocking the current work of AD Extension.

dbwiddis:
A full reading of https://github.com/opensearch-project/opensearch-java/issues/262 should also be conducted before considering the above options.

owaiskazi19:
@dblock @dbwiddis @saratvemulapalli need your inputs. Thanks!
[...]

PR #306, opened 2023-01-05
Added High Level Rest Client for SDK

owaiskazi19:
Signed-off-by: Owais Kazi <owaiskazi19@gmail.com>

### Description
Added High Level Rest Client for SDK

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-sdk-java/issues/294 and https://github.com/opensearch-project/opensearch-sdk-java/issues/302

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -79,6 +79,7 @@ dependencies {
     implementation("org.opensearch:opensearch:${opensearchVersion}")
     implementation("org.apache.logging.log4j:log4j-api:${log4jVersion}")
     implementation("org.apache.logging.log4j:log4j-core:${log4jVersion}")
+    implementation ("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
     implementation("org.opensearch.client:opensearch-rest-client:${opensearchVersion}")
     implementation("org.opensearch.client:opensearch-java:${opensearchVersion}")
     implementation("org.opensearch.plugin:transport-netty4-client:${opensearchVersion}")
--- a/build.gradle
+++ b/build.gradle
@@ -79,7 +79,7 @@ dependencies {
     implementation("org.opensearch:opensearch:${opensearchVersion}")
     implementation("org.apache.logging.log4j:log4j-api:${log4jVersion}")
     implementation("org.apache.logging.log4j:log4j-core:${log4jVersion}")
-    implementation ("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
+    implementation("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
     implementation("org.opensearch.client:opensearch-rest-client:${opensearchVersion}")
     implementation("org.opensearch.client:opensearch-java:${opensearchVersion}")
     implementation("org.opensearch.plugin:transport-netty4-client:${opensearchVersion}")
--- a/src/main/java/org/opensearch/sdk/SDKClient.java
+++ b/src/main/java/org/opensearch/sdk/SDKClient.java
@@ -41,7 +41,7 @@ import javax.net.ssl.SSLEngine;
  */
 public class SDKClient {
     private OpenSearchClient javaClient;
-    private RestClient restClient = null;
+    private RestClient restClient;
     private RestHighLevelClient highLevelClient;
 
     private RestClientBuilder builder(String hostAddress, int port) {
@@ -77,7 +77,7 @@ public class SDKClient {
      * Creates OpenSearchClient for SDK. It also creates a restClient as a wrapper around Java OpenSearchClient
      * @param hostAddress The address of OpenSearch cluster, client can connect to
      * @param port The port of OpenSearch cluster
-     * @return SDKClient which is internally an OpenSearchClient. The user is responsible for calling {@link #doCloseRestClient()} when finished with the client
+     * @return The SDKClient implementation of OpenSearchClient. The user is responsible for calling {@link #doCloseJavaClient()} when finished with the client
      */
     public OpenSearchClient initializeJavaClient(String hostAddress, int port) {
         RestClientBuilder builder = builder(hostAddress, port);
@@ -101,7 +101,7 @@ public class SDKClient {
      * Creates High Level Rest Client for SDK.
      * @param hostAddress The address of OpenSearch cluster, client can connect to
      * @param port The port of OpenSearch cluster
-     * @return SDKClient which is internally an RestHighLevelClient. The user is responsible for calling {@link #doCloseRestClient()} when finished with the client
+     * @return The SDKClient implementation of RestHighLevelClient. The user is responsible for calling {@link #doCloseHighLevelClient()} when finished with the client
      */
     public RestHighLevelClient initializeRestClient(String hostAddress, int port) {
         RestClientBuilder builder = builder(hostAddress, port);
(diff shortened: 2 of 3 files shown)


build.gradle
```diff
@@ -79,6 +79,7 @@ dependencies {
     implementation("org.opensearch:opensearch:${opensearchVersion}")
     implementation("org.apache.logging.log4j:log4j-api:${log4jVersion}")
     implementation("org.apache.logging.log4j:log4j-core:${log4jVersion}")
+    implementation ("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
```
dbwiddis, line 82: Nit: extra space before parenthesis doesn't align with others.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -41,14 +42,9 @@
 public class SDKClient {
     private OpenSearchClient javaClient;
     private RestClient restClient = null;
```
dbwiddis, line 44: Why is this one initialized null but the high level client isn't?  I presume null is the default in either case, but I'd like to be more intentional here on how these are initialized and/or null-checked.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -74,6 +70,17 @@ public TlsDetails create(final SSLEngine sslEngine) {
...
             }
         });
+        return builder;
+    }
+
+    /**
+     * Creates OpenSearchClient for SDK. It also creates a restClient as a wrapper around Java OpenSearchClient
+     * @param hostAddress The address of OpenSearch cluster, client can connect to
+     * @param port The port of OpenSearch cluster
+     * @return SDKClient which is internally an OpenSearchClient. The user is responsible for calling {@link #doCloseRestClient()} when finished with the client
```
dbwiddis, line 80: I'm not sure the words "SDKClient which is..." here and for the HLRC are right, as we are inside the SDKClient class so it's not really an SDKClient.  Would prefer words such as "The SDKClient implementation of ..."

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -91,7 +98,20 @@ public TlsDetails create(final SSLEngine sslEngine) {
     }
 
     /**
-     * Close this client.
+     * Creates High Level Rest Client for SDK.
+     * @param hostAddress The address of OpenSearch cluster, client can connect to
+     * @param port The port of OpenSearch cluster
+     * @return SDKClient which is internally an RestHighLevelClient. The user is responsible for calling {@link #doCloseRestClient()} when finished with the client
```
dbwiddis, line 104: See above comments on SDKClient usage. Also the linked doClose method should be for HLRC, not Rest Client.  Also, the one that's just "RestClient" should probably be Java Client.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -91,7 +98,20 @@ public TlsDetails create(final SSLEngine sslEngine) {
...
+     */
+    public RestHighLevelClient initializeRestClient(String hostAddress, int port) {
+        RestClientBuilder builder = builder(hostAddress, port);
+
+        highLevelClient = new RestHighLevelClient(builder);
+        return highLevelClient;
+    }
+
+    /**
+     * Close java client.
```
dbwiddis, line 114: Would it be a good idea for the SDKClient itself to implement `Closeable` and include a `close()` method that calls both of these methods?   Then we only have to do one single `close()` call for SDKClient.

src/test/java/org/opensearch/sdk/TestSDKClient.java
```diff
@@ -41,4 +43,19 @@ public void testCreateClient() throws Exception {
         sdkClient.doCloseRestClient();
     }
 
+    @Test
+    public void testCreateHighLevelRestClient() throws Exception {
+        RestHighLevelClient testClient = sdkClient.initializeRestClient("localhost", 9200);
+
+        // Using the package name here as Java uses package name if the filename from different packages are same
```
dbwiddis, line 50: This comment is confusing.  We import this package/class specifically above.  What other `CreateIndexRequest` is potentially being used here?

dbwiddis (changes requested):

src/test/java/org/opensearch/sdk/TestSDKClient.java
owaiskazi19: There are two `CreateIndexRequest`:
1. From Java Client: `org.opensearch.client.opensearch.indices.CreateIndexRequest`
2. From HLRC: `org.opensearch.client.indices.CreateIndexRequest`

I tried to look for different options to avoid importing it like this. Came across the below
1. Create a different test file to import same file name from different package
2. Use package name to let Java know which exact package we are referring to. Ref [here](https://softwareengineering.stackexchange.com/a/307631)

Open to suggestions here

src/main/java/org/opensearch/sdk/SDKClient.java
dbwiddis: I'm not saying to remove the individual close, but to have a single call that attempts to close both (and is a no-op if uninitialized.).  Each of these individual ones already does a null check so are safe to call.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -91,13 +98,37 @@ public TlsDetails create(final SSLEngine sslEngine) {
     }
 
     /**
-     * Close this client.
+     * Creates High Level Rest Client for SDK.
+     * @param hostAddress The address of OpenSearch cluster, client can connect to
+     * @param port The port of OpenSearch cluster
+     * @return The SDKClient implementation of RestHighLevelClient. The user is responsible for calling {@link #doCloseHighLevelClient()} when finished with the client
+     */
+    public RestHighLevelClient initializeRestClient(String hostAddress, int port) {
```
dbwiddis, line 106: Can we mark this as `@Deprecated` with a javadoc `@deprected Provided for compatibility with existing plugins to permit migration. New development should not use this client.`

src/test/java/org/opensearch/sdk/TestSDKClient.java
dbwiddis: I don't see where `org.opensearch.client.opensearch.indices.CreateIndexRequest` is imported in this class.  Since there is a statement `import org.opensearch.client.opensearch.indices.CreateIndexRequest;` does that not specifically indicate the correct import?

dbwiddis (dismissed):
Conditionally approved (where I'll tweak it later) given your time constraints, but some suggestions I'd like handled if you have the time.

src/test/java/org/opensearch/sdk/TestSDKClient.java
owaiskazi19: We can't import same filename from different package as it throws the error while importing
```
import org.opensearch.client.indices.CreateIndexRequest;
```
```
'org.opensearch.client.opensearch.indices.CreateIndexRequest' is already defined in a single-type import
```

src/main/java/org/opensearch/sdk/SDKClient.java
owaiskazi19: Ah I see your ask. That's a good point. Pushed the change. Let me know if that's right

dbwiddis (approved):

joshpalis (approved):

Issue #353, opened 2023-01-21
[FEATURE] Migrate Create/Update/Validate Detector using RestHighLevelClient


dbwiddis:
### Is your feature request related to a problem?

This is a summary issue replacing  validate detector #217 and all its sub-issues #221, #222, #223, #224,  #225, and #226 and replaces Create Detector functionality of #58, and adds Update Detector functionality (not yet tracked in an issue).

All three of these implementations share the same set of parent classes in the AD Plugin, so implementing one would create a large diff with duplicate classes. By waiting to submit a PR until all are complete the existing classes can all be replaced, the diff will be smaller and much more understandable/useful to future migration efforts.

### What solution would you like?

 - [x] Migrate the `RestIndexAnomalyDetectorAction` which implements `POST` (Create) and `PUT` (Update) functionality
 - [x] Complete Create Detector workflow
   - [x] Implement `SearchFeatureDao` component
   - Implement `GetFieldMappingsAction` (deferred) https://github.com/opensearch-project/opensearch-sdk-java/issues/361
 - [x] Complete Update Detector workflow
   - [x] Complete `getDetector()` functionality
   - Implement `ADTaskManager` component (deferred) https://github.com/opensearch-project/opensearch-sdk-java/issues/371
 - [x] Migrate the `RestValidateAnomalyDetectorAction` which implements validate functionality 
 - [x] Complete Validate Detector Workflow
   - [x] Some "dryRun" boolean code paths may not have been tested and will require further updates 
   - [x] There may be additional steps required for validate detector model which may be moved to a new issue
 - [x] Replace existing classes with migrated versions of classes to minimize the diff
 - [x] Create a MIGRATING.md document to help future migrators
 - [x] Fix or bypass failing tests
WIP is on this fork: https://github.com/dbwiddis/anomaly-detection/tree/rhlc

dbwiddis:
Completed in [AD #792](https://github.com/opensearch-project/anomaly-detection/pull/792)

PR #367, opened 2023-01-25
Create SDKRestClient wrapper class around RestHighLevelClient

dbwiddis:

### Description

Creates a wrapper class translating the `Client` class method signatures to `RestHighLevelClient` methods.   Existing code making the various REST calls does not need to change (mostly) except for:
 - Imports may change
 - Occasional minor differences in response class parsing (e.g., List vs. Set, Iterables vs. Map entrySet, etc.)

Massively speeds the process of migrating existing plugins.

Notes:
 - Committed deprecated, so nobody should use it for new code.
   - Intentionally only partially implementing `Client` interface, to minimize the (deprecated) footprint to what is necessary
   - We can add new wrapper methods as we come across them
 - Cluster client is a placeholder, need to actually implement cluster-level features

### Issues Resolved

Part of #353 

### DIFF
--- a/src/main/java/org/opensearch/sdk/SDKClient.java
+++ b/src/main/java/org/opensearch/sdk/SDKClient.java
@@ -208,6 +208,13 @@ public class SDKClient implements Closeable {
         }
 
         /**
+         * The admin client that can be used to perform administrative operations.
+         */
+        public SDKRestClient admin() {
+            return this;
+        }
+
+        /**
          * A client allowing to perform actions/operations against the cluster.
          */
         public SDKClusterAdminClient cluster() {
--- a/src/main/java/org/opensearch/sdk/SDKClient.java
+++ b/src/main/java/org/opensearch/sdk/SDKClient.java
@@ -196,7 +196,7 @@ public class SDKClient implements Closeable {
     @Deprecated
     public static class SDKRestClient implements Closeable {
 
-        private final RestHighLevelClient rhlc;
+        private final RestHighLevelClient restHighLevelClient;
 
         /**
          * Instantiate this class wrapping a {@link RestHighLevelClient}.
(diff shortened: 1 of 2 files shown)


src/test/java/org/opensearch/sdk/TestSDKClient.java
```diff
@@ -10,52 +10,81 @@
...
+        assertInstanceOf(Cancellable.class, indicesClient.delete(new DeleteIndexRequest(), ActionListener.wrap(r -> {}, e -> {})));
+        assertInstanceOf(Cancellable.class, indicesClient.getMapping(new GetMappingsRequest(), ActionListener.wrap(r -> {}, e -> {})));
+        assertInstanceOf(Cancellable.class, indicesClient.putMapping(new PutMappingRequest(), ActionListener.wrap(r -> {}, e -> {})));
+        assertInstanceOf(
+            Cancellable.class,
+            indicesClient.rolloverIndex(new RolloverRequest("", ""), ActionListener.wrap(r -> {}, e -> {}))
+        );
+        assertInstanceOf(Cancellable.class, indicesClient.getAliases(new GetAliasesRequest(), ActionListener.wrap(r -> {}, e -> {})));
 
+        restClient.close();
```
owaiskazi19, line 87: We need to close sdkClient here too. 
```
 sdkClient.doCloseHighLevelClient()

File: src/main/java/org/opensearch/sdk/SDKClient.java [hunk 2]
```diff
@@ -140,4 +182,202 @@ public void close() throws IOException {
...
+     * <p>
+     * Do not use this client for new development.
+     *
+     * @deprecated Use {@link org.opensearch.client.opensearch.OpenSearchClient}.
+     * @see <a href="https://github.com/opensearch-project/OpenSearch/issues/5424">OpenSearch Issue 5424</a>
+     */
+    @Deprecated
+    public static class SDKRestClient implements Closeable {
+
+        private final RestHighLevelClient rhlc;
```
owaiskazi19, line 199:
```suggestion
        private final RestHighLevelClient restHighLevelClient;
```

owaiskazi19 (approved):
Minor nits. This wrapper class would definitely ease the migration from a plugin to an extension. Thanks @dbwiddis for this!

src/test/java/org/opensearch/sdk/TestSDKClient.java
dbwiddis: All this does is call `restClient.close()` with a null check, which accomplishes the same thing here.  The only difference would be test coverage, but that's covered at line 55.

I'll go ahead and change all the individual client close calls to  this model, though.

owaiskazi19 (approved):

joshpalis (approved):

saratvemulapalli (approved):

PR #391, opened 2023-01-30
Update dev guide from _opensearch-sdk-1 to _opensearch-sdk-java-1 in example curl command

cwperks:
Signed-off-by: Craig Perkins <cwperx@amazon.com>

### Description

The example curl command in the developer guide still uses the legacy `_opensearch-sdk-1` route from before this repo was renamed to `opensearch-sdk-java`. The example `extensions.yml` file uses `opensearch-sdk-java-1` so the curl should match.

### DIFF
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -188,7 +188,7 @@ It is important that the OpenSearch SDK for Java is already up and running on a 
 
 The following request is configured to be handled by the sample `HelloWorldExtension` (note the matching uniqueId):
 ```
-curl -X GET localhost:9200/_extensions/_opensearch-sdk-1/hello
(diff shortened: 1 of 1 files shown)


dbwiddis (approved):
OK with this fix to make things work, but the other thing we could do is change the uniqueID to something a lot shorter like "hw" or "hi" or "yo" or "howdy".

Approving as is, and we can revamp all the things later.

cwperks:
I recall a conversation where `/.../_hi/hello` was a little too book-of-mormon-ish ;). I would be in favor of `hw` or `example-1`

dbwiddis:
> I would be in favor of `hw` or `example-1`

Or `custom-id`.  Whatever it is needs to communicate that the user has the choice.

Again, we can merge this and decide on something later.

saratvemulapalli (approved):

PR #394, opened 2023-01-31
Add auto-backport and backport branch deletion workflows

ryanbogan:

### Description
Adds a workflow to automatically backport PR's from main to other branches.  Also includes a workflow that automatically deletes the branch created for the backport once the backport PR is merged.


saratvemulapalli (approved):

.github/workflows/backport.yml
```diff
@@ -0,0 +1,39 @@
+name: Backport
+on:
+  pull_request_target:
```
dbwiddis, line 3: Would a PR merged to the backport branch also trigger this?  Should we add `branches: main` here?

.github/workflows/backport.yml
```diff
@@ -0,0 +1,39 @@
...
+    name: Backport
+    runs-on: ubuntu-latest
+    # Only react to merged PRs for security reasons.
+    # See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target.
+    if: >
+      github.event.pull_request.merged
+      && (
+        github.event.action == 'closed'
+        || (
+          github.event.action == 'labeled'
```
dbwiddis, line 19: I realize this is copied from [here](https://github.com/opensearch-project/OpenSearch/blob/main/.github/workflows/backport.yml) but I don't get the logic here.  Closed occurs when it's merged, so it will never get to the point of checking the backport label here.

Or I need more caffeine.

dbwiddis (commented):
I'm confused. 😕

.github/workflows/backport.yml
ryanbogan: I believe the 'closed' status occurs when the PR is closed but not merged

ryanbogan: But that's probably wrong because it would never return true in that case

dbwiddis: Closed is a superset of merged, this is only triggered on closed or labeled, there's an additional check for the merge action (but that wouldn't occur for the label just being added).

ryanbogan: Yeah I looked into it a little more, and I think you might be right

ryanbogan: I'll remove it and we can see if it still works

dbwiddis: Option 1: triggers on closed -> filters on merge event -> closed is true, doesn't check label -> always triggers on all merged PRs
Option 2: triggers on labeled -> filters on merge event, exits.

I know it's the same way on OpenSearch, so I hesitate to change it here without understanding why. :)

.github/workflows/backport.yml
ryanbogan: I added it

.github/workflows/backport.yml
dbwiddis: And I would think that if it's already merged but we want to add a backport label we should be able to trigger it.

dbwiddis: I'm ok with merging this as is, since it matches OpenSearch workflow, but Someone™️  should look into what we think this logic is supposed to be doing in both places.

dbwiddis (approved):

PR #417, opened 2023-02-05
Fix add backport label workflow syntax

dbwiddis:

### Description

1. The `branches` filter does not apply to the `on` triggering events and is causing [failed workflows](https://github.com/opensearch-project/opensearch-sdk-java/actions/workflows/add_backport_label.yml). Moved it to where it needs to be.
2. The `pull_request` event fails the labeling with a "Resource not accessible by integration" because the new workflow can't run from my fork. Switched it to `pull_request_target` which runs from SDK branch.  It won't run on this PR (by design) but once this PR is merged it will begin running on other PRs.
3. Changed the version for `tibdex/github-app-token` to major version only so we don't need to update each minor version release.  Closed #395 as unneeded.

### DIFF
--- a/.github/workflows/add_backport_label.yml
+++ b/.github/workflows/add_backport_label.yml
@@ -1,8 +1,7 @@
 name: Add Backport Label
-
 on:
-  branches: main
-  pull_request:
+  pull_request_target:
+    branches: main
     types: opened
 
 jobs:
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -26,7 +26,7 @@ jobs:
     steps:
       - name: GitHub App token
         id: github_app_token
-        uses: tibdex/github-app-token@v1.5.0
+        uses: tibdex/github-app-token@v1
         with:
           app_id: ${{ secrets.APP_ID }}
           private_key: ${{ secrets.APP_PRIVATE_KEY }}


.github/workflows/add_backport_label.yml
```diff
@@ -1,8 +1,7 @@
 name: Add Backport Label
-
 on:
-  branches: main
-  pull_request:
+  pull_request_target:
+    branches: main
     types: opened
```
owaiskazi19, line 5: Can we change the types to `closed` or else we won't be able to open backport PRs for closed PR?

.github/workflows/backport.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     steps:
       - name: GitHub App token
         id: github_app_token
-        uses: tibdex/github-app-token@v1.5.0
+        uses: tibdex/github-app-token@v1
```
owaiskazi19, line 29: Just for my understanding, will mend take care of upgrading to major version directly?

.github/workflows/add_backport_label.yml
owaiskazi19: Also, we can add `labeled` as a type too.
```
types:
   closed
   labeled
```

dbwiddis: > Can we change the types to `closed` or else we won't be able to open backport PRs for closed PR?

This is just the workflow to add a label automatically. It doesn't prevent us from adding the label ourselves manually.

.github/workflows/backport.yml
dbwiddis: > will mend take care of upgrading to major version directly?

Yep, see for example #403

.github/workflows/add_backport_label.yml
dbwiddis: Although looking around, seems we might want to replace all this with the `labeler` workflow.  See https://github.com/opensearch-project/opensearch-plugins/blob/main/WORKFLOWS.md#integration-1

@ryanbogan WDYT?

ryanbogan: Looks like that is for labeling based on the modified files.  I'm guessing there would be an option to just add the same label to everything, which would accomplish the same goal.

ryanbogan (approved):

.github/workflows/add_backport_label.yml
dbwiddis: @ryanbogan We can always make `*` the modified files.  Anyway, let's keep the current one for just this one but in the future if we find we need to do more auto-labeling let's consider migrating since it's one workflow but possibly multiple labels.

owaiskazi19 (approved):

PR #420, opened 2023-02-06
Change 1.x branch to pull from 2.6 snapshot and revert 'Migrate client transports to Apache HttpClient/Core5.x'

ryanbogan:

### Description
These changes enable the 1.x branch of opensearch-sdk-java to be compatible with the 2.x branch of OpenSearch

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,7 +76,7 @@ configurations {
 
 dependencies {
 
-    def opensearchVersion = "2.6.0-SNAPSHOT"
+    def opensearchVersion = "2.+"
     def log4jVersion = "2.19.0"
     def nettyVersion = "4.1.87.Final"
     def luceneVersion = "9.4.2"
--- a/build.gradle
+++ b/build.gradle
@@ -77,6 +77,7 @@ configurations {
 dependencies {
 
     def opensearchVersion = "2.+"
+    def opensearchJavaClientVersion = "2.1.0-SNAPSHOT"
     def log4jVersion = "2.19.0"
     def nettyVersion = "4.1.87.Final"
     def luceneVersion = "9.4.2"
@@ -92,7 +93,7 @@ dependencies {
     implementation("org.apache.logging.log4j:log4j-core:${log4jVersion}")
     implementation("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
     implementation("org.opensearch.client:opensearch-rest-client:${opensearchVersion}")
-    implementation("org.opensearch.client:opensearch-java:2.1.0-SNAPSHOT")
+    implementation("org.opensearch.client:opensearch-java:${opensearchJavaClientVersion}")
     implementation("org.opensearch.plugin:transport-netty4-client:${opensearchVersion}")
     implementation("io.netty:netty-all:${nettyVersion}")
     implementation("org.apache.lucene:lucene-core:${luceneVersion}")
--- a/build.gradle
+++ b/build.gradle
@@ -76,7 +76,7 @@ configurations {
 
 dependencies {
 
-    def opensearchVersion = "3.0.0-SNAPSHOT"
+    def opensearchVersion = "2.6.0-SNAPSHOT"
     def log4jVersion = "2.19.0"
     def nettyVersion = "4.1.87.Final"
     def luceneVersion = "9.4.2"
@@ -92,7 +92,7 @@ dependencies {
     implementation("org.apache.logging.log4j:log4j-core:${log4jVersion}")
     implementation("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
     implementation("org.opensearch.client:opensearch-rest-client:${opensearchVersion}")
-    implementation("org.opensearch.client:opensearch-java:${opensearchVersion}")
+    implementation("org.opensearch.client:opensearch-java:3.0.0-SNAPSHOT")
     implementation("org.opensearch.plugin:transport-netty4-client:${opensearchVersion}")
     implementation("io.netty:netty-all:${nettyVersion}")
     implementation("org.apache.lucene:lucene-core:${luceneVersion}")
--- a/build.gradle
+++ b/build.gradle
@@ -92,7 +92,7 @@ dependencies {
     implementation("org.apache.logging.log4j:log4j-core:${log4jVersion}")
     implementation("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
     implementation("org.opensearch.client:opensearch-rest-client:${opensearchVersion}")
-    implementation("org.opensearch.client:opensearch-java:3.0.0-SNAPSHOT")
+    implementation("org.opensearch.client:opensearch-java:2.1.0-SNAPSHOT")
     implementation("org.opensearch.plugin:transport-netty4-client:${opensearchVersion}")
     implementation("io.netty:netty-all:${nettyVersion}")
     implementation("org.apache.lucene:lucene-core:${luceneVersion}")
--- a/src/main/java/org/opensearch/sdk/SDKClient.java
+++ b/src/main/java/org/opensearch/sdk/SDKClient.java
@@ -104,7 +104,7 @@ public class SDKClient implements Closeable {
         mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
 
         // Create Client
-        OpenSearchTransport transport = new RestClientTransport(restClient, new JacksonJsonpMapper());
+        OpenSearchTransport transport = new RestClientTransport(restClient, new JacksonJsonpMapper(mapper));
         javaClient = new OpenSearchClient(transport);
         return javaClient;
     }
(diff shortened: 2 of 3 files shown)


build.gradle
```diff
@@ -76,7 +76,7 @@ configurations {
 
 dependencies {
 
-    def opensearchVersion = "3.0.0-SNAPSHOT"
+    def opensearchVersion = "2.6.0-SNAPSHOT"
```
dbwiddis, line 79: How are we making sure we keep this updated with the latest OpenSearch 2.x releases?  

When 2.6.0 goes out, this would need to be updated to 2.7.0-SNAPSHOT, but there is no mechanism to let us know this.  A breaking change can go in on 2.7 that we would never see.

We should be able to use Gradle's [`ComponentSelectionRules`](https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.ComponentSelectionRules.html) with a `2+` version, only accepting versions with `SNAPSHOT` in the name.  That should get us the highest 2.x SNAPSHOT version.

build.gradle
```diff
@@ -92,7 +92,7 @@ dependencies {
     implementation("org.apache.logging.log4j:log4j-core:${log4jVersion}")
     implementation("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
     implementation("org.opensearch.client:opensearch-rest-client:${opensearchVersion}")
-    implementation("org.opensearch.client:opensearch-java:${opensearchVersion}")
+    implementation("org.opensearch.client:opensearch-java:2.1.0-SNAPSHOT")
```
dbwiddis, line 95: For my own knowledge, is there a compatibility matrix for `opensearch-java` vs. `OpenSearch`?

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -113,8 +113,8 @@ public void testHandleExtensionInitRequest() throws UnknownHostException {
             "opensearch-sdk-1",
             sourceNode.getAddress(),
             new HashMap<String, String>(),
-            Version.fromString("3.0.0"),
-            Version.fromString("3.0.0"),
+            Version.fromString("2.6.0"),
+            Version.fromString("2.6.0"),
```
dbwiddis, line 117: Eww, not only do we need to change the dependency with import, we need to change code as well?  Or does any 2.x version work in this test?   

Might consider creating a system property in the `build.gradle` and pulling it in from here.

build.gradle
ryanbogan: Not up to date but it's here: https://github.com/opensearch-project/opensearch-java/blob/main/COMPATIBILITY.md

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
ryanbogan: That's a good idea

build.gradle
ryanbogan: Those rules allow us to read from the snapshot locations?

ryanbogan: I could have a variable for `opensearchVersion` that's just `2.6.0` and another variable for `opensearchSnapshot` that concatenates the version with `-SNAPSHOT`.  Then we just have to change the version whenever a release goes out, similar to bwc tests

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
ryanbogan: Potential solution in the comment above: https://github.com/opensearch-project/opensearch-sdk-java/pull/420#discussion_r1097997733

build.gradle
dbwiddis: I just edited the version like this:
```groovy
    def opensearchVersion = "2.+"
```
And got this:
```
➜  opensearch-sdk-java git:(revert_http_client) ✗ gradle dependencies | grep opensearch-core
|    +--- org.opensearch:opensearch-core:2.6.0-SNAPSHOT
|    |    +--- org.opensearch:opensearch-core:2.6.0-SNAPSHOT (*)
|    |    \--- org.opensearch:opensearch-core:2.6.0-SNAPSHOT (*)
|    +--- org.opensearch:opensearch-core:2.6.0-SNAPSHOT
|    |    +--- org.opensearch:opensearch-core:2.6.0-SNAPSHOT (*)
|    |    \--- org.opensearch:opensearch-core:2.6.0-SNAPSHOT (*)
|    +--- org.opensearch:opensearch-core:2.6.0-SNAPSHOT
|    |    +--- org.opensearch:opensearch-core:2.6.0-SNAPSHOT (*)
|    |    \--- org.opensearch:opensearch-core:2.6.0-SNAPSHOT (*)
|    |    +--- org.opensearch:opensearch-core:2.6.0-SNAPSHOT (*)
|    +--- org.opensearch:opensearch-core:2.6.0-SNAPSHOT
|    |    +--- org.opensearch:opensearch-core:2.6.0-SNAPSHOT (*)
|    |    \--- org.opensearch:opensearch-core:2.6.0-SNAPSHOT (*)
|    |    +--- org.opensearch:opensearch-core:2.6.0-SNAPSHOT (*)
```
The "+" follows a prefix and I believe matches the newest version with that prefix, be it release or snapshot.  You can further filter it, but I think "newest" woroks here.

dbwiddis: For "1.+" I get:
```
➜  opensearch-sdk-java git:(revert_http_client) ✗ gradle dependencies | grep opensearch-core
|    +--- org.opensearch:opensearch-core:1.4.0-SNAPSHOT
|    |    +--- org.opensearch:opensearch-core:1.4.0-SNAPSHOT
|    |    \--- org.opensearch:opensearch-core:1.4.0-SNAPSHOT
```

ryanbogan: Ok so we can change `opensearchVersion` to `2.+`, but we would still need another variable for the tests right?

dbwiddis: From [the docs](https://docs.gradle.org/current/userguide/single_versions.html):
> A _prefix_ version range: e.g. 1.+, 1.3.+
> 
> Only versions exactly matching the portion before the + are included.
>
> The range + on it’s own will include any version.

That includes snapshots (assuming the snapshot repo is included).  Also,
> Versions are ordered based on the following rules:
>
> Each version is split into it’s constituent "parts":
> 
> The characters [. - _ +] are used to separate the different "parts" of a version.

and then

> The equivalent parts of 2 versions are compared using the following rules:
> 
> If both parts are numeric, the highest numeric value is higher: 1.1 < 1.2
> 
> If one part is numeric, it is considered higher than the non-numeric part: 1.a < 1.1
> 
> If both are non-numeric, the parts are compared alphabetically, in a case-sensitive manner: 1.A < 1.B < 1.a < 1.b
> 
> A version with an extra numeric part is considered higher than a version without (even when it’s zero): 1.1 < 1.1.0
> 
> A version with an extra non-numeric part is considered lower than a version without: 1.1.a < 1.1

So what's happening is it's breaking it into 2, 6, 0, SNAPSHOT which is higher than any 2, 5, anything.  When 2.6.0 is released it will be on the top until 2, 7, anything is available.

So just put "2+". :)

(We should probably switch "main" to just "+" ;)

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
dbwiddis: I'm going to guess the exact version doesn't matter and we can leave this as is.

dbwiddis (approved):

build.gradle
dbwiddis: I"m not sure the version in the tests matter (since 2.6 isn't even out).  Does the calling code check the version?

ryanbogan: Yeah it checks with the recent minimum compatibility check

ryanbogan: I just checked and it works with previous minor versions, since it's the minimum compatibility.  So we won't have to change this ever

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
dbwiddis: Actually there's an equals comparison with CURRENT, I think.

So this would have been a nice red flag that we needed to update our version, but why not just set these lines to `Version.CURRENT`?

ryanbogan: It's minimum compatibility so it'll pass as long as the OpenSearch version > than 2.6

ryanbogan: I don't see any reason not to change it to `Version.CURRENT` though, so I did

dbwiddis:
Interesting, your latest push didn't invalidate my review.... oh... 2.X branch :)

ryanbogan:
Yep I noticed the same thing, different rules for this branch.  I'm still gonna wait for a second approval though since this isn't a backport

dbwiddis (approved):

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -130,7 +104,7 @@ public OpenSearchClient initializeJavaClient(String hostAddress, int port) {
         mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
 
         // Create Client
-        OpenSearchTransport transport = new RestClientTransport(restClient, new JacksonJsonpMapper(mapper));
+        OpenSearchTransport transport = new RestClientTransport(restClient, new JacksonJsonpMapper());
```
owaiskazi19, line 107: Any reason to remove mapper here? We are registering modules above already

build.gradle
owaiskazi19, line 95: Let's create a variable for the client version too

ryanbogan: Done

owaiskazi19 (approved):

PR #435, opened 2023-02-11
Remove unnecessary exception throwing for settings update consumers

### LINKED ISSUE #422: [FEATURE] Handle exceptions internally in SDKClusterSettings addSettingsUpdateConsumer

Currently the `SDKClusterSettings` `addSetingsUpdateConusmer()` methods throw `Exception`. This requires them to be handled by the extension, which isn't really useful.

Handle the exception internally in the `SDKClusterSettings` methods.

dbwiddis:

### Description

`sendAddSettingsUpdateConsumerRequest()` was throwing an `Exception` (checked) for an empty map.  This required callers to handle the exception.

This was unnecessary as an unchecked `IllegalArgumentException` would have been a better type to use.  Actual exceptions during the call are handled by the response handler which logs them.

I went further and just ignored the empty map case and made it a no-op.  Happy to put an IllegalArgumentExxception back in if reviewers think it's needed.

### Issues Resolved

Fixes #422. (After merge, will submit a PR on AD repo to take advantage of no need for handling.)

### DIFF
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -472,37 +472,32 @@ public class ExtensionsRunner {
      *
      * @param transportService  The TransportService defining the connection to OpenSearch.
      * @param settingUpdateConsumers A map of setting objects and their corresponding consumers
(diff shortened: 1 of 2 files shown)


joshpalis (approved):
These changes LGTM

ryanbogan (approved):

Issue #368, opened 2023-01-26
[FEATURE] Action dependency injection


dbwiddis:
### Is your feature request related to a problem?

Extension Actions are triggered by Rest Requests, or some actions trigger other actions.  Extensions are being designed to trigger other extension's actions as well.

These actions require many common components when they are initialized.  See for example the AnomalyDetector plugin actions (being migrated to an extension) ([search for `@Inject`](https://github.com/search?q=repo%3Aopensearch-project%2Fanomaly-detection%20%40Inject&type=code)).  These actions all have a constructor similar to this (number and ordering of objects differs).
```java
@Inject
public IndexAnomalyDetectorTransportAction(
    TransportService transportService,
    ActionFilters actionFilters,
    Client client,
    ClusterService clusterService,
    Settings settings,
    AnomalyDetectionIndices anomalyDetectionIndices,
    NamedXContentRegistry xContentRegistry,
    ADTaskManager adTaskManager,
    SearchFeatureDao searchFeatureDao
) { ... }
```

Instantiating these `Action` objects requires these objects to be passed to the constructor.  
### What solution would you like?

There are essentially three options for doing this:
1. Every time the object is instantiated, pass the appropriate objects.  
   - This leads to a lot of duplication
   - I'm currently using this approach in migration work, and after duplicating this approach 3 times, I decided there had to be a better way and was motivated to create this issue asking about the other 2 options.
2. Use an existing dependency injection framework such as [Google Guice](https://github.com/google/guice).  
   - This is what is currently implemented (and the reason for the `@Inject` annotation), with the dependency management in OpenSearch via the [`createComponents()`](https://github.com/opensearch-project/OpenSearch/blob/5273a86cc2630cca98681ff1f2e364ca7ea030a5/server/src/main/java/org/opensearch/plugins/Plugin.java#L120-L154) extension point.  
     - Note that the (pre-fork) comments mention "migration away from Guice", so I'm not sure whether that's a current intention or why it was (and if it still is) desired.
     - They are [read in during bootstrap](https://github.com/opensearch-project/OpenSearch/blob/5273a86cc2630cca98681ff1f2e364ca7ea030a5/server/src/main/java/org/opensearch/node/Node.java#L758) and then [injected into Guice](https://github.com/opensearch-project/OpenSearch/blob/5273a86cc2630cca98681ff1f2e364ca7ea030a5/server/src/main/java/org/opensearch/node/Node.java#L1096-L1102).
[...]

PR #436, opened 2023-02-11
Add Guice dependency and inject ExtensionsRunner into Extension

dbwiddis:

### Description

Use Guice for dependency injection.  

First of two or more PRs, splitting up into logical chunks to speed review.  

This PR just adds the dependency and initial injection setup.  Future PRs:
 - will integrate with `getActions()` implementation (pending merge of #434) to improve action handling
 - will add additional components necessary for AD extension 

This PR:
 - Injects (via lambda) a Guice `Module` which binds classes and makes them available via `@Inject`. 
 - Binds other objects associated with `createComponents()` (more to come in the future)
 - Injects the `ExtensionsRunner` into the `BaseExtension` as a sample/test case
 - Makes the `createComponents()` method on the `Extension` interface default, removing its arguments and removing the requirement for `BaseExtension` to mange/provide these objects (as they will be injected and available directly)

### Issues Resolved

Part of #368 


joshpalis (approved):
This looks great, thanks for implementing Guice within the project, this will definitely simply integration later down the line.

ryanbogan (dismissed):

joshpalis (approved):

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -154,10 +164,34 @@ protected ExtensionsRunner(Extension extension) throws IOException {
         // save custom transport actions
         this.transportActions = new TransportActions(extension.getActionsMap());
 
-        ThreadPool threadPool = new ThreadPool(this.getSettings());
+        Guice.createInjector(b -> {
+            b.bind(ExtensionsRunner.class).toInstance(this);
+            b.bind(Extension.class).toInstance(extension);
+
+            b.bind(NamedXContentRegistry.class).toInstance(getNamedXContentRegistry().getRegistry());
+            b.bind(ThreadPool.class).toInstance(getThreadPool());
```
owaiskazi19, line 172: Need to understand this. Do we need ThreadPool in SDK?

dbwiddis: This is SDK's own thread pool.  

Do we need it?  No.  Do we want a single common thread pool that the whole extension can share if it so chooses, to balance its resources?  Why not?

dbwiddis: Also from a PR standpoint, this does not change the previous behavior of a threadpool being created (line 157) and passed to the extension via create components.  Just changes the syntax of how an extension accesses that object.

owaiskazi19: Makes sense. Do we need to bind ThreadPool then, as it is specific to the SDK?

ryanbogan (approved):

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
dbwiddis: In theory we might make this object available for in-proc extensions to be passed as a param.  I don't know, but we can always easily remove it before GA.

owaiskazi19 (approved):

PR #452, opened 2023-02-14
Replace jackson YAML reading with snakeyaml

ryanbogan:
### Description
YAML reading for extension.yml used to utilize the jackson-databind library.  As part of an initiative to remove jackson-databind, this process was replaced with similar logic using the snakeyaml library.

### DIFF
--- a/src/main/java/org/opensearch/sdk/ExtensionSettings.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionSettings.java
@@ -107,7 +107,6 @@ public class ExtensionSettings {
         }
         try (InputStream inputStream = Files.newInputStream(Path.of(resource.toURI()))) {
             Map<String, Object> extensionMap = yaml.load(inputStream);
-            inputStream.close();
             if (extensionMap == null) {
                 throw new IOException("extension.yml is empty");
             }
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -96,12 +100,26 @@ public String toString() {
...
     public static ExtensionSettings readSettingsFromYaml(String extensionSettingsPath) throws IOException {
+        Yaml yaml = new Yaml(new SafeConstructor());
         URL resource = Extension.class.getResource(extensionSettingsPath);
         if (resource == null) {
-            return null;
+            throw new IOException("extension.yml does not exist at path [" + extensionSettingsPath + "]");
+        }
+        try (InputStream inputStream = Files.newInputStream(Path.of(resource.toURI()))) {
+            Map<String, Object> extensionMap = yaml.load(inputStream);
+            inputStream.close();
```
dbwiddis, line 110: This should be `flush()`.   

The `close()` method does not automatically flush (in fact the Javadocs say it [does nothing](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/InputStream.html#close())), and you're using try-with-resources so it will be "closed" automatically at the end of the try block.

dbwiddis (changes requested):

dbwiddis (approved):

joshpalis (approved):

PR #454, opened 2023-02-15
Changed the sequence of Guice injection

owaiskazi19:
### Description
Fix for the NPE caused by Guice injection which is done before invoking Rest Handlers on the extension's side.

### Issues Resolved


dbwiddis (approved):
LGTM for a quick fix.  In the long term I think if we inject the handlers it would also work, but that can wait for another PR

saratvemulapalli (approved):

PR #458, opened 2023-02-16
#508

varuntumbe:
Signed-off-by: varuntumbe1 <varuntumbe1@gmail.com>

Companion PR: https://github.com/opensearch-project/OpenSearch/pull/6191

### Description

### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -303,6 +303,15 @@ public class ExtensionsRunner {
         this.extensionNode = extensionNode;
     }
 
+    /**
+     * Returns the discovery extension node set during extension initialization
+     *
+     * @return the extensionNode
+     */
+    public DiscoveryExtensionNode getExtensionNode() {
+        return this.extensionNode;
+    }
+
     public DiscoveryNode getOpensearchNode() {
         return this.opensearchNode;
     }
--- a/src/test/java/org/opensearch/sdk/sample/helloworld/TestHelloWorldExtension.java
+++ b/src/test/java/org/opensearch/sdk/sample/helloworld/TestHelloWorldExtension.java
@@ -33,6 +33,7 @@ import org.opensearch.sdk.ExtensionRestHandler;
 import org.opensearch.sdk.ExtensionSettings;
 import org.opensearch.sdk.ExtensionsRunner;
 import org.opensearch.sdk.SDKClient;
+import org.opensearch.sdk.SDKClient.SDKRestClient;
 import org.opensearch.sdk.action.SDKActionModule;
 import org.opensearch.test.OpenSearchTestCase;
 import org.opensearch.threadpool.ThreadPool;
@@ -45,6 +46,7 @@ public class TestHelloWorldExtension extends OpenSearchTestCase {
     private HelloWorldExtension extension;
     private Injector injector;
     private SDKClient sdkClient;
+    private SDKRestClient sdkRestClient;
 
     static class UnregisteredAction extends ActionType<SampleResponse> {
         public static final String NAME = "helloworld/unregistered";
@@ -71,12 +73,14 @@ public class TestHelloWorldExtension extends OpenSearchTestCase {
             b.bind(SDKClient.class);
         });
         this.sdkClient = injector.getInstance(SDKClient.class);
+        this.sdkRestClient = sdkClient.initializeRestClient("localhost", 9200);
     }
 
     @Override
     @AfterEach
     public void tearDown() throws Exception {
         super.tearDown();
+        this.sdkRestClient.close();
         this.injector = null;
     }
 
@@ -127,6 +131,30 @@ public class TestHelloWorldExtension extends OpenSearchTestCase {
         assertEquals(expectedGreeting, response.getGreeting());
     }
 
+    @SuppressWarnings("deprecation")
+    @Test
+    public void testRestClientExecuteSampleActions() throws Exception {
+        String expectedName = "world";
+        String expectedGreeting = "Hello, " + expectedName;
+        SampleRequest request = new SampleRequest(expectedName);
+        CompletableFuture<SampleResponse> responseFuture = new CompletableFuture<>();
+
+        sdkRestClient.execute(SampleAction.INSTANCE, request, new ActionListener<SampleResponse>() {
+            @Override
+            public void onResponse(SampleResponse response) {
+                responseFuture.complete(response);
+            }
+
+            @Override
+            public void onFailure(Exception e) {
+                responseFuture.completeExceptionally(e);
+            }
+        });
+
+        SampleResponse response = responseFuture.get(1, TimeUnit.SECONDS);
+        assertEquals(expectedGreeting, response.getGreeting());
+    }
+
     @Test
     public void testExceptionalClientExecuteSampleActions() throws Exception {
         String expectedName = "";
@@ -134,6 +162,31 @@ public class TestHelloWorldExtension extends OpenSearchTestCase {
         CompletableFuture<SampleResponse> responseFuture = new CompletableFuture<>();
 
         sdkClient.execute(SampleAction.INSTANCE, request, new ActionListener<SampleResponse>() {
+            @Override
+            public void onResponse(SampleResponse response) {
+                responseFuture.complete(response);
+            }
+
+            @Override
+            public void onFailure(Exception e) {
+                responseFuture.completeExceptionally(e);
+            }
+        });
+
+        ExecutionException ex = assertThrows(ExecutionException.class, () -> responseFuture.get(1, TimeUnit.SECONDS));
+        Throwable cause = ex.getCause();
+        assertTrue(cause instanceof IllegalArgumentException);
+        assertEquals("The request name is blank.", cause.getMessage());
+    }
+
+    @SuppressWarnings("deprecation")
+    @Test
+    public void testExceptionalRestClientExecuteSampleActions() throws Exception {
+        String expectedName = "";
+        SampleRequest request = new SampleRequest(expectedName);
+        CompletableFuture<SampleResponse> responseFuture = new CompletableFuture<>();
+
+        sdkRestClient.execute(SampleAction.INSTANCE, request, new ActionListener<SampleResponse>() {
             @Override
             public void onResponse(SampleResponse response) {
                 responseFuture.complete(response);
--- a/src/main/java/org/opensearch/sdk/SDKClient.java
+++ b/src/main/java/org/opensearch/sdk/SDKClient.java
@@ -47,6 +47,8 @@ import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
 import org.opensearch.action.support.TransportAction;
 import org.opensearch.action.support.master.AcknowledgedResponse;
+import org.opensearch.action.update.UpdateRequest;
+import org.opensearch.action.update.UpdateResponse;
 import org.opensearch.client.Cancellable;
 import org.opensearch.client.Client;
 import org.opensearch.client.ClusterAdminClient;
@@ -200,7 +202,7 @@ public class SDKClient implements Closeable {
      */
     @Deprecated
     public SDKRestClient initializeRestClient(String hostAddress, int port) {
-        return new SDKRestClient(new RestHighLevelClient(builder(hostAddress, port)));
+        return new SDKRestClient(this, new RestHighLevelClient(builder(hostAddress, port)));
     }
 
     /**
@@ -232,13 +234,13 @@ public class SDKClient implements Closeable {
     }
 
     /**
-     * Executes a Transport Action
+     * Executes a generic action, denoted by an {@link ActionType}.
      *
-     * @param <Request> The Request type for the action
-     * @param <Response> The Response type for the action
-     * @param action The registered action
-     * @param request The Request
-     * @param listener An action listener for the Response
+     * @param action The action type to execute.
+     * @param request The action request.
+     * @param listener The listener to receive the response back.
+     * @param <Request> The request type.
+     * @param <Response> The response type.
      */
     public final <Request extends ActionRequest, Response extends ActionResponse> void execute(
         ActionType<Response> action,
@@ -266,14 +268,17 @@ public class SDKClient implements Closeable {
     @Deprecated
     public static class SDKRestClient implements Closeable {
 
+        private final SDKClient sdkClient;
         private final RestHighLevelClient restHighLevelClient;
 
         /**
          * Instantiate this class wrapping a {@link RestHighLevelClient}.
          *
+         * @param sdkClient The SDKClient instance.
          * @param restHighLevelClient The client to wrap.
          */
-        public SDKRestClient(RestHighLevelClient restHighLevelClient) {
+        public SDKRestClient(SDKClient sdkClient, RestHighLevelClient restHighLevelClient) {
+            this.sdkClient = sdkClient;
             this.restHighLevelClient = restHighLevelClient;
         }
 
@@ -289,6 +294,23 @@ public class SDKClient implements Closeable {
          */
         public SDKClusterAdminClient cluster() {
             return new SDKClusterAdminClient(restHighLevelClient.cluster());
+        }
+
+        /**
+         * Executes a generic action, denoted by an {@link ActionType}.
+         *
+         * @param action The action type to execute.
+         * @param request The action request.
+         * @param listener The listener to receive the response back.
+         * @param <Request> The request type.
+         * @param <Response> The response type.
+         */
+        public final <Request extends ActionRequest, Response extends ActionResponse> void execute(
+            ActionType<Response> action,
+            Request request,
+            ActionListener<Response> listener
+        ) {
+            this.sdkClient.execute(action, request, listener);
         }
 
         /**
@@ -330,6 +352,16 @@ public class SDKClient implements Closeable {
          */
         public void multiGet(MultiGetRequest request, ActionListener<MultiGetResponse> listener) {
             restHighLevelClient.mgetAsync(request, RequestOptions.DEFAULT, listener);
+        }
+
+        /**
+         * Updates a document based on a script.
+         *
+         * @param request  The update request
+         * @param listener A listener to be notified with a result
+         */
+        public void update(UpdateRequest request, ActionListener<UpdateResponse> listener) {
+            restHighLevelClient.updateAsync(request, RequestOptions.DEFAULT, listener);
         }
 
         /**
--- a/src/main/java/org/opensearch/sdk/SDKClusterService.java
+++ b/src/main/java/org/opensearch/sdk/SDKClusterService.java
@@ -14,6 +14,7 @@ import java.util.function.Consumer;
 
 import org.opensearch.cluster.ClusterState;
 import org.opensearch.common.settings.Setting;
+import org.opensearch.extensions.DiscoveryExtensionNode;
 
 /**
  * This class simulates methods normally called from OpenSearch ClusterService class.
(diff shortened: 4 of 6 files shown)


varuntumbe:
Hi @dbwiddis ,

I will ad the spotlessApply and other minor things tomorrow. Can you please check functionality wise anything I should change ?

Thanks and regards
Varun

src/main/java/org/opensearch/sdk/Extension.java
```diff
@@ -84,4 +85,14 @@ default Collection<Object> createComponents() {
...
         return Collections.emptyList();
     }
+
+    /**
+     * Provides the list of implimented interfaces by an extension, empty if
+     * none.
+     *
+     * @return list of Interfaces implimented by extension
+     */
+    default List<ExtensionsManager.ExtensionInterfaceType> getImplimentedInterfaces() {
```
dbwiddis, line 95: We don't need a separate extension point for this.  The extension will have done `implements Extension, ActionExtension, ... ` so Java has this information available to it.

A brute force way to do it is just a long set of `if (extension instanceof ActionExtension) { .... add ACTION to enumset ... }`.   

Another option (I'll refrain from saying it's more elegant) is to iterate over `extension.getClass().getInterfaces()` and do a little bit of parsing to translate the class type to the enum (truncate the before the word `Extension` and [convert camelcase to underscorecase](http://www.java2s.com/example/java-utility-method/string-camel-case-to-underscore/camelcasetounderscores-string-camel-02556.html)).

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -392,6 +400,26 @@ public void sendRegisterCustomSettingsRequest(TransportService transportService)
         }
     }
 
+    /**
+     * Requests that OpenSearch register all the Interfaces implimented by the extension.
+     *
+     * @param transportService  The TransportService defining the connection to OpenSearch.
+     */
+    public void sendRegisterImplimentedInterfacesRequest(TransportService transportService) {
```
dbwiddis, line 408: I made a comment over on the OpenSearch PR that I don't think we need to do a separate request for this.  During extension initialization, OpenSearch sends an `ExtensionInitRequest` and I think including the values in the response to that is a better implementation point.

src/main/java/org/opensearch/sdk/sample/helloworld/HelloWorldExtension.java
```diff
@@ -51,6 +53,11 @@ public List<ExtensionRestHandler> getExtensionRestHandlers() {
         return List.of(new RestHelloAction());
     }
 
+    @Override
+    public List<ExtensionsManager.ExtensionInterfaceType> getImplimentedInterfaces() {
+        return List.of(ExtensionInterfaceType.ACTION,ExtensionInterfaceType.CIRCUITBREAKER);
```
dbwiddis, line 58: We don't want to change HelloWorld to this if it doesn't actually implement these extensions.  We can add them to a test class though.

dbwiddis (commented):
Comments inline!

src/main/java/org/opensearch/sdk/Extension.java
varuntumbe: Yes yes, the second option is elegant. I will implement that way.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
varuntumbe: That is the first request sent by opensearch once it gets to know what all extensions are registered I believe. Cool i will try to add this along with whatever sdk is sending for this request.

src/main/java/org/opensearch/sdk/sample/helloworld/HelloWorldExtension.java
varuntumbe: Got it. I just added this, just to let you know that values are coming to the other side. Will put it in testcase side

src/main/java/org/opensearch/sdk/Extension.java
dbwiddis: > Yes yes, the second option is better. I will implement that way.

So as usual in the process of going through code we find a better way.  I am also thinking all this parsing of class name to enum isn't really needed either.  Ultimately all we want to do is an `implementedExtensions.contains(interfaceName)` and we can do that with a string as easily as an enum.  

Which means you could just take that `getInterfaces()` result and pass it.  (Probably don't need the package name prefix on it.).  So we'd just do `implementedExtensions.contains("ActionExtension")` when we wanted to check that, for example.

dbwiddis: And thinking more on this, I see a flaw in that if someone implements their own interface that's not part of the SDK but has the same name, it'd trip us up unless we used the fully qualified class name.  Which ends up being a lot more big long strings being passed around everywhere, and I'm thinking the brute force `instanceof` approach may be the best.

varuntumbe: Hi @dbwiddis ,

Yes these will create problem. Hadn't thought of it before. But what if we just check whether Interfaces belonging to org.opensearch.sdk and only send those ? ( I am imagining developer who builds the extension would have placed his interface in different package ). 

Yes now that I think of it, List<String> is better than List<Enum> as even after we store, while comparing, it will be trouble. So I have changed it to String  so we can just use .contains() on implementedInterfaces of DiscoverExtension.

I will commit these and I have added one testcase as well. Tell me what else can be improved.

Thanks and Regards
Varun

varuntumbe: As of now I have modified the HelloWorldExtension class, but for testcases may be I should define some private extension classes which impliments Engine and Search Extension ? ( we can add more test scenarios )

Thanks and Regards
Varun

dbwiddis: Sorry, trying to track both PRs and not sure where we are or if you're waiting on my review.

I think we're going with a list (or set) of String interface names, filtering with `org.opensearch.sdk` prefix?

varuntumbe: Hi @dbwiddis ,

Yes currently I have implemented like this. I can change to method of just checking with **instaceof** approach if you like. And I have updated the other PR as well. I believe once both of the PR gets ready, We can update on both the side simultaneously as we are modifying ExtensionInitResponse method structure. 

I will change HelloWorldExtension, I had just changed it just to test once. Apart from it Any other changes you want me to do ? ( In both here as well as other PR )

Also, I did not understand what you meant in the first comment review **Comments inline!**.

Thanks and regards
Varun.

src/main/java/org/opensearch/sdk/sample/helloworld/HelloWorldExtension.java
varuntumbe: I have removed it

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
varuntumbe: I have added the response with ExtensionInitResponse

src/main/java/org/opensearch/sdk/Extension.java
dbwiddis: OK, looks good, I like this approach!  Reviewed the companion PR as well.   Some suggestions:
 - rather than package name string equality use `equals` or possibly a `startsWith` as we may move the extension interfaces into subpackages and that string wouldn't auto-update.  Might even be good to get the package location of one of the known-good interfaces (`Extnsion`) automatically instead of hardcoding a string, so as long as all the extension interfaces are in the same package, anywhere, it works.
 - for the `Writeable` implementations, there are `readStringArray` and `writeStringArray` methods that do the same logic you are with the vInt, and you can go to/from list to array easily. I'd suggest that.
 - The openSearch side will need the test case improved for the class you're editing

By "comments inline" I meant this comment thread we're on, rather than giving a summary comment. It's usually used in email when you add comments in the middle of the text you're replying to, and adding comments in the middle of code is kind of the same thing.  You can ignore it.

I'm on vacation until next Sunday. Will probably poke around on these PRs in the evening but responses will be slower than usual.  Thanks for all the work you're doing.

varuntumbe: Hi @dbwiddis,

- Yes good point I will replace the current one with startswith. That will be better.
- I changed both writeTo and constructor to use readStringArray and writeStringArray.
- I looked at the testcase of ExtensionsManager (TestExtensionsManager), the method we are interested in is **testInitialize**, where current implimentation is they just create ExtensionsManager instance with the yml file which they have defined and they try to initalize, and they just assert the exception logs which they get as initialize will try to send request to sdk node ( they have not mocked it yet ). there is one issue raised on behalf of it. should I try to resolve this https://github.com/opensearch-project/OpenSearch/issues/4045? only then I can actually test the interfaces coming from the response.

Reply only if you get time and enjoy the vacation !!. I might need some time to write this testcase. 

Thanks and regards
Varun.

dbwiddis: Hmm, looks like that `InitializeExtensionResponse` (and the corresponding `InitializeExtensionRequest` got missed in unit testing.   Would be nice to either add a new test class in `org.opensearch.discovery` class to test these, or add a test for them in the `ExtensionManagerTests` class where there are plenty of examples exercizing the writable streaminput / streamoutput stuff.

varuntumbe: Yes, I will write unit tests for both and that for extensionsmanager as well.

Thanks
Varun

varuntumbe: Hi @dbwiddis ,

I have done all the changes. I have taken actionExtension Interface as a reference to compare and I have used .startswith() function and I have added a testcase for InitExtensionResponse ( where I have changed ). InitExtensionRequest involves transportService which I need to understand better to write it. Testcase of initialize of ExtensionsManager also needs more time ( Since I need to mock sdk node ). I thought I will come to these again later ( Especially the one with mocking sdk node ).

Tell me if its okay else I can spend more time for writing testcase for InitExtensionRequest if its necessary. Also tell me if any other changes needs to be done.

Thanks and Regards
Varun.

src/test/java/org/opensearch/sdk/sample/helloworld/TestHelloWorldExtension.java
```diff
@@ -49,4 +51,17 @@ public void testExtensionRestHandlers() {
...
+    public void testImplimentedInterfacesByExtension() {
+        Class<?>[] interfaces = extension.getClass().getInterfaces();
+        List<String> interfacesOfOpenSearch = new ArrayList<String>();
+        for (Class<?> anInterface : interfaces) {
+            if (anInterface.getPackageName() == "org.opensearch.sdk") {
+                interfacesOfOpenSearch.add(anInterface.getSimpleName());
+            }
+        }
+
+        assertEquals(interfacesOfOpenSearch, new ArrayList<String>(Arrays.asList("EngineExtension")));
```
dbwiddis, line 64: I'm a little confused here, as `HelloWorldExtension` extends `BaseExtension` which implements `Extension` and `ActionExtension`, so we shouldn't see `EngineExtension` in the list here and this test case should fail.

More importantly, this test case is just repeating the internal code for another method, which doesn't really test that method.  I'd expect to see `List<String> interfacesOfOpenSearch = someMethodThatReturnsTheList()`.

Since that code is in the handler which has a field for the `ExtensionsRunner`, which itself has a field for the `Extension`, you may want to make the `ExtensionsRunner` do that code logic and return the list in a method which you test here.  I think `getExtensionImplementedInterfaces()` is a good name.

src/main/java/org/opensearch/sdk/sample/helloworld/HelloWorldExtension.java
```diff
@@ -13,6 +13,7 @@
 import java.util.List;
 
 import org.opensearch.sdk.BaseExtension;
+import org.opensearch.sdk.EngineExtension;
```
dbwiddis, line 16: I see this import but don't see where it's used?

src/main/java/org/opensearch/sdk/sample/helloworld/HelloWorldExtension.java
```diff
@@ -28,7 +29,7 @@
  * <p>
  * To execute, pass an instatiated object of this class to {@link ExtensionsRunner#run(Extension)}.
  */
-public class HelloWorldExtension extends BaseExtension {
+public class HelloWorldExtension extends BaseExtension{
```
dbwiddis, line 32: This looks like something spotless would complain about

src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
```diff
@@ -49,9 +54,21 @@ public InitializeExtensionResponse handleExtensionInitRequest(InitializeExtensio
         logger.info("Registering Extension Request received from OpenSearch");
         extensionsRunner.opensearchNode = extensionInitRequest.getSourceNode();
         extensionsRunner.setUniqueId(extensionInitRequest.getExtension().getId());
+
```
dbwiddis, line 57: per previous comment on the test case, I think all this new code would fit nicely in a method on the ExtensionsRunner

src/main/java/org/opensearch/sdk/Extension.java
```diff
@@ -20,7 +20,7 @@
 import org.opensearch.common.settings.Setting;
 import org.opensearch.common.xcontent.NamedXContentRegistry;
 import org.opensearch.threadpool.ExecutorBuilder;
-import org.opensearch.common.settings.Settings;
+import org.opensearch.common.settings.Settings;;
```
dbwiddis, line 23: Added an extra semicolon here. Since the only other change to this file is whitespace you may want to just revert to the original (upstream/HEAD) version.

dbwiddis (changes requested):
Good logic, would like it moved to its own method (so you can test it). 

Looks like a few formatting issues that spotless would fix, and there are some merge conflicts, you probably need to rebase with main.

src/main/java/org/opensearch/sdk/sample/helloworld/HelloWorldExtension.java
varuntumbe: Yes, I had used it to test locally and have removed it but import was left to remove. I have removed it now

src/test/java/org/opensearch/sdk/sample/helloworld/TestHelloWorldExtension.java
varuntumbe: Yes that makes more sense. I have added that logic to ExtensionRunner and have added test method in TestExtensionRunner.java but I have 2 doubts here.

1)  I see there that they are using ExtensionRunnerForTest class as extensionRunner( which extends extensionRunner class ) and while constructing it, we construct its super class by providing instance of BaseExtension as param. How can we instantiate an abstract class ? ( BaseExtension is declared as an abstract class )

2) when I debug the method testGetExtensionImplementedInterfaces, Ideally getExtension() method of ExtensionRunner should getinvoked as its not been overridden in ExtensionRunnerForTestClass. so it should give me the instance of BaseExtension which impliments ActionExtesion, But I am getting ExtensionRunnerForTest$1@7829 as an object and getting an empty list ( implementedInterfaces ) as output.

so as of now as coverage wise we are good. testGetExtensionImplementedInterfaces teste the method and I am asserting whether returned list is empty, so the test is passing.

I might lack some knowledge here, So whenever you get time after you comeback from vacation, help me resolve these doubts. 


Thanks

src/main/java/org/opensearch/sdk/Extension.java
varuntumbe: done

src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
varuntumbe: Yes, changes are pushed

src/test/java/org/opensearch/sdk/sample/helloworld/TestHelloWorldExtension.java
dbwiddis: For 1.  You can instantiate any interface or abstract class like `new FooInterface() { @Override implemented methods here };`.  Although I think between the `ExtensionRunnerForTest` which internally has the extension extending `BaseExtension` or `HelloWorldExtension` (see `TestHelloWorldExtension` class) you should be able to test that method.

For 2. Should be a `getExtension()` method on the extensions runner that returns the extension?

dbwiddis: When I originally answered I hadn't seen your code, but now that I have I see what's going on.  The `ExtensionsRunnerForTest` class is creating an anonymous subclass of `BaseExtension`. That's what you see.

And on further reading the docs, the `getInterfaces()` only gets the interfaces of that subclass, which aren't any; it doesn't `implement` anything, it `extends BaseExtension`.  So what we really need to do is use its `superClass()` method to retrieve `BaseExtension` which should have the appropriate extensions.

This needs to be done recursively of course, for multiple levels of inheritance.  So your code should:
 - create a collection (Set, probably) of classes
 - extract the `Class` of the `extension`
 - call `getInterfaces()` and add all (possibly empty) to the set
 - call `getSuperClass()`. 
   - If you get a `null` return, you've hit the `Object` class, and you're done
   - otherwise, return to the `getInterfaces()` step, add to the set, and check the next superclass

Writing code from memory without an IDE, entirely untested, but hopefully explaining the above logic... it might work, or not, and you might need to add generics like `Class<?>` for the collection API, feel free to write your own version!
```
Set<Class> interfaceSet = new HashSet<>();
Class clazz = extension.getClass();
do {
  interfaceSet.addAll(clazz.getInterfaces());
  clazz = clazz.getSuperclass();
} while (clazz != null);
```

varuntumbe: Hi @dbwiddis ,

Understood now..  yes the approach which you told perfectly covers the case. amazing that so many corner cases will be there in a small feature. thanks for your input. I will change this appropriately. Currently I am drowned in resolving merge conflicts. I should start a habit of regularly syncing my repo with central and pulling it to local.

Thanks and regards
Varun.

varuntumbe: Hi @dbwiddis ,

I have added the change but rebasing taking a lot of time.. there are around 8 to 9 commits to this branch when it was created and I had not merged the latest now getting lot of merge conflicts while rebasing and having little confusion which changes to take and which not to take.. I will update you once its done. In the extreme case If I am not able to rebase successfully , I have the diff of my change, Is it okay If I create fresh PR with latest ?

Thanks and Regards
Varun

dbwiddis: Totally fine to start with a fresh PR!

src/test/java/org/opensearch/sdk/TestExtensionsRunner.java
```diff
@@ -102,7 +103,7 @@ public void testStartTransportService() {
         verify(transportService, times(5)).registerRequestHandler(anyString(), anyString(), anyBoolean(), anyBoolean(), any(), any());
     }
 
-    @Test
+    // @Test
```
ryanbogan, line 106: Is there a reason this is commented out?

varuntumbe: oops sorry @ryanbogan , I thought it was commented before and I had uncommented it locally to test. Looks like I was wrong. I will revert back this change

Thanks and Regards
Varun

varuntumbe: Hi @ryanbogan,

Actually this PR is stale. I had created a new PR as this was old and had some 15+ commits behind from main and was giving lot of merge conflicts . To save the pain of rebasing it, I had created a altogether fresh PR here. https://github.com/opensearch-project/opensearch-sdk-java/pull/508. Can you please review this ?

I will delete this PR if its okay.

Thanks and Regards
Varun

ryanbogan: Sure! I'll take a look now

PR #468, opened 2023-02-19
Create SDKNamedXContentRegistry class and provide updated registry

### LINKED ISSUE #447: [BUG] ExtensionNamedXContent registry initialization is sticky
### What is the bug?

When an extension is initialized, it registers the extension's Rest Handlers:
https://github.com/opensearch-project/opensearch-sdk-java/blob/514f796049ec693dddbd44bf3ef3eabee8869f52/src/main/java/org/opensearch/sdk/ExtensionsRunner.java#L152-L157

This passes the current state of the `ExtensionsRunner` to extensions, which often in their constructors do something like:
```java
    public RestIndexAnomalyDetectorAction(ExtensionsRunner extensionsRunner, AnomalyDetectorExtension anomalyDetectorExtension) {
        super(extensionsRunner);
        this.namedXContentRegistry = extensionsRunner.getNamedXContentRegistry().getRegistry();
```
This registry doesn't include the extension's `NamedXContent` which is only updated after extension initialization.

Under the Guice injection model, this could be improved except that we are essentially doing the same thing, injecting the pre-initiialization version:
https://github.com/opensearch-project/opensearch-sdk-java/blob/514f796049ec693dddbd44bf3ef3eabee8869f52/src/main/java/org/opensearch/sdk/ExtensionsRunner.java#L171

Attempt to parse an AD object using the above constructor setup.

Current values of NamedXContent

Suggested fix:
 - Pass the `ExtensionNamedXContnentRegistry` around, not the `getRegistry()`.
 - Change the above registry to a singleton object and update its internal registry
 - Inject the extension registry, not the internal registry, which should be extracted on-demand when needed.

dbwiddis:
### Description

1. Renames `ExtensionNamedXContentRegistry` to `SDKNamedXContentRegistry`
2. Uses the `ExtensionsRunner` as an initialization and update parameter
3. Reorders initialization so the initial registry copy in extensions will have the needed content if it's saved
4. Injects the SDK registry rather than the wrapped value so developers have a means to access the updated registry
5. Updates migration guide

### Issues Resolved

Fixes #447 

### DIFF
--- a/PLUGIN_MIGRATION.md
+++ b/PLUGIN_MIGRATION.md
@@ -42,12 +42,20 @@ Change the transport action inheritance from HandledTransportAction to directly 
 
 Pass the `ExtensionsRunner` and `Extension` objects to the handler and access `createComponent` equivalents, such as:
 ```java
-this.namedXContentRegistry = extensionsRunner.getNamedXContentRegistry().getRegistry();
+this.sdkNamedXContentRegistry = extensionsRunner.getNamedXContentRegistry();
+```
+
+When a `NamedXContentRegistry` object is required, get the current one from `this.sdkNamedXContentRegistry.getRegistry()`.
+
+Other potential initialization values:
+```java
 this.environmentSettings = extensionsRunner.getEnvironmentSettings();
 this.transportService = extensionsRunner.getExtensionTransportService();
 this.restClient = anomalyDetectorExtension.getRestClient();
 this.sdkClusterService = new SDKClusterService(extensionsRunner);
 ```
+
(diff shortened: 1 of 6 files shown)


joshpalis (approved):

owaiskazi19 (approved):

dbwiddis:
@owaiskazi19 something I just learned (after this happening 2x in 24 hours), when PRs are competing with possible merge conflicts like this one, we should make sure we merge the `1.x` backport of the first one before we merge the `main` version of the competing one.

PR #472, opened 2023-02-20
Bumping version of main and publishing snapshots after every push

owaiskazi19:
### Description
Cron job is currently scheduled for 12 hrs and with changes coming in for SDK repo, snapshot should be published after every push to be available for migration.

### Issues Resolved
Fixes https://github.com/opensearch-project/opensearch-sdk-java/issues/473

### DIFF
--- a/.github/workflows/publish-snapshots.yml
+++ b/.github/workflows/publish-snapshots.yml
@@ -2,8 +2,10 @@ name: Publish snapshots to maven
 
 on:
   workflow_dispatch:
-  schedule:
-    - cron: 0 */12 * * *
+  push:
+    branches: [
+        main
+    ]
 
 jobs:
   build-and-publish-snapshots:
--- a/.github/workflows/publish-snapshots.yml
+++ b/.github/workflows/publish-snapshots.yml
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - main
-      - 1.x
 
 jobs:
   build-and-publish-snapshots:
--- a/.github/workflows/publish-snapshots.yml
+++ b/.github/workflows/publish-snapshots.yml
@@ -3,9 +3,9 @@ name: Publish snapshots to maven
 on:
   workflow_dispatch:
   push:
-    branches: [
-        main
-    ]
+    branches:
+      - main
+      - 1.x
 
 jobs:
   build-and-publish-snapshots:
--- a/build.gradle
+++ b/build.gradle
@@ -37,7 +37,7 @@ mainClassName = 'org.opensearch.sdk.sample.helloworld.HelloWorldExtension'
 
 
 group 'org.opensearch.sdk'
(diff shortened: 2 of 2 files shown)


ryanbogan (approved):

joshpalis:
Let's not backport this change to `1.x` branch, since this will publish `main`. Instead we should have the `1.x` branch publish itself as `1.0.0-SNAPSHOT`, and main as `2.0.0-SNAPSHOT`

joshpalis (approved):

.github/workflows/publish-snapshots.yml
```diff
@@ -2,8 +2,11 @@ name: Publish snapshots to maven
 
 on:
   workflow_dispatch:
-  schedule:
-    - cron: 0 */12 * * *
+  push:
+    branches: [
```
dbwiddis, line 6: eww, I don't like this multiline array syntax.  Either do 
```yaml
branches: [ main, 1.x ]
```
or
```yaml
branches:
  - main
  - 1.x
```

build.gradle
```diff
@@ -37,7 +37,7 @@ mainClassName = 'org.opensearch.sdk.sample.helloworld.HelloWorldExtension'
 
 
 group 'org.opensearch.sdk'
-version '1.0.0-SNAPSHOT'
+version '2.0.0-SNAPSHOT'
```
dbwiddis, line 40: we need a conditional on here to set the version depending on which branch is running the workflow

dbwiddis (changes requested):

build.gradle
gaiksaya: The GHA takes care of checking out the right commit. https://github.com/actions/checkout#usage
```
    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
```
so if the event of trigger is 1.x branch, that commit on 1.x is checked out which is already pointing at 1.0.0

gaiksaya (approved):

build.gradle
dbwiddis: I'm skeptical.  This workflow runs on a 1.x branch trigger, so I don't think it's going to switch to run the workflow on that branch, it will keep running this one.  The code will get checked out at 1.x, but the variables in the workflow being executed won't change.

dbwiddis: I think if you want this to publish 2.0 you have to limit it to the `main` branch.   And the workflow in #474 does the same on `1.x` branch.  If that works it's the simplest solution.    Otherwise one single workflow (backported) with the logic here as I commented.  This current PR state mixes both ideas.

owaiskazi19: I have removed `1.x` from main's CI file.

joshpalis (approved):

dbwiddis (approved):

Issue #467, opened 2023-02-19
[FEATURE] Replace TransportActions class functionality in SDKActionModule


dbwiddis:
### Is your feature request related to a problem?

The `TransportActions` class contains an internal map that must be provided at instantiation. This map contains the same information in `SDKActionModule.getActions()` in a different format and is only relevant for an `ActionExtension`. 

This requires unnecessary conditionals outside the class.

### What solution would you like?

1. Move the `sendRegisterTransportActionsRequest` method to the `SDKActionModule`
2. Use the `getActions()` method of `SDKActionModule` to retrieve the necessary map
3. Change the registration method to send only the necessary string info and not a full class
~4. Change the handler for actions sent from OpenSearch to use the `SDKActionModule` `getActions()` map (or use the injected `transportAction(instance)` lookup, or other approach)~ This handler doesn't exist. Opened #525 

### What alternatives have you considered?

Creating the map outside the class, conditionally populating it, and then initializing it, which was done in #466

Adding a getter for the map to `TransportActions` so it can be initialized once externally (but this seems silly).

### Do you have any additional context?

Consider whether we need this separate map signature and whether the values from `SDKActionModule.getActions()` are sufficient to serve the intended purpose.

dbwiddis:
@saratvemulapalli assigning this to you as you did #164, are continuing to iterate on dynamic actions, and may have better ideas than I did above for implementation.

saratvemulapalli:
@dbwiddis Trying to catchup from PR: https://github.com/opensearch-project/opensearch-sdk-java/pull/466

Looks like we do have `SDKActionModule`, which registers actions in the SDK. 
Couple of questions (you can point me back to the PR, I just skimmed through):
- How are these actions registered with `ActionModule` on OpenSearch(core)?
- Having an `ActionModule` within the SDK will bring in state inconsistency problems, I am worried that the extension thinks it has registered the action but the namespace might have conflicted in OpenSearch.

dbwiddis:
> @dbwiddis Trying to catchup from PR: #466
> 
> Looks like we do have `SDKActionModule`, which registers actions in the SDK. Couple of questions (you can point me back to the PR, I just skimmed through):
> 
> * How are these actions registered with `ActionModule` on OpenSearch(core)?

The same way they already were (or weren't, due to bug #484) with your previous `getActions()` method.

Background:

[...]

PR #477, opened 2023-02-21
Pass ExtensionsRunner instance to createComponents

dbwiddis:
### Description

Guice has not finished initializing the injector when calling `createComponents` as the return objects need to be injected.  Components which require objects from the `ExtensionsRunner` for instantiation need a copy of them available.

While the constructor could be filled with a long list of these parameters (as is done in plugins) this creates a hard API dependency on the parameters, and changing the number of parameters would break compatibility.

By passing the ExtensionsRunner, additional objects can be made available via getters without changing the API.

### Issues Resolved

Improves on #466

### DIFF
--- a/src/main/java/org/opensearch/sdk/Extension.java
+++ b/src/main/java/org/opensearch/sdk/Extension.java
@@ -53,9 +53,10 @@ public interface Extension {
     /**
      * Returns components added by this extension.
      *
+     * @param runner the ExtensionsRunner instance. Use getters from this object as required to instantiate components to return.
      * @return A collection of objects which will be bound to themselves for dependency injection.
      */
-    default Collection<Object> createComponents() {
+    default Collection<Object> createComponents(ExtensionsRunner runner) {
         return Collections.emptyList();
     }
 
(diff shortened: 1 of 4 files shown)


ryanbogan (approved):

owaiskazi19 (approved):
We might need to revisit this logic while working on https://github.com/opensearch-project/opensearch-sdk-java/issues/467. Approved for now.

dbwiddis:
> We might need to revisit this logic while working on #467. 

Agreed.   Here's the overall logical constraints (which probably should find its way into code comments):

Order matters:
1. Initialize concrete classes needed by create components. If these are things that may change during operation (e.g. NamedXContents) use a concrete wrapper class with the ability to update its internals.   
2. Inject classes (any order, Guice figures out the dependency map later)
   - Inject the concrete classes.  Since they aren't yet available to create components, make sure they have getters and then just pass the runner to create components.  Injecting them using the getters is a subtle reminder of this. :)
   - Inject the objects returned from create components.
   - Inject any other extension points that require injection (in this case getActions).
3. After injection, call other extension points (e.g., rest handlers). This permits these other extension points to have access to the injected classes.

PR #490, opened 2023-02-27
Added localNode() to SDKCluster service to retrieve the extensionNode from the extensionsRunner

joshpalis:
Gives SDKClusterService access to the localNode(), ie the extension node set during extension initialization. This is needed for ADTaskManager

### DIFF
--- a/src/main/java/org/opensearch/sdk/SDKClusterService.java
+++ b/src/main/java/org/opensearch/sdk/SDKClusterService.java
@@ -14,6 +14,7 @@ import java.util.function.Consumer;
 
 import org.opensearch.cluster.ClusterState;
 import org.opensearch.common.settings.Setting;
+import org.opensearch.extensions.DiscoveryExtensionNode;
 
 /**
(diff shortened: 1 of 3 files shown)


owaiskazi19 (dismissed):

owaiskazi19 (approved):

src/main/java/org/opensearch/sdk/SDKClusterService.java
```diff
@@ -42,6 +43,15 @@ public ClusterState state() {
         return extensionsRunner.sendClusterStateRequest(extensionsRunner.getExtensionTransportService());
     }
 
+    /**
+     *  Returns the local extension node
+     *
+     * @return the local extension node
+     */
+    public DiscoveryExtensionNode localNode() {
```
dbwiddis, line 51: We should come up with a consistent policy on when to use record-style getters like this vs. `getLocalNode()` style. 

In general, record-style is appropriate for immutable return values.

dbwiddis (approved):
LGTM. See comment for a possible new DISCUSS issue.

PR #510, opened 2023-03-02
Delay sending Settings Update Consumers until initialization

### LINKED ISSUE #507: [BUG] Adding a Settings Update Consumer before connection with OpenSearch fails with NPE
### What is the bug?

> java.lang.NullPointerException: Cannot invoke "org.opensearch.transport.TransportService.sendRequest(org.opensearch.cluster.node.DiscoveryNode, String, org.opensearch.transport.TransportRequest, org.opensearch.transport.TransportResponseHandler)" because "transportService" is null
>         at org.opensearch.sdk.ExtensionsRunner.sendAddSettingsUpdateConsumerRequest(ExtensionsRunner.java:566)
>         at org.opensearch.sdk.SDKClusterService$SDKClusterSettings.addSettingsUpdateConsumer(SDKClusterService.java:81)
>         at org.opensearch.sdk.SDKClusterService$SDKClusterSettings.addSettingsUpdateConsumer(SDKClusterService.java:72)
>         at org.opensearch.ad.MemoryTracker.<init>(MemoryTracker.java:81)
> 

1. Get the latest `feature/extensions` branch on AD extension following merge of #825 
2. `./gradlew run`

AD Extension starts up.

The root cause of the problem is that on initialization, an extension does not have a valid `transportService` object because it has not yet been initialized.  However, settings update consumers in injected objects all attempt to register their settings update consumers on injection (eager singleton).

### Proposed fix

Maintain a list (or map) of pending update requests in the `SDKClusterSettings` object.  Periodically poll for non-null transport service and send the settings when possible to do so.  This gains the benefit of keeping the existing plugin API of single-line injection while building a map to send more efficiently at a later time.  This is a similar approach to storing a local copy of the Rest Handlers on initialization and sending them to OpenSearch during the init sequence.

dbwiddis:
### Description

Before `ExtensionsRunner` initialization, there is no transport service and settings have not yet been sent to OpenSearch to update.  This change adds a pending updates map to the `SDKClusterSettings` object to keep track of settings that haven't yet been sent.

### Issues Resolved

Fixes #507 

### DIFF
--- a/src/main/java/org/opensearch/sdk/SDKClusterService.java
+++ b/src/main/java/org/opensearch/sdk/SDKClusterService.java
@@ -10,6 +10,7 @@
 package org.opensearch.sdk;
 
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 
 import org.opensearch.cluster.ClusterState;
@@ -40,7 +41,10 @@ public class SDKClusterService {
      * @return the cluster state of OpenSearch
      */
     public ClusterState state() {
-        return extensionsRunner.sendClusterStateRequest(extensionsRunner.getExtensionTransportService());
+        if (extensionsRunner.isInitialized()) {
+            return extensionsRunner.sendClusterStateRequest(extensionsRunner.getExtensionTransportService());
+        }
+        throw new IllegalStateException("The Extensions Runner has not been initialized.");
     }
 
     /**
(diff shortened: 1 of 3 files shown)


owaiskazi19 (approved):
LGTM! Also, I think we should have a common design to check if the extension has been initialized before sending any request to OpenSearch or other way.

joshpalis (approved):

PR #520, opened 2023-03-03
Properly initialize SDKClient actions map

### LINKED ISSUE #519: [BUG] SDKClient execute() fails with NPE
### What is the bug?

SDKClient use of `execute()` implemented in #491 fails with NPE (`actions` is null) here:
```java
TransportAction<Request, Response> transportAction = actions.get(action);
```
The root cause is that we incorrectly attempt to access an injected `actions` map:
```java
// Used by client.execute
@SuppressWarnings("rawtypes")
@Inject
private Map<ActionType, TransportAction> actions;
```
However, the `SDKClient` was instantiated before injection so the map wasn't injected.

1. Run an extension and use the `execute()` method from `SDKClient` or `SDKRestClient`.
2. Observe REST error response that "actions" is null.

`actions.get(action)` returns an instance of a class implementing `TransportAction`.

The implementation was ported from `NodeClient` with this field populated by the `initialize()` method (and not by injection):
```java
private Map<ActionType, TransportAction> actions;
```
This code from `Node.java` was not included in `ExtensionsRunner` shortly after injection completes, so porting it should fix this bug:
```java
client.initialize(injector.getInstance(new Key<Map<ActionType, TransportAction>>() {}));
```

dbwiddis:
### Description

Adds an `initialize()` method to the `SDKClient` to allow its `actions` map to be populated after Guice injection, even though the class was instantiated before.  This is the same pattern used with `NodeClient` that we are emulating.

In addition, maintains a field for the Guice `Injector` in the `ExtensionsRunner`.  It was getting garbage collected and losing its usefulness, particularly in Rest Handlers.

### Issues Resolved

Fixes #519

### DIFF
--- a/src/main/java/org/opensearch/sdk/SDKClient.java
+++ b/src/main/java/org/opensearch/sdk/SDKClient.java
@@ -18,7 +18,6 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
-import com.google.inject.Inject;
 import com.fasterxml.jackson.datatype.guava.GuavaModule;
 import org.apache.hc.core5.function.Factory;
 import org.apache.hc.core5.http.HttpHost;
@@ -85,15 +84,19 @@ public class SDKClient implements Closeable {
     private RestClient restClient;
     private RestHighLevelClient sdkRestClient;
 
-    // Used by client.execute
+    // Used by client.execute, populated by initialize method
     @SuppressWarnings("rawtypes")
-    @Inject
     private Map<ActionType, TransportAction> actions;
 
     /**
-     * Instantiate this client.
+     * Initialize this client.
+     *
+     * @param actions The injected map of ActionType instances to TransportAction.
      */
-    public SDKClient() {}
+    @SuppressWarnings("rawtypes")
+    public void initialize(Map<ActionType, TransportAction> actions) {
+        this.actions = actions;
+    }
 
     /**
      * Create and configure a RestClientBuilder
--- a/src/main/java/org/opensearch/sdk/SDKClient.java
+++ b/src/main/java/org/opensearch/sdk/SDKClient.java
@@ -254,7 +254,7 @@ public class SDKClient implements Closeable {
         ActionListener<Response> listener
     ) {
         if (actions == null) {
-            throw new IllegalStateException("SDKClient was not initialized because the Extension does not implement ActionExtession.");
+            throw new IllegalStateException("SDKClient was not initialized because the Extension does not implement ActionExtension.");
         }
         @SuppressWarnings("unchecked")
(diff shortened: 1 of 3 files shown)


owaiskazi19 (dismissed):

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -250,6 +253,9 @@ public final <Request extends ActionRequest, Response extends ActionResponse> vo
         Request request,
         ActionListener<Response> listener
     ) {
+        if (actions == null) {
```
owaiskazi19, line 256: Do we need this check? Just for my understanding, what if there's an extension which doesn't have `getActions` extension point?

dbwiddis: > Do we need this check? Just for my understanding, what if there's an extension which doesn't have `getActions` extension point?

Then they shouldn't call `client.execute()` and this gives them a more useful error message than the NPE.

dbwiddis: Another possibility is just to initialize with an empty map so you don't get the NPE but you'd still get the "unregistered action" error.

owaiskazi19: Makes sense!

owaiskazi19 (approved):

src/main/java/org/opensearch/sdk/SDKClient.java
joshpalis: +1, There are several use cases in which Plugins/Extensions need to trigger client.execute to invoke actions provided by OpenSearch, such as `UpdateByQueryAction`. We should still allow this without requiring an extension to provide its own actions.

Also minor nit (if we decide to keep this check) change `ActionExtession` to `ActionExtension`

src/test/java/org/opensearch/sdk/sample/helloworld/TestHelloWorldExtension.java
```diff
@@ -227,4 +237,29 @@ public void onFailure(Exception e) {
...
+                }
+
+                @Override
+                public void onFailure(Exception e) {
+                    responseFuture.completeExceptionally(e);
+                }
+            })
+        );
+
+        assertEquals("SDKClient was not initialized because the Extension does not implement ActionExtession.", ex.getMessage());
```
owaiskazi19, line 263:
```suggestion
        assertEquals("SDKClient was not initialized because the Extension does not implement ActionExtension.", ex.getMessage());
```

dbwiddis: To my credit I had a test case making sure the typo was there too.

joshpalis (approved):

owaiskazi19 (approved):

src/main/java/org/opensearch/sdk/SDKClient.java
dbwiddis: Good point about the OpenSearch actions.  They probably have their own injected constructor values on the OpenSearch side so might need to actually be remotely executed on OS node client. 

Will look into that as part of #467 work.

PR #530, opened 2023-03-06
Adding ExtensionsRunner parameter to getExtensionRestHandlers

joshpalis:
### Description
Provides the ExtensionRunner to getExtensionsRestHandlers. This allows us to pass in the same instance of the extension runner to all registered rest actions.

Companion PR : https://github.com/opensearch-project/anomaly-detection/pull/835

### DIFF
--- a/src/main/java/org/opensearch/sdk/ActionExtension.java
+++ b/src/main/java/org/opensearch/sdk/ActionExtension.java
@@ -69,7 +69,7 @@ public interface ActionExtension {
      *
      * @return a list of REST handlers (REST actions) this extension handles.
      */
-    default List<ExtensionRestHandler> getExtensionRestHandlers() {
+    default List<ExtensionRestHandler> getExtensionRestHandlers(ExtensionsRunner extensionsRunner) {
         return Collections.emptyList();
     }
 
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -199,7 +199,9 @@ public class ExtensionsRunner {
             // initialize SDKClient action map
             initializeSdkClient();
             // store REST handlers in the registry
-            for (ExtensionRestHandler extensionRestHandler : ((ActionExtension) extension).getExtensionRestHandlers()) {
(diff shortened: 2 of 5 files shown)


src/main/java/org/opensearch/sdk/ActionExtension.java
```diff
@@ -69,7 +69,7 @@ default List<ActionFilter> getActionFilters() {
      *
      * @return a list of REST handlers (REST actions) this extension handles.
      */
-    default List<ExtensionRestHandler> getExtensionRestHandlers() {
+    default List<ExtensionRestHandler> getExtensionRestHandlers(ExtensionsRunner extensionsRunner) {
```
dbwiddis, line 72: I really don't like the idea of changing the method signature of an existing extension point.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -199,7 +199,9 @@ protected ExtensionsRunner(Extension extension) throws IOException {
             // initialize SDKClient action map
             initializeSdkClient();
             // store REST handlers in the registry
-            for (ExtensionRestHandler extensionRestHandler : ((ActionExtension) extension).getExtensionRestHandlers()) {
+            for (ExtensionRestHandler extensionRestHandler : ((ActionExtension) extension).getExtensionRestHandlers(
+                injector.getInstance(ExtensionsRunner.class)
```
dbwiddis, line 203: What benefit does this give us over simply passing `this` here, if we do keep this?

dbwiddis (commented):
I'm not clear why we need this or what is actually broken.
1. There is exactly one `ExtensionsRunner` instance and it is `this` in the context of executing code in the runner.
2. We shouldn't change the signature of existing extension points.
3. We already pass the instance of the `ExtensionsRunner` into the `Extension` interface `createComponents()` method.  If the extension needs to persist a copy of that instance, it could simply set a local field to that value as part of implementing `createComponents()`.
4. As demonstrated in the `BaseExtension` class, as an alternative to the method in 3, we can simply inject the value in an extension (since the extension is also injected, this appears to work just fine).

Issue #168, opened 2022-10-06
[FEATURE] OpenSearch java doc tags


saratvemulapalli:
Adding up to the conventions, opening this issue to discuss java doc tags.
We use:
- `opensearch.internal`
- `opensearch.api`
- etc 
Lets chat and clean up the code base for `feature/extensions` and SDK with clear communication about which tag to be used at what times.

dbwiddis:
My 2c:

> clean up the code base for `feature/extensions`

This will be eventually merged to main, so we should simply follow the OpenSearch policy there.   Yes, we should clean things up and use this issue to ask any questions.

> and SDK 

There's no requirement that we use exactly the same convention and I'd rather not.  I'm a big fan of package-based API.  It's very simple:
 - If the class is in `org.opensearch.sdk` package, it's API.  Otherwise it's internal.

If we want another api package, we can add one.

This aligns with the vision of JPMS, where there is package-level control over what other modules can use.  Mixing API classes in packages with non-API stuff forces us to open the whole non-API package in a module.

dbwiddis:
Hey @saratvemulapalli what do you think we need to do to complete this?   My thoughts:
1. inventory all the classes
2. categorize them as "definitely API" or "definitely internal" or leave blank
3. engage in a spirited discussion on this issue on the blank ones
4. assign someone this issue to add the annotation (OpenSearch feature/extensions side) or put in the right package (SDK side).

PR #534, opened 2023-03-07
Created ScriptExtension interface #318

### LINKED ISSUE #318: [FEATURE] Create ScriptExtension interface

The `getScriptEngine()` and `getContexts()` extension points communicate an extension's script capabilities to OpenSearch.

Add a new interface `ScriptExtension` with the above two methods, with default implementations matching `ScriptPlugin.java`.

petiveriaalliacea:
### Description
Added a new interface ScriptExtension with the above two methods, with default implementations matching ScriptPlugin.java.

### Issues Resolved
#318 

### DIFF
--- a/src/main/java/org/opensearch/sdk/ScriptExtension.java
+++ b/src/main/java/org/opensearch/sdk/ScriptExtension.java
@@ -20,8 +20,6 @@ import java.util.Map;
 
 /**
  * An additional extension point for {@link Extension}s that extends OpenSearch's scripting functionality.
- *
- * @opensearch.api
  */
 public interface ScriptExtension {
 
--- a/src/main/java/org/opensearch/sdk/ScriptExtension.java
+++ b/src/main/java/org/opensearch/sdk/ScriptExtension.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.sdk;
+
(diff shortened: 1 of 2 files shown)


dbwiddis (changes requested):
Thanks for your contribution!

Two minor nits (Spotless check is failing) and this should be GTG.

src/main/java/org/opensearch/sdk/ScriptExtension.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * An additional extension point for {@link Extension}s that extends OpenSearch's scripting functionality.
+ *
+ * @opensearch.api
+ */
+public interface ScriptExtension {
```
saratvemulapalli, line 26: @dbwiddis as we move over more extension points, we should probably group all these interfaces in one directory.

saratvemulapalli (approved):
Thank you for the contribution!!
LGTM other than comments from @dbwiddis

src/main/java/org/opensearch/sdk/ScriptExtension.java
dbwiddis: @saratvemulapalli already would have done it if I could [think of a good package name](https://martinfowler.com/bliki/TwoHardThings.html).

dbwiddis (changes requested):
Gradle check is failing:

> /home/runner/work/opensearch-sdk-java/opensearch-sdk-java/src/main/java/org/opensearch/sdk/ScriptExtension.java:24: error: unknown tag: opensearch.api

We don't (yet) use that annotation on the SDK, see discussion at #168.

So for now just remove this.

dbwiddis (approved):

joshpalis (approved):

saratvemulapalli (approved):

PR #546, opened 2023-03-09
Created the MapperExtension interface and added the MapperPlugin content

Kuanysh-kst:
### Description

### Issues Resolved


dbwiddis (changes requested):

src/test/java/org/opensearch/sdk/TestExtensionMapper.java
```diff
@@ -0,0 +1,66 @@
...
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.opensearch.index.mapper.Mapper;
+import org.opensearch.index.mapper.MetadataFieldMapper;
+
+import java.util.Map;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+public class TestExtensionMapper {
```
dbwiddis, line 21: Nice test class, very thorough!   Can we either name it `TestMapperExtension` since that's the class we're testing, or just move some of these tests into the `TestExtensionInterfaces` class? I'm not sure we need all of them.

src/test/java/org/opensearch/sdk/TestExtensionMapper.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Test
+    public void testGetFieldFilter() {
+        MapperExtension extension = new MapperExtension() {
+        };
+        Function<String, Predicate<String>> fieldFilter = extension.getFieldFilter();
+        String indexName = "myIndex";
+        Predicate<String> predicate = fieldFilter.apply(indexName);
+        Assertions.assertNotNull(predicate);
```
dbwiddis, line 48: I think the above two lines might be better asserting the class instanceof for the return value of `apply()`.

dbwiddis (commented):
Looks great, one request to change the test class name.

Not sure all the null checks are needed but they don't hurt, either :)

dbwiddis (approved):
LGTM.   Approving, but noting that the other tests group all the assertions for an extension in the same method. Not sure either style is preferable so we can merge this as is, and decide later which way to go.

ryanbogan (approved):

PR #527, opened 2023-03-06
Created ScriptExtension interface #318

petiveriaalliacea:
### Description
Added a new interface ScriptExtension with the above two methods, with default implementations matching ScriptPlugin.java.

### Issues Resolved
#318 
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #553, opened 2023-03-13
Added getNamedWriteableRegistry() API for extensions

### LINKED ISSUE #291: [Feature/extensions] Add getNamedWriteableRegistry() API for extensions
The old model for the getNamedWriteableRegistry() was recently removed, due to concerns here: https://github.com/opensearch-project/OpenSearch/pull/5518#discussion_r1049668323.  A new model must be created, which should be similar to the implementation of getNamedXContent().

petiveriaalliacea:
### Description
Added getNamedWriteableRegistry() API for extensions

### Issues Resolved
#291 


src/main/java/org/opensearch/sdk/SDKNamedWriteableRegistry.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Combines Extension NamedWriteable with core OpenSearch NamedWriteable
+ */
+public class SDKNamedWriteableRegistry {
+    private NamedWriteableRegistry namedXWriteableRegistry;
```
dbwiddis, line 31: Remove the `X`

src/main/java/org/opensearch/sdk/SDKNamedWriteableRegistry.java
```diff
@@ -0,0 +1,63 @@
...
+     * Creates and populates a NamedWriteableRegistry with the NamedWriteableRegistry entries for this extension and locally defined content.
+     *
+     * @param runner The ExtensionsRunner instance.
+     */
+    public SDKNamedWriteableRegistry(ExtensionsRunner runner) {
+        this.namedXWriteableRegistry = createRegistry(runner.getEnvironmentSettings(), runner.getCustomNamedWriteable());
+    }
+
+    private NamedWriteableRegistry createRegistry(Settings settings, List<Entry> extensionNamedWriteable) {
+        Stream<Entry> extensionContent = extensionNamedWriteable == null ? Stream.empty() : extensionNamedWriteable.stream();
```
dbwiddis, line 43: Add an `s` to writeable

src/main/java/org/opensearch/sdk/SDKNamedWriteableRegistry.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Gets the NamedWriteableRegistry.
+     *
+     * @return The NamedWriteableRegistry. Includes both extension-defined Writeable and core OpenSearch Writeable.
+     */
+    public NamedWriteableRegistry getRegistry() {
+        return this.namedXWriteableRegistry;
+    }
+}
```
dbwiddis, line 63: You need to add methods (See `SDKNamedXContentRegistry` for `updateNamedWriteableRegistry(ExtensionsRunner runner)` and a setter for the private field.

src/main/java/org/opensearch/sdk/ScriptExtension.java
```diff
@@ -0,0 +1,41 @@
+/*
```
dbwiddis, line 1: This file looks associated with a different PR....

src/test/java/org/opensearch/sdk/TestExtensionInterfaces.java
```diff
@@ -48,6 +48,15 @@ void testActionExtension() {
         assertTrue(actionExtension.mappingRequestValidators().isEmpty());
     }
 
+    @Test
```
dbwiddis, line 51: Also from a different PR.  :)

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -167,8 +174,12 @@ protected ExtensionsRunner(Extension extension) throws IOException {
         this.customSettings = extension.getSettings();
         // save custom namedXContent
         this.customNamedXContent = extension.getNamedXContent();
+        // save custom namedWriteable
+        this.customNamedWriteable = extension.getNamedWriteable();
         // initialize NamedXContent Registry. Must happen after getting extension namedXContent
         this.sdkNamedXContentRegistry = new SDKNamedXContentRegistry(this);
+        // initialize NamedWriteable Registry. Must happen after getting extension namedWriteable
+        this.sdkNamedWriteableRegistry = new SDKNamedWriteableRegistry(this);
```
dbwiddis, line 182: Let's add this variable (via its getter) to the injected components in Guice (below line 196).

dbwiddis (changes requested):
Thanks for this contribution!   See comments below.  Also you need a test class (should be very similar to `TestSDKNamedXContentRegistry`)

src/main/java/org/opensearch/sdk/NettyTransport.java
```diff
@@ -60,20 +52,9 @@ public NettyTransport(ExtensionsRunner extensionsRunner) {
      * @param threadPool  A thread pool to use.
      * @return The configured Netty4Transport object.
      */
-    public Netty4Transport getNetty4Transport(Settings settings, ThreadPool threadPool) {
+    public Netty4Transport getNetty4Transport(Settings settings, ThreadPool threadPool, NamedWriteableRegistry namedWriteableRegistry) {
```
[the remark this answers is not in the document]
petiveriaalliacea: Should I remove this parameter?

src/test/java/org/opensearch/sdk/TestExtensionInterfaces.java
```diff
@@ -1,92 +0,0 @@
-/*
```
dbwiddis, line 1: This file should not have been deleted!

dbwiddis (changes requested):

src/test/java/org/opensearch/sdk/TestExtensionInterfaces.java
petiveriaalliacea: May I just create a new pull request? I didn't know how to remove those files :(

dbwiddis: You can, but let's try something else first:
1. use `git reset --hard HEAD~2` to back up your repository to where it was before those last two deletions
2. change to the main branch, pull from upstream
3. go back to your develop branch and type `git rebase main`
4. force-push the results

petiveriaalliacea: @dbwiddis, I've tried to do what you said. Does that work?

dbwiddis: Well those deleted files are back, which is good.   They are part of a merged PR #527 so I'm not sure the rebase worked as you wanted.  It's fine for now, it's just noise on the PR.  But a merge or rebase to main should fix it.

I'm not clear about your question on the XContentParser.  Is the code above taken from the test for the XContent parsing and you're trying to make the equivalent for the named writeables?

dbwiddis:
Hey @petiveriaalliacea let's take another shot at trying to sync things up.  I think when you initially tried to rebase, you had merge conflicts and it didn't go through and resolve them.  There are a few ways of doing that.

First, there is a GitHub interface: look below the status checks and you will see an alert "This branch has conflicts that must be resolved".  It shows 3 files that you need to fix up.  If you click on the "Resolve conflicts" button you'll see each of the files.  Look for the conflicts, they will look like this:
```
<<<<<<< develop
public abstract class BaseExtension implements Extension, ActionExtension, ScriptExtension {
=======
public abstract class BaseExtension implements Extension {
>>>>>>> main
```
The part above the `=======` is from your branch, the part below is what's currently on main branch.  Pick which one you want and delete the other one, or sometimes pick from both of them.  So the above 5 lines would reduce to simply
```java
public abstract class BaseExtension implements Extension {
```
When you're done with that file, click "Mark as resolved" and move on to the next file.

For ExtensionsRunner there are two conflicting sections, you'll want to keep your `SDKNamedWriteableRegistry` addition from your branch (top line under `<<<<<<< develop` but then keep the rest of the code from the main branch's portion of those conflicts.

For ScriptExtension just keep whatever's on `main`.

After marking all 3 files resolved you can create a commit to github fixing those, and then pull it down to your local copy of the branch.

There are ways to do this locally as well:
 - make sure your main branch is up to date with SDK (`git pull upstream main` `git reset --hard`)
 - merge or rebase `git rebase main` or `git merge main` 
   - that will note the same conflicts above. You can edit them using the editor of your choice or your IDE.  Some IDEs have merge tools you can use.   
   - once done editing, use "git add" (or in your IDE stage them) and then continue the rebase/merge

If all else fails, rather than recreating the PR you can just clean up the existing branch and force push it.  One strategy I do is:
 - create a temp branch matching my "broken" branch (`git checkout -b temp`)
 - delete my "broken" branch
 - checkout `main` branch, make sure it's up to date
 - checkout -b to the same name as my just-deleted broken branch
 - cherry-pick commits from the temp branch into the new branch
 - force push the results, the PR will be updated.

src/main/java/org/opensearch/sdk/NettyTransport.java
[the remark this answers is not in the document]
dbwiddis: Yes, you should leave this line unchanged.

src/test/java/org/opensearch/sdk/TestExtensionInterfaces.java
petiveriaalliacea: @dbwiddis, Yes, that's what I am trying to do.

petiveriaalliacea: The thing is that ToXContentObject, XContentParser, JsonXContent, XContentBuilder, etc. are taken from the library. Should I create equivalent for all of these name writeables(Example: WritableParser)? If yes, where should I put it? What folder?

dbwiddis: Not completely sure as I haven't dug into this but some quick observations/advice:
 - none of the XContent things are really relevant here.  writeable is a totally different implementation
 - there is a `NamedWriteableRegistryTests` class on OpenSearch that is probably useful here, but it really just shows a regisration.  It may be much simpler in your case as well.
 - bottom line: if you have a test that invokes all the new methods you are adding and produces expected results you should be good
 - you can run `gradle diffCoverage` to make sure you've covered the new code

src/test/java/org/opensearch/sdk/TestSDKNamedWriteableRegistry.java
```diff
@@ -0,0 +1,103 @@
...
+        );
+
+        private final String name;
+
+        public Example(String name) {
+            this.name = name;
+        }
+
+
+        @Override
```
dbwiddis, line 36: we probably don't need equals and hashcode for a test class, but it does no harm ...

src/test/java/org/opensearch/sdk/TestSDKNamedWriteableRegistry.java
```diff
@@ -0,0 +1,103 @@
...
+    public void testDefaultNamedWriteableRegistry() throws IOException {
+        NamedWriteableRegistry registry = runner.sdkNamedWriteableRegistry.getRegistry();
+
+        XContentParseException ex = assertThrows(
+                XContentParseException.class,
+                () -> registry.getReader(TestSDKNamedWriteableRegistry.Example.class, TestSDKNamedWriteableRegistry.Example.NAME)
+        );
+        assertEquals("unknown named object category [" + TestSDKNamedWriteableRegistry.Example.class.getName() + "]", ex.getMessage());
+    }
+
```
dbwiddis, line 102: need one more test method for the `testCustom...` version where you call `runner.update...()` and the test that fails above should pass in this other method

dbwiddis (changes requested):
Thanks for working on this!  Almost there, just need one more test method to test the update method.

dbwiddis (approved):
LGTM, still one unresolved suggestion to add the registry to the Guice bindings.  That can be done later.

dbwiddis (changes requested):
Actually... looks like your new tests are failing. Can you see what's missing?

petiveriaalliacea:
@dbwiddis I've fixed testDefaultNamedWriteableRegistry. I am not sure how to fix the other one (testCustomNamedWriteableRegistry). Test failes of NullPointerException, because I am passing null in line 33. But I don't know what should I pass instead. Could you please help me?

dbwiddis:
> @dbwiddis I am not sure how to fix the other one (testCustomNamedWriteableRegistry). Test failes of NullPointerException, because I am passing null in line 33. But I don't know what should I pass instead. Could you please help me?

Take a look at `NamedWriteableRegistry`.  You need to pass a `Writeable.Reader` there.
```java
        /** Creates a new entry which can be stored by the registry. */
        public <T extends NamedWriteable> Entry(Class<T> categoryClass, String name, Writeable.Reader<? extends T> reader) {
            this.categoryClass = Objects.requireNonNull(categoryClass);
            this.name = Objects.requireNonNull(name);
            this.reader = Objects.requireNonNull(reader);
        }
```
So jump over to `NamedWriteaableRegistryTests` and see what they do:
```java
    public void testBasic() throws IOException {
        NamedWriteableRegistry.Entry entry = new NamedWriteableRegistry.Entry(NamedWriteable.class, "test", DummyNamedWriteable::new);
        NamedWriteableRegistry registry = new NamedWriteableRegistry(Collections.singletonList(entry));
        Writeable.Reader<? extends NamedWriteable> reader = registry.getReader(NamedWriteable.class, "test");
        assertNotNull(reader.read(null));
    }
```
So we see `DummyNamedWriteable::new` there.  Basically this is the constructor taking StreamInput:
```java
    private static class DummyNamedWriteable implements NamedWriteable {
        DummyNamedWriteable(StreamInput in) {}

        @Override
        public String getWriteableName() {
            return "test";
        }

        @Override
        public void writeTo(StreamOutput out) throws IOException {}
    }
```
So my recommendation is to copy that dummy class as is and use `DummyNamedWriteable::new`.

dbwiddis (approved):
LGTM! Thanks for this contribution!

ryanbogan (approved):
Thanks for the contribution!

Issue #557, opened 2023-03-13
[FEATURE] Integrate Async Java Client


owaiskazi19:
### Is your feature request related to a problem?
To invoke the Actions residing in OpenSearch. Extensions will use Java Client to call the action using REST requests available. Currently, we support synchronous Java Client. As AD plugin uses asynchronous request calls, create an async Java Client for the same.

### What solution would you like?
opensearch-java supports OpenSearchAsyncClient. Create a SDKClient wrapper for the same.

### What alternatives have you considered?
_A clear and concise description of any alternative solutions or features you've considered._

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

dbwiddis:
Duplicate of #296 :-)

owaiskazi19:
> Duplicate of #296 :-)

Ah! Didn't realize. Will close this one

Issue #296, opened 2022-12-23
[FEATURE] Create Async SDKClient


owaiskazi19:
### Is your feature request related to a problem?
Currently, SDKClient which sits on top of opensearch-java client only supports synchronous request calls. This issue talks about adding Async functionality for SDKClient.

### What solution would you like?
_A clear and concise description of what you want to happen._

### What alternatives have you considered?
_A clear and concise description of any alternative solutions or features you've considered._

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

dbwiddis:
I think we should give users options:
 - Async client (preferred, recommended)
 - Sync client (an option for simplicity)
 - High Level Rest Client (deprecated, provided for compatibility with existing plugins, not recommended)

Issue #383, opened 2023-01-29
Migrate Start Detector job to SDK using SDKRestClient


dbwiddis:
### Is your feature request related to a problem?
 
Part of https://github.com/opensearch-project/OpenSearch/issues/5224
 
### What solution would you like?
 
Migrate the Start Detector job Rest API to Extensions SDK using SDKRestClient.
 
Use the implementation of #353 as a reference.  Add new issues for any steps you skip.
 
Implement the following API endpoints:
 - [x] `POST _plugins/_anomaly_detection/detectors/<detectorId>/_start`

joshpalis:
### Remaining Tasks :
Replace the following instances in which `client.execute()` is used to trigger actions registered within OpenSearch with corresponding methods made available by the Java Client ([OpenSearchAsyncClient](https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/main/java/org/opensearch/client/opensearch/OpenSearchAsyncClient.java)).

- [ ] [AnomalyDetectionIndicies.updateJobIndexSettingIfNecessary()](https://github.com/joshpalis/anomaly-detection/blob/1e097c9d26aed1ab806f2643e1fe4465cc40fd88/src/main/java/org/opensearch/ad/indices/AnomalyDetectionIndices.java#L1086)
  - Action : GetSettingsAction
  - Request: GetSettingsRequest
  - Response : GetSettingsResponse
  - Note : Once this client.execute is replaced, uncomment `anomalyDetectionIndices.update()` in the following two places
     - [IndexAnomalyDetectorTransportAction](https://github.com/joshpalis/anomaly-detection/blob/1e097c9d26aed1ab806f2643e1fe4465cc40fd88/src/main/java/org/opensearch/ad/transport/IndexAnomalyDetectorTransportAction.java#L150)
     - [ADTaskManager](https://github.com/joshpalis/anomaly-detection/blob/1e097c9d26aed1ab806f2643e1fe4465cc40fd88/src/main/java/org/opensearch/ad/task/ADTaskManager.java#L300)
- [ ] [ADTaskManager.updateLatestFlagOfOldTasksAndCreateNewTask()](https://github.com/joshpalis/anomaly-detection/blob/1e097c9d26aed1ab806f2643e1fe4465cc40fd88/src/main/java/org/opensearch/ad/task/ADTaskManager.java#L1523)
  - Action : UpdateByQueryAction
  - Request: UpdateByQueryRequest
[...]

PR #558, opened 2023-03-13
Integrated Async Java Client

owaiskazi19:
### Description
To invoke the Actions residing in OpenSearch. Extensions will use Java Client to call the action using REST requests available. Currently, we support synchronous Java Client. As AD plugin uses asynchronous request calls, create an async Java Client for the same.

This is needed for the migration of [Start Detector](https://github.com/opensearch-project/opensearch-sdk-java/issues/383).

### Issues Resolved
Fixes https://github.com/opensearch-project/opensearch-sdk-java/issues/557
Fixes https://github.com/opensearch-project/opensearch-sdk-java/issues/296

### DIFF
--- a/src/main/java/org/opensearch/sdk/SDKClient.java
+++ b/src/main/java/org/opensearch/sdk/SDKClient.java
@@ -166,7 +166,7 @@ public class SDKClient implements Closeable {
      *
      * @param settings The Extension settings
      * @return The SDKClient implementation of OpenSearchClient. The user is responsible for calling
-     *         {@link #doCloseJavaClient()} when finished with the client
+     *         {@link #doCloseJavaClients()} when finished with the client
      */
     public OpenSearchClient initializeJavaClient(ExtensionSettings settings) {
         return initializeJavaClient(settings.getOpensearchAddress(), Integer.parseInt(settings.getOpensearchPort()));
@@ -178,7 +178,7 @@ public class SDKClient implements Closeable {
      * @param hostAddress The address of OpenSearch cluster, client can connect to
      * @param port The port of OpenSearch cluster
      * @return The SDKClient implementation of OpenSearchClient. The user is responsible for calling
-     *         {@link #doCloseJavaClient()} when finished with the client
+     *         {@link #doCloseJavaClients()} when finished with the client
      */
     public OpenSearchClient initializeJavaClient(String hostAddress, int port) {
         OpenSearchTransport transport = initializeTransport(hostAddress, port);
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -147,6 +172,17 @@ public OpenSearchClient initializeJavaClient(ExtensionSettings settings) {
         return initializeJavaClient(settings.getOpensearchAddress(), Integer.parseInt(settings.getOpensearchPort()));
     }
 
+    /**
```
dbwiddis, line 175: (Style) Please pair the overloaded client initialize methods together. See https://google.github.io/styleguide/javaguide.html#s3.4.2-ordering-class-contents

Specifically, move this method below `initializeJavaClient()`.

src/test/java/org/opensearch/sdk/TestSDKClient.java
```diff
@@ -63,6 +66,15 @@ public void testCreateJavaClient() throws Exception {
         sdkClient.doCloseJavaClient();
     }
 
+    @Test
+    public void testCreateJavaAsyncClient() throws Exception {
+        OpenSearchAsyncClient javaAsyncClient = sdkClient.initializeJavaAsyncClient(settings);
+        assertInstanceOf(OpenSearchIndicesAsyncClient.class, javaAsyncClient.indices());
+        assertInstanceOf(OpenSearchClusterAsyncClient.class, javaAsyncClient.cluster());
+
+        sdkClient.doCloseJavaClient();
```
dbwiddis, line 75: Shouldn't this be `doCloseJavaAsyncClient()` (which doesn't yet exist but should, along with adding to `close()`?)

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -147,6 +172,17 @@ public OpenSearchClient initializeJavaClient(ExtensionSettings settings) {
...
+     * Initializes an OpenAsyncSearchClient using OpenSearch JavaClient
+     *
+     * @param settings The Extension settings
+     * @return The SDKClient implementation of OpenSearchAsyncClient. The user is responsible for calling
+     *         {@link #doCloseJavaClient()} when finished with the client
```
dbwiddis, line 180: The user may be responsible for closing this client, but this points them to the wrong client to close. :-)

dbwiddis (changes requested):

src/test/java/org/opensearch/sdk/TestSDKClient.java
owaiskazi19: This is how `doCloseJavaClient` looks like:

```java
public void doCloseJavaClient() throws IOException {
        if (restClient != null) {
            restClient.close();
        }
    }
```

We are actually closing the restClient and it's the same for `JavaClient` and `JavaAsyncClient` as I have created a new method to `initalizeTransport` where we actually build the `restClient`. I'm not sure how different `doCloseJavaAsyncClient` gonna look like.

src/main/java/org/opensearch/sdk/SDKClient.java
dbwiddis: If we're using the same close method for both clients that should be documented here as well.

dbwiddis (approved):
LGTM for now.   The whole initialize and close thing really could use an overhaul at some point.

ryanbogan (approved):

PR #581, opened 2023-03-17
Create RepositoryExtension interface

### LINKED ISSUE #329: [FEATURE] Create RepositoryExtension interface
**Is your feature request related to a problem?**

Extension points (see #315) will be (optionally) implemented on Extensions. Eventually these points may need to be communicated to OpenSearch to be registered in the appropriate service or module.

Allowing extensions to inherit from appropriate interfaces depending on their type will match the current Plugin implementation.

**What solution would you like?**

Add a new interface RepositoryExtension with default implementations for extension points matching RepositoryPlugin.java.

kokibas:
### Description

### Issues Resolved
Fixes #329 


src/test/java/org/opensearch/sdk/TestExtensionInterfaces.java
```diff
@@ -123,15 +129,6 @@ public void testGetFieldFilter() {
         Assertions.assertNotNull(predicate);
     }
 
-    @Test
```
dbwiddis, line 126: Looks like you have an older version of this class before this was merged.  So you shouldn't remove this.

Rebasing or merging to main should resolve this, or you can copy-paste this back from the current class on main.

src/test/java/org/opensearch/sdk/TestExtensionInterfaces.java
```diff
@@ -150,4 +147,63 @@ void testSystemIndexExtension() {
...
+        RepositoryExtension repositoryExtension = new RepositoryExtension() {
+        };
+
+        Map<String, Repository.Factory> repositories = repositoryExtension.getInternalRepositories(
+            env,
+            namedXContentRegistry,
+            clusterService,
+            recoverySettings
+        );
+        assertEquals(Collections.emptyMap(), repositories);
```
dbwiddis, line 207: Generally easier to assertTrue that the map `isEmpty()`.

src/test/java/org/opensearch/sdk/TestExtensionInterfaces.java
```diff
@@ -150,4 +147,63 @@ void testSystemIndexExtension() {
...
+    }
+
+    @Test
+    void getInternalRepositoriesReturnsEmptyMapByDefault() {
```
dbwiddis, line 191: I know we have a mix of extension interface tests vs. extension point tests. I'd prefer to name this method `testRepositoryExtension()`

dbwiddis (changes requested):
Thanks for this contribution. Looks like you're out of sync with main and deleting someone else's addition, so please sync that up.   Also the DCO check is failing, make sure you use `-s` to sign your commits.

PR #600, opened 2023-03-22
Created ExtensibleExtension interface

bexultantokan:
### Description
Copied the content of the ExtensiblePlugin and pasted it to the new ExtensibleExtension file. Changed the wording from "plugin" to "extension."

### Issues Resolved
Created the interface to help move away from plugins to extensions.


saratvemulapalli:
@dbwiddis @owaiskazi19 do we need `loadExtensions` to be supported by extensions? 
Sounds like a feature where extensions can be built on top of extensions?

owaiskazi19:
> @dbwiddis @owaiskazi19 do we need `loadExtensions` to be supported by extensions? Sounds like a feature where extensions can be built on top of extensions?

Is this similar to having dependent extensions for an extension?

dbwiddis:
> @dbwiddis @owaiskazi19 do we need `loadExtensions` to be supported by extensions? Sounds like a feature where extensions can be built on top of extensions?

Yeah it seems like a way to say "hey, look at my `getActions()` and you can also use the `getActions()` from these other extensions.   I can see some use cases for "dependent extensions" which are tightly integrated with each other.

src/main/java/org/opensearch/sdk/ActionExtension.java
```diff
@@ -189,3 +190,5 @@ default Collection<RequestValidators.RequestValidator<IndicesAliasesRequest>> in
     }
 
 }
+
+
```
dbwiddis, line 194: The little red symbol here tells me this file doesn't end with the proper end of file marker.   And sure enough, it's failing the spotless check in CI.   Running `./gradlew spotlessApply` will fix this.

dbwiddis (commented):
LGTM except need to run spotless to fix formatting.

src/main/java/org/opensearch/sdk/ActionExtension.java
dbwiddis: Actually, you shouldn't be changing this file at all!

bexultantokan:
Hello @dbwiddis , could you please review my latest commit?

src/main/java/org/opensearch/sdk/ExtensibleExtension.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.List;
+
+/**
+ * An extension point for {@link Extension} implementations to be themselves extensible.
+ *
+ * This class provides a callback for extensible Extensions to be informed of other Extensions
+ * which extend them.
+ *
+ * @opensearch.api
```
dbwiddis, line 20: SDK doesn't use this annotation so it will fail compiling.  Please remove this.

We use package-based API definitions.

src/main/java/org/opensearch/sdk/ExtensibleExtension.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public interface ExtensibleExtension {
+
+    /**
+     * Extension point for external Extensions to be extendable
+     *
+     * @opensearch.api
```
dbwiddis, line 27: Same here.

dbwiddis (changes requested):
Looks good other than the annotation that doesn't work here.

bexultantokan:
@dbwiddis could you review the recent changes, please? Much appreciated.

dbwiddis (changes requested):
Sorry if the previous review wasn't clear, it was only the `@opensearch.api` annotation that needed to be deleted.  The class itself needs the rest of that javadoc.

dbwiddis (approved):
LGTM!

ryanbogan (approved):

ryanbogan:
Thanks for the contribution!

PR #605, opened 2023-03-24
Added support for headers for ExtensionRestRequest

### LINKED ISSUE #431: [FEATURE] Add support for headers for ExtensionRestRequest
Currently, there's no provision to fetch headers from ExtensionRestRequest which is required for various features of Anomaly Detector plugin and for the future.

Create 
1. `header` - To get the value of the header
2. `getAllHeaderValues` - To get all values for the header
3. `getHeaders` - To get all of the headers and values associated with the headers

These methods are similar to what RestRequest currently has.

_A clear and concise description of any alternative solutions or features you've considered._

_Add any other context or screenshots about the feature request here._

nassipkali:
### Description
Added support for headers for ExtensionRestRequest

### Issues Resolved
#431

### DIFF
--- a/src/main/java/org/opensearch/sdk/AnalysisExtension.java
+++ b/src/main/java/org/opensearch/sdk/AnalysisExtension.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.sdk;
+
+import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.analysis.CharFilter;
+import org.apache.lucene.analysis.TokenFilter;
+import org.apache.lucene.analysis.Tokenizer;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.env.Environment;
+import org.opensearch.index.IndexSettings;
+import org.opensearch.index.analysis.AnalyzerProvider;
+import org.opensearch.index.analysis.CharFilterFactory;
+import org.opensearch.index.analysis.PreBuiltAnalyzerProviderFactory;
+import org.opensearch.index.analysis.PreConfiguredCharFilter;
+import org.opensearch.index.analysis.PreConfiguredTokenFilter;
+import org.opensearch.index.analysis.PreConfiguredTokenizer;
+import org.opensearch.index.analysis.TokenFilterFactory;
+import org.opensearch.index.analysis.TokenizerFactory;
+import org.opensearch.indices.analysis.AnalysisModule;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+import static java.util.Collections.emptyList;
+import static java.util.Collections.emptyMap;
+
+/**
+ * OpenSearch doesn't have any automatic mechanism to share these components between indexes. If any component is heavy enough to warrant
+ * such sharing then it is the Extension's responsibility to do it in their {@link AnalysisModule.AnalysisProvider} implementation. We recommend against doing
+ * this unless absolutely necessary because it can be difficult to get the caching right given things like behavior changes across versions.
+ */
+public interface AnalysisExtension {
+    /**
+     * Override to add additional {@link CharFilter}s. See {@link #requiresAnalysisSettings(AnalysisModule.AnalysisProvider)}
+     * how to on get the configuration from the index.
+     */
+    default Map<String, AnalysisModule.AnalysisProvider<CharFilterFactory>> getCharFilters() {
+        return emptyMap();
+    }
+
+    /**
+     * Override to add additional {@link TokenFilter}s. See {@link #requiresAnalysisSettings(AnalysisModule.AnalysisProvider)}
+     * how to on get the configuration from the index.
+     */
+    default Map<String, AnalysisModule.AnalysisProvider<TokenFilterFactory>> getTokenFilters() {
+        return emptyMap();
+    }
+
+    /**
+     * Override to add additional {@link Tokenizer}s. See {@link #requiresAnalysisSettings(AnalysisModule.AnalysisProvider)}
+     * how to on get the configuration from the index.
+     */
+    default Map<String, AnalysisModule.AnalysisProvider<TokenizerFactory>> getTokenizers() {
+        return emptyMap();
+    }
+
+    /**
+     * Override to add additional {@link Analyzer}s. See {@link #requiresAnalysisSettings(AnalysisModule.AnalysisProvider)}
+     * how to on get the configuration from the index.
+     */
+    default Map<String, AnalysisModule.AnalysisProvider<AnalyzerProvider<? extends Analyzer>>> getAnalyzers() {
+        return emptyMap();
+    }
+
+    /**
+     * Override to add additional pre-configured {@link Analyzer}s.
+     */
+    default List<PreBuiltAnalyzerProviderFactory> getPreBuiltAnalyzerProviderFactories() {
+        return emptyList();
+    }
+
+    /**
+     * Override to add additional pre-configured {@link CharFilter}s.
+     */
+    default List<PreConfiguredCharFilter> getPreConfiguredCharFilters() {
+        return emptyList();
+    }
+
+    /**
+     * Override to add additional pre-configured {@link TokenFilter}s.
+     */
+    default List<PreConfiguredTokenFilter> getPreConfiguredTokenFilters() {
+        return emptyList();
+    }
+
+    /**
+     * Override to add additional pre-configured {@link Tokenizer}.
+     */
+    default List<PreConfiguredTokenizer> getPreConfiguredTokenizers() {
+        return emptyList();
+    }
+
+    /**
+     * Override to add additional hunspell {@link org.apache.lucene.analysis.hunspell.Dictionary}s.
+     * @return an empty Map that has a key of type String and a value of type org.apache.lucene.analysis.hunspell.Dictionary.
+     */
+    default Map<String, org.apache.lucene.analysis.hunspell.Dictionary> getHunspellDictionaries() {
+        return emptyMap();
+    }
+
+    /**
+     * Mark an {@link AnalysisModule.AnalysisProvider} as requiring the index's settings.
+     * @param provider this is an instance of the basic interface for analysis components.
+     * @param <T> is a type parameter that specifies the type of the analysis module that will be returned by the provider.
+     */
+    static <T> AnalysisModule.AnalysisProvider<T> requiresAnalysisSettings(AnalysisModule.AnalysisProvider<T> provider) {
+        return new AnalysisModule.AnalysisProvider<T>() {
+            @Override
+            public T get(IndexSettings indexSettings, Environment environment, String name, Settings settings) throws IOException {
+                return provider.get(indexSettings, environment, name, settings);
+            }
+
+            @Override
+            public boolean requiresAnalysisSettings() {
+                return true;
+            }
+        };
+    }
+}
--- a/src/test/java/org/opensearch/sdk/TestExtensionInterfaces.java
+++ b/src/test/java/org/opensearch/sdk/TestExtensionInterfaces.java
@@ -20,6 +20,7 @@ import org.opensearch.common.breaker.CircuitBreaker;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.index.mapper.Mapper;
 import org.opensearch.index.mapper.MetadataFieldMapper;
+import org.opensearch.indices.analysis.AnalysisModule;
 import org.opensearch.indices.breaker.BreakerSettings;
 import org.opensearch.ingest.Processor;
 import org.opensearch.test.OpenSearchTestCase;
@@ -188,5 +189,24 @@ public class TestExtensionInterfaces extends OpenSearchTestCase {
         };
         assertTrue(repositoryExtension.getRepositories(null, null, null, null).isEmpty());
         assertTrue(repositoryExtension.getInternalRepositories(null, null, null, null).isEmpty());
+    }
+
+    @Test
+    public void testAnalysisExtension() {
+        AnalysisExtension analysisExtension = new AnalysisExtension() {
+        };
+        AnalysisModule.AnalysisProvider<?> provider = (indexSettings, environment, name, settings) -> null;
+        AnalysisModule.AnalysisProvider<?> analysisProvider = AnalysisExtension.requiresAnalysisSettings(provider);
+
+        assertTrue(analysisExtension.getCharFilters().isEmpty());
+        assertTrue(analysisExtension.getTokenFilters().isEmpty());
+        assertTrue(analysisExtension.getTokenizers().isEmpty());
+        assertTrue(analysisExtension.getAnalyzers().isEmpty());
+        assertTrue(analysisExtension.getPreBuiltAnalyzerProviderFactories().isEmpty());
+        assertTrue(analysisExtension.getPreConfiguredCharFilters().isEmpty());
+        assertTrue(analysisExtension.getPreConfiguredTokenFilters().isEmpty());
+        assertTrue(analysisExtension.getPreConfiguredTokenizers().isEmpty());
+        assertTrue(analysisExtension.getHunspellDictionaries().isEmpty());
+        assertTrue(analysisProvider.requiresAnalysisSettings());
     }
 }


dbwiddis:
> @dbwiddis should I implement methods that returns null in SDKHttpRequest?

If it's something we don't currently pass, yes.  We can add those things later.  Although if it's something we can easily pass over transport (like http version) you might want to add that to your existing PR(s).

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -129,7 +129,7 @@ public class ExtensionsRunner {
      */
     private final Injector injector;
 
-    private final SDKNamedXContentRegistry sdkNamedXContentRegistry;
+    private SDKNamedXContentRegistry sdkNamedXContentRegistry;
```
dbwiddis, line 132: Please keep this as `final`.  Also all the existing handlers down at line 138 are now final with a just-merged PR.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -138,7 +138,10 @@ public class ExtensionsRunner {
     private ExtensionsIndicesModuleRequestHandler extensionsIndicesModuleRequestHandler = new ExtensionsIndicesModuleRequestHandler();
     private ExtensionsIndicesModuleNameRequestHandler extensionsIndicesModuleNameRequestHandler =
         new ExtensionsIndicesModuleNameRequestHandler();
-    private ExtensionsRestRequestHandler extensionsRestRequestHandler = new ExtensionsRestRequestHandler(extensionRestPathRegistry);
+    private ExtensionsRestRequestHandler extensionsRestRequestHandler = new ExtensionsRestRequestHandler(
```
dbwiddis, line 141: This is now final and the initialization moved down into the constructor, properly sequenced.  Please rebase to main and update here.

src/main/java/org/opensearch/sdk/SDKHttpRequest.java
```diff
@@ -0,0 +1,138 @@
...
+    private final RestRequest.Method method;
+    private final BytesReference content;
+
+    /**
+    * Instantiates this class with a copy of request
+    *
+    * @param request The request
+    */
+    public SDKHttpRequest(ExtensionRestRequest request) {
+        this.method = request.method();
```
dbwiddis, line 35: We need more than just these two.  Path should be copied over although it might be that we want to change that in the `ExtensionRestRequest` to be the `uri()`. 

 The headers you are adding needs to be copied over.  

Consider sending over the HTTPVersion, it's just a wrapper for major and minor version ints that you can serialize.

src/main/java/org/opensearch/sdk/SDKNamedXContentRegistry.java
```diff
@@ -43,7 +43,7 @@ public class SDKNamedXContentRegistry {
     /**
      * Creates an empty registry.
      */
-    private SDKNamedXContentRegistry() {
+    SDKNamedXContentRegistry() {
```
dbwiddis, line 46: Please leave this private.

src/main/java/org/opensearch/sdk/handlers/ExtensionsRestRequestHandler.java
```diff
@@ -50,6 +60,15 @@ public ExtensionsRestRequestHandler(ExtensionRestPathRegistry restPathRegistry)
      */
     public RestExecuteOnExtensionResponse handleRestExecuteOnExtensionRequest(ExtensionRestRequest request) {
 
+        RestRequest.Method method = request.method();
```
dbwiddis, line 63: you probably don't need to extract any of these things from the request here since you are passing the request object itself into the SDKHttpRequest constructor.   You'll extract them there and assign them to those fields.

dbwiddis (changes requested):
Good start, but need to move lots of code into the SDKHttpRequest class.

Also please rebase with main first, the code at the top of ExtensionsRunner has changed.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -141,7 +141,10 @@ public class ExtensionsRunner {
     private final ExtensionsIndicesModuleRequestHandler extensionsIndicesModuleRequestHandler = new ExtensionsIndicesModuleRequestHandler();
     private final ExtensionsIndicesModuleNameRequestHandler extensionsIndicesModuleNameRequestHandler =
         new ExtensionsIndicesModuleNameRequestHandler();
-    private final ExtensionsRestRequestHandler extensionsRestRequestHandler = new ExtensionsRestRequestHandler(extensionRestPathRegistry);
+    private final ExtensionsRestRequestHandler extensionsRestRequestHandler = new ExtensionsRestRequestHandler(
+        extensionRestPathRegistry,
+        sdkNamedXContentRegistry
```
dbwiddis, line 146: This isn't initialized yet so it would be null when you passed it here.   The extensionRestPathRegistry is initialized on  line 103 so it's fine.  So just declare the request handler variable here, and move the initialization down somewhere below where sdkNamedXContent is initialized (line 183)

src/main/java/org/opensearch/sdk/SDKHttpRequest.java
```diff
@@ -0,0 +1,142 @@
...
+        this.content = request.content();
+        this.headers = request.headers();
+        this.content = request.content();
```
dbwiddis, line 40: you've duplicated content and are still missing the uri or path (I'm not sure the difference, you might need to look into that.)

src/main/java/org/opensearch/sdk/handlers/ExtensionsRestRequestHandler.java
```diff
@@ -32,13 +36,15 @@
 public class ExtensionsRestRequestHandler {
     private static final Logger logger = LogManager.getLogger(ExtensionsRestRequestHandler.class);
     private final ExtensionRestPathRegistry extensionRestPathRegistry;
+    private NamedXContentRegistry namedXContentRegistry;
```
dbwiddis, line 39: This needs to be the "wrapper" `SDKNamedXContentRegistry` that doesn't change (and make it `final`).  We will use `getRegistry()` on each Rest call to fetch the most recent registry.

src/main/java/org/opensearch/sdk/handlers/ExtensionsRestRequestHandler.java
```diff
@@ -32,13 +36,15 @@
...
 
     /**
      * Instantiate this class with an existing registry
      *
      * @param restPathRegistry The ExtensionsRunnerer's REST path registry
      */
-    public ExtensionsRestRequestHandler(ExtensionRestPathRegistry restPathRegistry) {
+    public ExtensionsRestRequestHandler(ExtensionRestPathRegistry restPathRegistry, SDKNamedXContentRegistry sdkNamedXContentRegistry) {
+        this.namedXContentRegistry = sdkNamedXContentRegistry.getRegistry();
```
dbwiddis, line 47: (Similar to above, don't call `getRegistry()` here on initialization, call it when it's needed in the handler method.

src/main/java/org/opensearch/sdk/handlers/ExtensionsRestRequestHandler.java
```diff
@@ -49,6 +55,8 @@ public ExtensionsRestRequestHandler(ExtensionRestPathRegistry restPathRegistry)
      * @return A response acknowledging the request.
      */
     public RestExecuteOnExtensionResponse handleRestExecuteOnExtensionRequest(ExtensionRestRequest request) {
+        SDKHttpRequest sdkHttpRequest = new SDKHttpRequest(request);
+        RestRequest restRequest = RestRequest.request(namedXContentRegistry, sdkHttpRequest, null);
```
dbwiddis, line 59: Here is where you will use `.getRegistry()`.

src/main/java/org/opensearch/sdk/SDKHttpRequest.java
```diff
@@ -0,0 +1,146 @@
...
+    * Instantiates this class with a copy of request
+    *
+    * @param request The request
+    */
+    public SDKHttpRequest(ExtensionRestRequest request) {
+        this.method = request.method();
+        this.content = request.content();
+        this.headers = request.headers();
+
+        Path path = Paths.get(request.path());
```
dbwiddis, line 43: The REST path is not the same as a file path, so this isn't the right type to use here.

What I would like to do is change the `ExtensionRestRequest` itself to pass the `uri()` instead of the `path()`, and this would just be `this.uri = request.uri()`.

src/main/java/org/opensearch/sdk/SDKHttpRequest.java
```diff
@@ -0,0 +1,146 @@
...
+        return null;
+    }
+
+    /**
+    * This method returns request HTTP protocol version
+    * @return version of request HTTP protocol
+    */
+    @Override
+    public HttpVersion protocolVersion() {
+        return null;
```
dbwiddis, line 98: Since you are changing the request to add headers, and I've asked to pass uri instead of path, let's also add passing this enum in the request and returning it here.

dbwiddis (changes requested):

src/main/java/org/opensearch/sdk/SDKHttpRequest.java
nassipkali: > The REST path is not the same as a file path, so this isn't the right type to use here.
> 
> What I would like to do is change the `ExtensionRestRequest` itself to pass the `uri()` instead of the `path()`, and this would just be `this.uri = request.uri()`.

Should I just add uri to `ExtensionRestRequest`? Or I should also remove `path()`, `path` from `ExtensionRestRequest`

dbwiddis: > > The REST path is not the same as a file path, so this isn't the right type to use here.
> > What I would like to do is change the `ExtensionRestRequest` itself to pass the `uri()` instead of the `path()`, and this would just be `this.uri = request.uri()`.
> 
> Should I just add uri to `ExtensionRestRequest`? Or I should also remove `path()`, `path` from `ExtensionRestRequest`

You should add uri as a field and `uri()` as a method.

You should remove path as a field and `path()` as a method.

dbwiddis: (And you should also add a new `HttpRequest.HttpVersion` field and new method `protocolVersion()` to retrieve it.)

dbwiddis:
Hey @nassipkali a quick update.  I implemented the "second half" of this in #623 and learned a few lessons along the way that may change how you want to implement this.

The biggest difference between what I've said and what has to happen is in the `uri` and `params` fields.   The simplest explanation is that there are two ways to send params in a request.  One is the traditional URL parameters, such as `http://example.com?foo=bar` which assigns parameter "foo" the value of "bar".  The other is integrated into how OpenSearch handles Rest Requests with named parameters, so the REST API might say `GET /example/{foo}` and a user submitting a request would send `GET /example/bar` and it would also assign "foo" the value of "bar".

Related:
1. "path" is just the portion of "uri" before the "?".
2. The static method we've chosen to recreate the `RestRequest` from an `HttpRequest` does not consider the params that have already been parsed.  The only way to get them back in (see my PR) is to mash them back into the URI.

In my PR, it's easy, I use the path and recreate a full `uri` by reassembling the `?foo=bar` portion.  This may be sufficient, and might be a reason to keep path and *not* URI.  However, this also ends up creating a URI that the user didn't actually send, if we wanted to handle it differently in error messages, etc.

Alternately, we might consider making a subclass of `RestRequest` (call it `SDKRestRequest`) which:
 - we can still pass as a parameter meeting the `RestRequest` type
 - we can construct using a `protected` inherited constructor, preserving the `params` as they are already parsed

I'm leaning toward this second option right now, but it's your PR so I will let you decide which one you would rather do!

dbwiddis:
> Moving the conversation here: [#623 (comment)](https://github.com/opensearch-project/opensearch-sdk-java/pull/623#discussion_r1153569115)

Can we please keep the conversation about this PR on this PR?  It's ok to reference #623 conversation but this is the actual work being done and any fixes for anything that you think #623 missed can be added in this PR.

nassipkali:
> @nassipkali This comment is an attempt to merge conversations both on this PR and on #623. Since this PR is the one still open I think it is the best place to discuss, and make any changes to #623's implementation in this PR.
> 
> A useful reference for this conversation is the DESIGN document showing how the Rest Requests are handled: https://github.com/opensearch-project/opensearch-sdk-java/blob/main/DESIGN.md#extension-rest-actions-walk-through
> 
> At a high level, this is how REST requests are handled:
> 
>     1. A user submits a REST request via the API (either command-line `curl` or via some user interface somewhere) which is received by OpenSearch as an `HttpRequest`.  Key components of this request are:
>        
>        * The method (GET, PUT, etc.)
>        * The uri (the full command string after the http coordinates)
>        * Any headers, most importantly "Content-Type" which identifies the format of the body
>        * The body (content)
> 
>     2. Opensearch process the `HttpRequest` and turns it into a `RestRequest`.  As part of this process:
>        
>        * The `uri` is reduced to a `path` by stripping any part after `?` and adding it into a `params` map.
>        * The `path` is further matched against its matching REST API spec.  If that API included a "named parameter" (a registered path like `/hello/{name}` for example) then that is also added to the `params` map (equivalent to `?name=foo` in the URI if the user used `/hello/foo` as the path).
>        * The `Content-Type` header is parsed into an `XContentType` field for convenience.
> 
>     3. In OpenSearch and in plugins, we can use this `RestRequest` exactly as it is!  It's in the same JDK, with the same object, same security login rules, etc.  All is well.   In an extension, we also want to use a `RestRequest` but we don't have the same object and there is no existing method to serialize/deserialize that request.
>        
>        * The purpose of the `ExtensionRestRequest` object is to fulfill that purpose, of carrying information between OpenSearch and the Extension via its writeable functions (`StreamInput` constructor and `writeTo(StreamOutput)` method)
>        * Up until [Generate RestRequest to pass to Rest Handlers #623](https://github.com/opensearch-project/opensearch-sdk-java/pull/623), we were _also_ using the `ExtensionRestRequest` in place of the `RestRequest` inside extensions, and trying to copy over methods as we needed them. We don't want to do this any more.
> 
>     4. In extensions, we need _most_, but not all, of the information in the `RestRequest`.
>        
>        * We want the Method, uri or path (see below), params, selected headers, and body
>        * We do not want any security sensitive (Authorization) or privacy sensitive (IP address, etc.) headers
>        * We may want to add our own headers to send to extension (such as an identiy/auth token)
> 
>     5. As noted above in a comment, while some `params` come from the full URI, some also come from the registered path with named param wildcards.  We have to convey the `params()` map from the RestRequest on the OpenSearch side and make sure it somehow gets into the `params()` map on the RestRequest on the SDK side.
>        
>        * This has not yet been solved!
>        * Currently we just send the `params()` map in the `ExtensionsRestRequest.  We need to continue to do this.
>        * We need to get those `params()` into the new `RestRequest` object.  We have two options for this:
>          
>          * Option 1: just set the `params()` via constructor.  This can be done using the `protected` constructor on `RestClient` but requires us to create a subclass and call `super()`.  I think this is probably how we want to solve this
>          * Option 2: "hack" the params into the URI so that the existing static constructor I've used in [Generate RestRequest to pass to Rest Handlers #623](https://github.com/opensearch-project/opensearch-sdk-java/pull/623) can understand them (basically add `?foo=bar&baz=qux` etc. to the path / uri).  While this works (see [Generate RestRequest to pass to Rest Handlers #623](https://github.com/opensearch-project/opensearch-sdk-java/pull/623)) it's unnecessarily converting from a map to a string and back to a map.
>          * If we go with option 1, we can keep the full URI as is and it is better to change ExtensionRestRequest to send the full URI and not just the path
>          * If we go with option 2 we have to "hack" the URI and it is better to send just the path (as we currently do)
> 
>     6. In addition to deciding how to handle uri + params, we should _add_ to the ExtensionRestRequest:
>        
>        * The (filtered) headers map.  YOu're already doing that in this PR
>        * The HttpVersion enum value.
> 
> 
> Currently I have a temporary `RestRequest` anonymous subclass that I've created in the ExtensionRestRequestHandler class. Ideally that will be replaced by a `RestRequest` subclass that you create to solve the issues above in option 1. As an alternative, you could implement option 2 and update the existing anonymous subclass with more data from the ExtensionRestRequest.
> 
> Let me know if you have any questions... I know it's complex and I didn't really understand a lot of this until I had to go through and implement and test everything in #623.

@dbwiddis Looks like I should do first step from OpenSearch side. I want to choose first option: pass uri from OpenSearch side. `RestRequest` splits uri to params and path (path != uri) from `HttpRequest`. We can pass uri to `ExtensionRestRequest` if we use on RestRequest  `getHttpRequest()`:
`public HttpRequest getHttpRequest() {`
 `       return httpRequest;`
 `}`
 And then get uri from `httpRequest` that we got from `getHttpRequest()` method and pass to `ExtensionRestRequest` as params.
 Is this correct way to pass uri to Extensions?

dbwiddis:
> Looks like I should do first step from OpenSearch side. 

Yep, although make sure you note in the PR description (first comment) not to merge the PR until you're done with this one so they can be merged back to back.

> I want to choose first option: pass uri from OpenSearch side. `RestRequest` splits uri to params and path (path != uri) from `HttpRequest`. We can pass uri to `ExtensionRestRequest` if we use on RestRequest `getHttpRequest()`:

Agreed this is a good plan.  And as mentioned let's grab the HttpVersion too.  

> And then get uri from `httpRequest` that we got from `getHttpRequest()` method and pass uri to `ExtensionRestRequest` as params.
> Is this correct way to pass uri to Extensions?

There are lots of ways to do it... main point is to get all the pieces you want and make sure the `writeTo()` and constructor taking `StreamInput` match up with all the items.   From our dicussion I think we want to:
 - change from passing `path` to passing `uri` (still just a string)
 - add on passing the header map (your draft pr does this)
 - add on passing the HttpVersion enum

nassipkali:
@dbwiddis What I should to do next in sdk after changes in ExtensionRestRequest?

nassipkali:
> @dbwiddis What I should to do next in sdk after changes in ExtensionRestRequest?

I found comment in code. I should implement SDKRestRequest class.

dbwiddis:
> > @dbwiddis What I should to do next in sdk after changes in ExtensionRestRequest?
> 
> 
> 
> I found comment in code. I should implement SDKRestRequest class.

That is one option. As an alternative you could do an anonymous subclass in the handler similar to the way the current request from my PR is done. 

Both are good ideas, do whichever one you find easier.

PR #608, opened 2023-03-25
Changed the test for RegexValidator class

### LINKED ISSUE #591: [BUG] Hello World sample validator reverses logic
### What is the bug?

A sample validated setting was added to the HelloWorld extension in #421.  It is based on original assumptions that a matching validator would fail.  During review of the companion PR, the implementation reversed the logic so it now does the opposite of what's intended.

The unit test was missing the `@Test` annotation so it wasn't being adequately tested, as I noticed when reviewing code coverage.

1. Add the `@Test` annotation to `TestHelloWorldExtension` method `testValidatedSettings()`.
2. Run `gradle check`

Tests pass.

Possible resolutions:
1. (Easy) Change the regular expression from "forbidden" to one that verifies a word doesn't exist.  This is a [confusing regex](https://stackoverflow.com/questions/406230/regular-expression-to-match-a-line-that-doesnt-contain-a-word): "^((?!forbidden).)*$"
2. (Preferred) Update the validator to take a boolean parameter switching between "fail if it matches" or "fail if it doesn't match".  This provides the most flexibility.

Bonus points:
 - Add more validated settings like IP address matching

Kuanysh-kst:
### Description
In this PR, I am refactoring the test so that it passes, this PR is related to this PR: https://github.com/opensearch-project/OpenSearch/pull/6823

### Issues Resolved
#591 


src/test/java/org/opensearch/sdk/sample/helloworld/TestHelloWorldExtension.java
```diff
@@ -245,11 +245,8 @@ public void onFailure(Exception e) {
         assertEquals("failed to find action [" + UnregisteredAction.INSTANCE + "] to execute", ex.getMessage());
     }
 
+    @Test
     public void testValidatedSettings() {
-        final String expected = randomAlphaOfLengthBetween(1, 5);
-        final String actual = VALIDATED_SETTING.get(Settings.builder().put(VALIDATED_SETTING.getKey(), expected).build());
-        assertEquals(expected, actual);
```
dbwiddis, line 251: You're actually removing the test case here that a setting *not* vailing validation doesn't throw an exception.  So this part needs to stay, just change your expected away from a random alpha to a string you choose.

Also I don't see `VALIDATED_SETTING` being updated to use the new boolean.  We would want to create two settings, one with the forbidden regex (this one) and one with a "must match" regex (I would suggest an IP address matcher, or file path matcher, or URL matcher, or any of the simpler matches here:  https://digitalfortress.tech/tips/top-15-commonly-used-regex/

dbwiddis (changes requested):

src/test/java/org/opensearch/sdk/sample/helloworld/TestHelloWorldExtension.java
```diff
@@ -245,11 +245,8 @@ public void onFailure(Exception e) {
...
-
         final IllegalArgumentException exception = expectThrows(
             IllegalArgumentException.class,
             () -> VALIDATED_SETTING.get(Settings.builder().put("custom.validated", "it's forbidden").build())
```
Kuanysh-kst, line 252: Is there any difference if I use "custom.validated" or "custom.validate" , just because of "custom.validated" my test doesn't pass

dbwiddis: > Is there any difference if I use "custom.validated" or "custom.validate" , just because of "custom.validated" my test doesn't pass

The setting name is defined in the sample class that you are testing, so "validated" is not an existing setting.
https://github.com/opensearch-project/opensearch-sdk-java/blob/ebc684ae223e87f1e05712e086fc1dab911a3b5f/src/main/java/org/opensearch/sdk/sample/helloworld/ExampleCustomSettingConfig.java#L25-L30

dbwiddis: (and you probably need to add the boolean to that sample, and add another sample with the opposite boolean)

src/main/java/org/opensearch/sdk/sample/helloworld/ExampleCustomSettingConfig.java
```diff
@@ -17,16 +17,36 @@
...
-        new RegexValidator(FORBIDDEN_REGEX),
+        new RegexValidator(FORBIDDEN_REGEX, true),
+        Property.NodeScope,
+        Property.Dynamic
+    );
+
+    /**
+     * A string setting. If the string setting doesn't match the FORBIDDEN_REGEX string and parameter isMatching is false the validation will fail.
+     */
+    public static final Setting<String> VALIDATED_SETTING_FALSE = Setting.simpleString(
```
Kuanysh-kst, line 36: added to check if isMatching is false

src/main/java/org/opensearch/sdk/sample/helloworld/ExampleCustomSettingConfig.java
```diff
@@ -17,16 +17,36 @@
...
+        "custom.validate",
+        new RegexValidator(FORBIDDEN_REGEX, false),
         Property.NodeScope,
         Property.Dynamic
     );
 
+    /**
+     * A string setting. If the string setting matches the VALIDATED_URL_SETTING string and parameter isMatching is true the validation will fail.
+     */
+    public static final Setting<String> VALIDATED_URL_SETTING = Setting.simpleString(
```
Kuanysh-kst, line 46: I added it but haven't used it anywhere yet

src/main/java/org/opensearch/sdk/sample/helloworld/ExampleCustomSettingConfig.java
```diff
@@ -17,16 +17,36 @@
  * {@link ExampleCustomSettingConfig} contains the custom settings value and their static declarations.
  */
 public class ExampleCustomSettingConfig {
-    private static final String FORBIDDEN_REGEX = "forbidden";
-
+    private static final String FORBIDDEN_REGEX = "^((?!forbidden).)*$";
```
dbwiddis, line 20: While I listed this as a possibility in the issue, the whole purpose of the boolean is so that we don't need these confusing regexes that nobody understands.

So I'd really like this kept as "forbidden" with the boolean set for it to NOT match (the false case).

For the URL check you can set the boolean to true.  This will effectively test both boolean values and you don't need to create a third one.

Kuanysh-kst: can I change the validation condition in the validate() method to this :
 ` if (isMatching && value.contains(pattern.pattern())) {`, replacing matcher() with contains() check

Kuanysh-kst: Can I use the contains() method in the validate() method in order not to compare through regular expressions, but to compare a word with a string? example :
`        public void validate(String value) {
            if (isMatching && value.contains(pattern.pattern())) {
                throw new IllegalArgumentException("Setting [" + value + "] does not match regex [" + pattern.pattern() + "]");
            } else if (!isMatching && !value.contains(pattern.pattern())) {
                throw new IllegalArgumentException("Setting [" + value + "] must match regex [" + pattern.pattern() + "]");
            }
        }`

dbwiddis: > Can I use the contains() method in the validate() method 

You could create a new `contains` validator, but it's no different than matching a regex.  That was the whole point of building this single validator.

Matching a string "foo" in a regex is pretty much exactly the same thing as `contains`.  If you want the entire word you can use beginning and end of line markers, etc.

We basically want two validators here and four tests:

1. A plain string "forbidden" (which happens to work out to the same as "must not contain" in a "must not match"), testing against one string containing the word (exception) and one not (does not throw).
2. A regex which "must match" testing against a string that matches (does not throw) and does not (exception)

src/main/java/org/opensearch/sdk/sample/helloworld/ExampleCustomSettingConfig.java
```diff
@@ -17,16 +17,14 @@
...
-
+    private static final String FORBIDDEN_VALUE = "forbidden";
     /**
-     * A string setting. If the string setting matches the FORBIDDEN_REGEX string, the validation will fail.
+     * A string setting. If the string setting matches the FORBIDDEN_REGEX string and parameter isMatching is true the validation will fail.
      */
     public static final Setting<String> VALIDATED_SETTING = Setting.simpleString(
         "custom.validate",
-        new RegexValidator(FORBIDDEN_REGEX),
+        new RegexValidator("^((?!" + FORBIDDEN_VALUE + ").)*$", true),
```
dbwiddis, line 26:
```suggestion
        new RegexValidator(FORBIDDEN_VALUE, false),
```

dbwiddis (changes requested):
Let's make this even simpler!

src/test/java/org/opensearch/sdk/sample/helloworld/TestHelloWorldExtension.java
```diff
@@ -253,9 +253,9 @@ public void testValidatedSettings() {
 
         final IllegalArgumentException exceptionTrue = expectThrows(
             IllegalArgumentException.class,
-            () -> VALIDATED_SETTING.get(Settings.builder().put(VALIDATED_SETTING.getKey(), "it's forbidden").build())
+            () -> VALIDATED_SETTING.get(Settings.builder().put(VALIDATED_SETTING.getKey(), "forbidden").build())
```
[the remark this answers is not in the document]
dbwiddis: Huh. So I learned something about Java Regex today.

At Regex 101 it clearly matches: https://regex101.com/r/y9g2s7/1

But in Java with `matches()` it doesn't:

```java
String regex = "forbidden";
Pattern pattern = Pattern.compile(regex);
System.out.println("Does 'forbidden' match? " + pattern.matcher("forbidden").matches());
System.out.println("Does 'it's forbidden' match? " + pattern.matcher("it's forbidden").matches());
```

> Does 'forbidden' match? true
> Does 'it's forbidden' match? false

So I looked up the Javadoc for [`matches()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/regex/Matcher.html#matches())
> Returns:
> true if, and only if, the entire region sequence matches this matcher's pattern

Compared to what I thought the behavior was, in [`find()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/regex/Matcher.html#find())
> Returns:
> true if, and only if, a subsequence of the input sequence matches this matcher's pattern

So we need to change `matches()` to `find()` on the Regex on the OpenSearch side to match this expectation.

dbwiddis (dismissed):

ryanbogan (dismissed):

joshpalis:
@Kuanysh-kst Looks like gradle check is failing. Please rebase your PR with SDK main

dbwiddis:
> @Kuanysh-kst Looks like gradle check is failing. Please rebase your PR with SDK main

The failure is the signature of the class just merged in companion PR.  May need to wait for snapshot publishing.

owaiskazi19:
> The failure is the signature of the class just merged in companion PR. May need to wait for snapshot publishing.

Snapshot has been published: https://github.com/opensearch-project/OpenSearch/actions/runs/4639883181

owaiskazi19:
Build is failing because of:
```
> Task :compileTestJava
/home/runner/work/opensearch-sdk-java/opensearch-sdk-java/src/test/java/org/opensearch/sdk/TestExtensionsRunner.java:145: error: no suitable constructor found for ExtensionRestRequest(Method,String,Map<Object,Object>,<null>,BytesArray,String)
        ExtensionRestRequest request = new ExtensionRestRequest(
                                       ^
    constructor ExtensionRestRequest.ExtensionRestRequest(Method,String,String,Map<String,String>,Map<String,List<String>>,XContentType,BytesReference,String,HttpVersion) is not applicable
      (actual and formal argument lists differ in length)
    constructor ExtensionRestRequest.ExtensionRestRequest(StreamInput) is not applicable
      (actual and formal argument lists differ in length)
1 error

> Task :compileTestJava FAILED

dbwiddis:
> Build is failing because of ...

I got confused on which PR this comment was on.

That method signature is the just-merged OS 6826, companion PR of #643.  We need to merge #643 and then rebase this PR and it should pass.

dbwiddis (approved):

ryanbogan (approved):

Issue #612, opened 2023-03-27
[BUG] SDKClient implementations are not thread safe and are difficult to close


dbwiddis:
### What is the bug?

The various clients exposed by `SDKClient` include this warning in the `initialize()` methods:
> `The user is responsible for calling {@link #doCloseJavaClients()} when finished with the client`

This is possible with the `SDKRestClient` which delegates its `close()` method to the internal RestHighLevelClient and implements Closeable.  Users can simply use a try-with-resources to wrap their usage of the client.

However, a "responsible" user can't really do that safely for the Java Clients (`OpenSearchAsyncClient` and `OpenSearchClient`) because:
 - Neither of these clients exposes a `close()` method nor implements `Closeable`
 - Initializing either of these clients sets the same `RestClient` class field, which is the real thing that needs to be closed
 - We could work around this multithreading issue by instantiating a new `SDKClient()` each time, but with the `getActions()` functionality they require initialization so we are using a singleton `SDKClient` object.
 - So ultimately we are working around this in AD work by initializing a single Sync OR Async client at the Extension level and passing it around.  This works effectively but we never close it per the docs.

### How can one reproduce the bug?

Call SDKClient.initializeJavaClient and initializeJavaAsyncClient from separate threads using different host/port options, and watch chaos commence.

### What is the expected behavior?

Users have the ability to use try-finally:
```java
try (OpenSearchClient client = someMethod()) {
  // do stuff
} finally {
  // some method that safely closes client
}
```

Alternately, enforce the singleton nature of each client and do not share the same `RestClient` between multiple clients and provide an easier closing method than the existing call to close all clients.

### Do you have any additional context?

I am not sure of the best approach but I'm certain that what we have is not it.

kokibas:
Hello @dbwiddis , I suggest using a singleton for each client to avoid sharing the same RestClient between multiple clients. This will allow closing the RestClient only once when all clients using it are finished.

dbwiddis:
> Hello @dbwiddis , I suggest using a singleton for each client to avoid sharing the same RestClient between multiple clients. This will allow closing the RestClient only once when all clients using it are finished.

That's one possible solution, but they aren't implemented as such.

That also complicates things like adding multiple configurations (in theory we should be able to connect to two different OpenSearch clusters and talk to both of them).

Another option is to just create one client per initialize request and make sure it only has its own resources (no shared RestClient), and manage closing it better, such as using finalizers.

PR #613, opened 2023-03-27
Add identity/authorization headers into Rest Client calls

### LINKED ISSUE #430: [FEATURE] Add identity/authorization headers into Rest Client calls

Presently `SDKClient` rest calls do not include security features.   Upcoming integration of identity features will include authentication and/or identity tokens that will need to be included with REST requests via the client.

See https://github.com/opensearch-project/OpenSearch/issues/3846

SDKClient REST calls need to edit the `RequestOptions` to add headers.

There may be a need to close/re-initialize the clients or just change the headers, that will need to be determined later.

kokibas:
### Description
Here I create a new Request Options object based on Request Options.DEFAULT using the to Builder() method, add the "Authorization" header with the token, and build a new RequestOptions object using the build() method.

### Issues Resolved
#430  


src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -460,12 +460,12 @@ public void delete(DeleteRequest request, ActionListener<DeleteResponse> listene
          * @see Requests#searchRequest(String...)
          */
         public void search(SearchRequest request, ActionListener<SearchResponse> listener) {
-            restHighLevelClient.searchAsync(request, RequestOptions.DEFAULT, listener);
+            RequestOptions options = RequestOptions.DEFAULT.toBuilder().addHeader("Authorization", "Bearer TOKEN_HERE").build();
```
dbwiddis, line 463: Thanks for showing the code on how to do this.  Can we make these two Strings something we set on the client?

dbwiddis (changes requested):
Thanks for this initial PR.  It's very helpful that you have shown how to add the headers.

Can we make a setter for the client that updates the options?  Something like `setOptions(options)` which you create this way?

dbwiddis (changes requested):
A few general comments here:
1. We need to continue to provide the default option if nothing is passed. Don't hard-code strings.
2. We should just have a single string for a header, and not require the user to pass two strings that we append to each other.
3. We should allow adding the header after instantiation. A `setHeader()` method would work for this but I think a much better idea is just having a `setRequestOptions()` method.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -460,12 +473,12 @@ public void delete(DeleteRequest request, ActionListener<DeleteResponse> listene
          * @see Requests#searchRequest(String...)
          */
         public void search(SearchRequest request, ActionListener<SearchResponse> listener) {
-            restHighLevelClient.searchAsync(request, RequestOptions.DEFAULT, listener);
+            RequestOptions options = RequestOptions.DEFAULT.toBuilder().addHeader(this.headerName, "Bearer" + this.token).build();
```
dbwiddis, line 476: Let's not construct the `options` here.  Have the user write this code and set the RequestOptions somewhere (with a default of `RequestOptions.DEFAULT` if it's not specified).

There are two ways of doing this.

1. Have an instance field `RequestOptions options` that you use here, and have a setter for it.
2. Add a new method that takes the options as the default.  (Probably not the best idea because we already have that ability to do with the client itself, which we are calling with `searchAsync()`.)

So I'd go with option 1.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -341,16 +348,22 @@ public static class SDKRestClient implements Closeable {
 
         private final SDKClient sdkClient;
         private final RestHighLevelClient restHighLevelClient;
+        private final String token;
```
dbwiddis, line 351: delete these and replace with a `RequestOptions` field, and add a setter.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -341,16 +348,22 @@ public static class SDKRestClient implements Closeable {
...
         /**
          * Instantiate this class wrapping a {@link RestHighLevelClient}.
          *
          * @param sdkClient The SDKClient instance.
          * @param restHighLevelClient The client to wrap.
+         * @param token The bearer token.
+         * @param headerName The header name.
          */
-        public SDKRestClient(SDKClient sdkClient, RestHighLevelClient restHighLevelClient) {
+        public SDKRestClient(SDKClient sdkClient, RestHighLevelClient restHighLevelClient, String token, String headerName) {
```
dbwiddis, line 362: just leave this as is

dbwiddis (changes requested):
Let's simplify this to just having a `RequestOptions` field.

dbwiddis (dismissed):
LGTM, not sure this is the "final solution" but should work for now until we figure out something better!

src/main/java/org/opensearch/sdk/SDKClient.java
kokibas: @dbwiddis I don't understand how the sdkclient communicate with opensearch, could you please give some explanations about this?

owaiskazi19:
@dbwiddis I'm not clear with what doesthis change actually does? Talks about adding identity/authorization headers but all I see is a setter and the default value replaced?

dbwiddis:
> @dbwiddis I'm not clear with what does this change actually does? Talks about adding identity/authorization headers but all I see is a setter and the default value replaced?

Per the issue being addressed, #430:
> SDKClient REST calls need to edit the RequestOptions to add headers.
> 
> There may be a need to close/re-initialize the clients or just change the headers, that will need to be determined later.

Such editing was explored earlier in this PR in commits now superceded, such as [this one](https://github.com/opensearch-project/opensearch-sdk-java/pull/613/commits/ba2a728bcdad16ebb0c48e93416a3e4df2984679):
```java
public void search(SearchRequest request, ActionListener<SearchResponse> listener) {
    restHighLevelClient.searchAsync(request, RequestOptions.DEFAULT, listener);
    RequestOptions options = RequestOptions.DEFAULT.toBuilder().addHeader("Authorization", "Bearer TOKEN_HERE").build();
kokibas marked this conversation as resolved.
    restHighLevelClient.searchAsync(request, options, listener);
}
```

Of course, that solution hard codes a string as the token, which obviously doesn't work.   But it points out the fact that if we want to use `client.search(request, listener)` we need the ability to pass other request options than the default.

Given the singleton nature of our RestClient, we can accept an incoming `RestRequest` from OpenSearch, which will (in the future) contain a token that we can use to send back.  We can set this token into the SDKRestClient using the setter in this PR.

Is this the final solution?  I doubt it.  It's not thread safe.  There are other client changes needed per #612.  This is also just a hack to our (deprecated) client used to aid migration, see also #625.

Per my approval comment:
> not sure this is the "final solution" but should work for now until we figure out something better!

If you have a better idea, this would be a good time to post it!

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -341,18 +341,23 @@ public static class SDKRestClient implements Closeable {
 
         private final SDKClient sdkClient;
         private final RestHighLevelClient restHighLevelClient;
+        private RequestOptions options = RequestOptions.DEFAULT;
 
         /**
          * Instantiate this class wrapping a {@link RestHighLevelClient}.
          *
-         * @param sdkClient The SDKClient instance.
+         * @param sdkClient           The SDKClient instance.
```
owaiskazi19, line 349:
```suggestion
         * @param sdkClient The SDKClient instance.
```

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -597,5 +601,7 @@ public Cancellable rolloverIndex(RolloverRequest rolloverRequest, ActionListener
         public Cancellable getAliases(GetAliasesRequest getAliasesRequest, ActionListener<GetAliasesResponse> listener) {
             return this.indicesClient.getAliasAsync(getAliasesRequest, RequestOptions.DEFAULT, listener);
         }
+
```
owaiskazi19, line 604:
```suggestion
```

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -597,5 +601,7 @@ public Cancellable rolloverIndex(RolloverRequest rolloverRequest, ActionListener
...
     }
+
```
owaiskazi19, line 606:
```suggestion
```

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -597,5 +601,7 @@ public Cancellable rolloverIndex(RolloverRequest rolloverRequest, ActionListener
         public Cancellable getAliases(GetAliasesRequest getAliasesRequest, ActionListener<GetAliasesResponse> listener) {
             return this.indicesClient.getAliasAsync(getAliasesRequest, RequestOptions.DEFAULT, listener);
```
owaiskazi19, line 602: Shouldn't we add `options` for all the requests then? Why only `search`?

owaiskazi19 (changes requested):
@dbwiddis this PR handles the Request Options and provides a way to add a new header to the async requets and not use only the DEFAULT one. It doesn't actually adds `identity/authorization headers`. A little confusing PR title got me thinking in the first place.

src/main/java/org/opensearch/sdk/SDKClient.java
dbwiddis: > Shouldn't we add `options` for all the requests then? Why only `search`?

\<insert facepalm emoji\>

Yeah, it should replace the default everywhere the default is used.  This is what happens when you look at the whole sausage-making process and miss things at the end.

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -631,5 +635,7 @@ public Cancellable rolloverIndex(RolloverRequest rolloverRequest, ActionListener
...
         }
+
```
owaiskazi19, line 638: @kokibas can we remove this extra line here

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -631,5 +635,7 @@ public Cancellable rolloverIndex(RolloverRequest rolloverRequest, ActionListener
...
     }
+
```
owaiskazi19, line 640: Here as well

owaiskazi19 (changes requested):
Still few places left to replace `RequestOptions.DEFAULT;` value with `options` like [this](https://github.com/opensearch-project/opensearch-sdk-java/pull/613/files#diff-da305de522ebd01a3b7821d9d3ee90d514a0b92d29a272fd5bb09bcb3a3e2c16R625). We should add options for all the clients which have DEFAULT currently.

src/main/java/org/opensearch/sdk/SDKClient.java
kokibas: @owaiskazi19  yes it added by spotless

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -545,6 +549,11 @@ public SDKClusterAdminClient(ClusterClient clusterClient) {
     public static class SDKIndicesClient {
 
         private final IndicesClient indicesClient;
+        private RequestOptions options = RequestOptions.DEFAULT;
+
+        public void setOptions(RequestOptions options) {
+            this.options = options;
+        }
```
owaiskazi19, line 556: @dbwiddis do you think we can pull this method to outer class and let inner class use the same and not duplciate it?

dbwiddis: > @dbwiddis do you think we can pull this method to outer class and let inner class use the same and not duplciate it?

@owaiskazi19 No, because it may change per-call.

From what I understand of the security framework (And @peternied or @scrawfor99 can correct me if I'm wrong) an incoming REST request from a user will get authenticated and then have a short-lived access token associated with that request.   This token will be forwarded in the `ExtensionRestRequest` object (we currently have a placeholder for it).

At that point we will need to do something with it that makes it accessible in the RestRequest being processed by Extensions. We haven't figured this part out yet.  It may be that we just add a header when passing the rest request, consisting of the desired header(s) to send back. 

Because of the need to close these clients when we're done with them, in my ideal world we do all our REST work inside a try-with-resources block, so in a RestHandler class we would:
```java
try (SDKRestClient restClient = sdkClient.initializeRestClient()) {
    RequestOptions options = RequestOptions.Builder().addHeader("Authorization", "Bearer TOKEN_HERE").build();
    restClient.setOptions(options);
    restClient.indices().search(foo);
} // client is auto-closed here
```
Or better, change the return value of setOptions to be `this` and do a fluent call
```java
restClient.setOptions(options).indices().search(foo);
```

There's still the question of how to get either this client (with its options) or the options themselves, from the Rest Handler to the Transport Action, but I presume we can add a field in the request being sent via `client.execute()`.

However, all of this assumes setting the options in a temporary SDKRestClient object that will go away and we'll re-initialize again with a different set of options the next time around.   If we put it up in the higher level SDKClient, that's a singleton and we'd share the same token/options/whatever everywhere.  That won't work well with multithreading/async calls.

dbwiddis:
> We should add options for all the clients which have DEFAULT currently.

I'm not so sure we should be adding more stuff to the Java Clients, rather than just documenting how to use them.  Do we actually have anything wrapped in the other clients yet?

src/main/java/org/opensearch/sdk/SDKClient.java
owaiskazi19: Makes sense! Tokens are short-lived. So it's fine to keep setter different for the clients.

owaiskazi19:
> I'm not so sure we should be adding more stuff to the Java Clients, rather than just documenting how to use them. Do we actually have anything wrapped in the other clients yet?

That was regards to `SDKIndicesClient` which @kokibas has already taken care of. Are you saying we should not add options for `SDKIndicesClient`?

dbwiddis:
> That was regards to `SDKIndicesClient` which @kokibas has already taken care of. Are you saying we should not add options for `SDKIndicesClient`?

Ah, sorry, I missed that context.  Yes the options should be at the highest level of the Rest Client and copied to the other clients.  If we removed the `static` from these classes, they would have access to the outer class object.

owaiskazi19 (approved):

dbwiddis (approved):

PR #619, opened 2023-03-29
Setup extension TLS

cwperks:
### Description

Companion PR in core: https://github.com/opensearch-project/OpenSearch/pull/6866
Companion PR in security repo: https://github.com/opensearch-project/security/pull/2599

I am opening a draft PR to solicit feedback for setting up extension TLS. The TLS setup here is modeled after the setup in security plugin, but any reference to http setup was removed and all transport setup kept in. While testing extensions with the Security plugin installed, I ran into issues where the extension was not able to handshake with the cluster as it was hitting this block: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java#L260-L268

```
//this is a netty request from a non-server node (maybe also be internal: or a shard request)
//and therefore issued by a transport client

//since OS 2.0 we do not support this any longer because transport client no longer available

final OpenSearchException exception = ExceptionUtils.createTransportClientNoLongerSupportedException();
log.error(exception.toString());
transportChannel.sendResponse(exception);
return;
```

This PR updates the handshake process and injects a ThreadContext header (`extension_unique_id`) that is added on all transport requests to core, to identify the transport request as explicitly coming from an extension.

For the TLS setup in this PR, I have added a general area in the extension settings (i.e. `helloworld-settings.yml`) where the TLS can be configured using any of the ConfigConstants present in SSLConfigConstants. All settings for SSL setup are prefixes with `ssl.transport.*` and SSL must be explicitly enabled with default to false (though I imagine this will change in a future PR as the security changes are closer to being finalized for the experimental release).

See a snippet from `helloworld-settings.yml` with SSL config:

```
ssl.transport.enabled: true
ssl.transport.pemcert_filepath: certs/extension-01.pem
ssl.transport.pemkey_filepath: certs/extension-01-key.pem
ssl.transport.pemtrustedcas_filepath: certs/root-ca.pem
ssl.transport.enforce_hostname_verification: false
path.home: /Users/cwperx/Projects/opensearch/opensearch-sdk-java // Looks for config/ folder under this path when looking for certs
```

I added commands in `Docs/CERTIFICATE_GENERATION.md` that you can use to generate certs for the extension and for the OS node to test with.

I was not sure how to add automated testing for this - I will open up an issue for adding tests

### Issues Resolved

https://github.com/opensearch-project/security/issues/2638


src/main/java/org/opensearch/sdk/ssl/DefaultSslKeyStore.java
```diff
@@ -0,0 +1,834 @@
...
+                san = sans.toString();
+            }
+        } catch (CertificateParsingException e) {
+            log.error("Issue parsing SubjectAlternativeName:", e);
+        }
+
+        return san;
+    }
+
+    // private List<String> getOtherName(List<?> altName) {
```
cwperks, line 803: Why is this commented out?

This originally came to the security plugin to fix a bug with certificate hot reloads: https://github.com/opensearch-project/security/pull/1369

I chose to comment this out because it would have required a dependency on bouncycastle.

cwperks:
@saratvemulapalli @dbwiddis @owaiskazi19 @peternied Would love to get your thoughts on this and companion PRs. This PR is fundamental to any security change going forward because without a secure TLS channel between an extension and nodes of the cluster they will not be able to communicate if the nodes have the security plugin installed. There is no option of turning off transport-layer encryption with the security plugin, its a must and since the core/security plugin will be issuing access tokens and forwarding to an extension the traffic should be encrypted.

If anyone has ideas to write automated tests then please chime in! In the security plugin, [SSLTest](https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/ssl/SSLTest.java) generally sets different SSL settings for nodes of a cluster and verifies cluster formation. It also calls on security plugin API endpoints to verify SSL info. For an extension, there is no endpoint to call and no concept of a cluster to form (at least atm) - what would be the best way to test this in an automated fashion?

saratvemulapalli:
Thanks @cwperks ill take a look.

src/main/java/org/opensearch/sdk/ssl/SSLNettyTransport.java
```diff
@@ -0,0 +1,238 @@
...
+import org.opensearch.indices.breaker.CircuitBreakerService;
+import org.opensearch.threadpool.ThreadPool;
+import org.opensearch.transport.SharedGroupFactory;
+import org.opensearch.transport.TcpChannel;
+import org.opensearch.transport.netty4.Netty4Transport;
+
+public class SSLNettyTransport extends Netty4Transport {
+
+    private static final Logger logger = LogManager.getLogger(SSLNettyTransport.class);
+    private final SslKeyStore ossks;
```
scrawfor99, line 44: Nit: Consider making this variable different named--I assume this is `OpenSearchSSLKeyStore`?

src/main/java/org/opensearch/sdk/ssl/SSLNettyTransport.java
```diff
@@ -0,0 +1,238 @@
...
+                            hostname,
+                            inetSocketAddress.getHostName(),
+                            inetSocketAddress.getHostString(),
+                            hostnameVerificationResovleHostName
+                        );
+                    }
+
+                    engine = sks.createClientTransportSSLEngine(hostname, inetSocketAddress.getPort());
+                } else {
+                    engine = sks.createClientTransportSSLEngine(null, -1);
```
scrawfor99, line 174: Please add a comment explaining this pattern.

src/main/java/org/opensearch/sdk/ssl/SslExceptionHandler.java
```diff
@@ -0,0 +1,29 @@
...
+package org.opensearch.sdk.ssl;
+
+import org.opensearch.rest.RestRequest;
+import org.opensearch.tasks.Task;
+import org.opensearch.transport.TransportRequest;
+
+public interface SslExceptionHandler {
+
+    default void logError(Throwable t, RestRequest request, int type) {
+        // no-op
```
scrawfor99, line 19: Should this extend runtime and then super()?

src/main/java/org/opensearch/sdk/ssl/util/PemKeyReader.java
```diff
@@ -0,0 +1,382 @@
...
+        try {
+            InputStream in = new FileInputStream(file);
+
+            try {
+                return readPrivateKey(in);
+            } finally {
+                safeClose(in);
+            }
+        } catch (FileNotFoundException e) {
+            throw new KeyException("could not fine key file: " + file);
```
scrawfor99, line 78: Typo: I think you mean "find" not 'fine'

src/main/java/org/opensearch/sdk/ssl/util/PemKeyReader.java
```diff
@@ -0,0 +1,382 @@
...
+            );
+        }
+
+        return Base64.getDecoder().decode(m.group(1));
+    }
+
+    private static String readContent(InputStream in) throws IOException {
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        try {
+            byte[] buf = new byte[8192];
```
scrawfor99, line 104: Nit: Add comment explaining why 2^13

src/main/java/org/opensearch/sdk/ssl/util/PemKeyReader.java
```diff
@@ -0,0 +1,382 @@
...
+        CertificateFactory fact = CertificateFactory.getInstance("X.509");
+        return (X509Certificate) fact.generateCertificate(in);
+    }
+
+    public static KeyStore loadKeyStore(String storePath, String keyStorePassword, String type) throws Exception {
+        if (storePath == null) {
+            return null;
+        }
+
+        if (type == null || !type.toUpperCase().equals(JKS) || !type.toUpperCase().equals(PKCS12)) {
```
scrawfor99, line 212: JKS are in PKCS12 I thought?

src/main/java/org/opensearch/sdk/ssl/util/SSLCertificateHelper.java
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    private static void logKeyStore(final KeyStore ks) {
+        try {
+            final List<String> aliases = toList(ks.aliases());
+            if (log.isDebugEnabled()) {
+                log.debug("Keystore has {} entries/aliases", ks.size());
+                for (String _alias : aliases) {
+                    log.debug(
+                        "Alias {}: is a certificate entry?{}/is a key entry?{}",
```
scrawfor99, line 146: Nit: Missing spaces in debug statement?

src/main/java/org/opensearch/sdk/ssl/util/Utils.java
```diff
@@ -0,0 +1,35 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.sdk.ssl.util;
+
+public class Utils {
```
scrawfor99, line 12: When is this necessary? Why not just for loop etc.?

scrawfor99 (commented):
Hi Craig, great work with this! I came over to get you started with some feedback. I think it looks good, so hopefully we can keep making headway based on all your hard work.

Docs/CERTIFICATE_GENERATION.md
```diff
@@ -0,0 +1,137 @@
+# Certificate Generation
+
+Use the script below to generate certificates for running an extension. 
+The certificate for the extension needs to be in the same chain as the 
+root certificate for the OpenSearch cluster defined in the `plugins.security.ssl.http.pemtrustedcas_filepath`
+setting of `opensearch.yml`
+
+
+## Certificate Generation Script
```
saratvemulapalli, line 9: @cwperks TLS solves the problem of identifying the extension. 
Shouldn't these scripts be part of the OpenSearch distribution packaged from security repo[1]?
These certs should be generated by a CA which security plugin understands (a.k.a security plugin tools should issue client certificates). 

[1] https://github.com/opensearch-project/security/tree/main/tools

src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -30,6 +32,8 @@ public class ExtensionSettings {
     private String opensearchAddress;
     private String opensearchPort;
 
+    private Map<String, String> otherSettings;
```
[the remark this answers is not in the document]
saratvemulapalli: +1 lets keep it simpler, lets make it `securitySettings`/`securityConfig`

src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -109,12 +147,20 @@ public static ExtensionSettings readSettingsFromYaml(String extensionSettingsPat
             if (extensionMap == null) {
                 throw new IOException("extension.yml is empty");
             }
+            Map<String, String> otherSettings = new HashMap<>();
+            Set<String> defaultSettings = Set.of("extensionName", "hostAddress", "hostPort", "opensearchAddress", "opensearchPort");
+            for (String settingKey : extensionMap.keySet()) {
+                if (!defaultSettings.contains(settingKey)) {
+                    otherSettings.put(settingKey, extensionMap.get(settingKey).toString());
+                }
+            }
```
saratvemulapalli, line 156: I am worried we are loading all pieces of config which we might not understand. This could potentially have security problems as we might load unknown code. 
An example vulnerability which was fixed in `yaml.load`[1]
WDYT?

[1] https://github.com/yaml/pyyaml/issues/243

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -179,11 +200,37 @@ protected ExtensionsRunner(Extension extension) throws IOException {
...
+            addSettingsToBuilder(settingsBuilder, SSL_TRANSPORT_SERVER_PEMCERT_FILEPATH, extensionSettings);
+            addSettingsToBuilder(settingsBuilder, SSL_TRANSPORT_SERVER_PEMTRUSTEDCAS_FILEPATH, extensionSettings);
+            addSettingsToBuilder(settingsBuilder, SSL_TRANSPORT_ENABLED_PROTOCOLS, extensionSettings);
+            addSettingsToBuilder(settingsBuilder, SSL_TRANSPORT_ENABLED_CIPHERS, extensionSettings);
+            addSettingsToBuilder(settingsBuilder, SSL_TRANSPORT_KEYSTORE_FILEPATH, extensionSettings);
+            addSettingsToBuilder(settingsBuilder, SSL_TRANSPORT_KEYSTORE_ALIAS, extensionSettings);
+            addSettingsToBuilder(settingsBuilder, SSL_TRANSPORT_KEYSTORE_TYPE, extensionSettings);
+            addSettingsToBuilder(settingsBuilder, SSL_TRANSPORT_TRUSTSTORE_ALIAS, extensionSettings);
+            addSettingsToBuilder(settingsBuilder, SSL_TRANSPORT_TRUSTSTORE_FILEPATH, extensionSettings);
+            addSettingsToBuilder(settingsBuilder, SSL_TRANSPORT_TRUSTSTORE_TYPE, extensionSettings);
```
saratvemulapalli, line 231: Probably have a utility function in `ExtensionSettings` and have a settingsBuilder?

src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
```diff
@@ -48,6 +48,7 @@ public ExtensionsInitRequestHandler(ExtensionsRunner extensionsRunner) {
     public InitializeExtensionResponse handleExtensionInitRequest(InitializeExtensionRequest extensionInitRequest) {
         logger.info("Registering Extension Request received from OpenSearch");
         extensionsRunner.opensearchNode = extensionInitRequest.getSourceNode();
+        extensionsRunner.getThreadPool().getThreadContext().putHeader("extension_unique_id", extensionInitRequest.getExtension().getId());
```
saratvemulapalli, line 51: Trying to understand, the header is plumbed in the threadcontext how does security plugin on OpenSearch see this?

src/main/java/org/opensearch/sdk/ssl/DefaultSslKeyStore.java
```diff
@@ -0,0 +1,834 @@
...
+import static org.opensearch.sdk.ssl.SecureSSLSettings.SSLSetting.SSL_TRANSPORT_TRUSTSTORE_PASSWORD;
+
+public class DefaultSslKeyStore implements SslKeyStore {
+
+    private static final String DEFAULT_STORE_TYPE = "JKS";
+
+    private void printJCEWarnings() {
+        try {
+            final int aesMaxKeyLength = Cipher.getMaxAllowedKeyLength("AES");
+
```
saratvemulapalli, line 93: I see most of the changes are from security repo[1]. It would be hard to maintain this in multiple places. 
Can these classes be packaged as a library and SDK can consume them?

[1] https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/ssl/DefaultSecurityKeyStore.java

saratvemulapalli (commented):
Thanks @cwperks , these changes are great. I did drop a few comments..

src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
cwperks: If I understand the logic in the security plugin correctly, a node is identified as being a node in the cluster iff one of 2 conditions are met:

1. The node's principal (extracted from the node cert) matches a node in the static nodes_dn list from `plugins.security.nodes_dn` or from the dynamic list in `nodes_dn.yml` and dynamic nodes enabled
2. The OID extracted from the certificates subjectAltName matches the `plugins.security.cert.oid` setting which defaults to `1.2.3.4.5.5` (Curious, why is this chosen as a default?)

In addition to one of the 2 methods up above, the security plugin also expects the `_opendistro_security_remotecn` thread context header is set. If this header on a transport request matches the cluster name from the cluster state then its identified as a transport request originating from another node in the cluster. If it doesn't match, then its a trusted cross cluster request. 

`extension_unique_id` here mimics the `_opendistro_security_remotecn` behavior.

src/main/java/org/opensearch/sdk/ssl/DefaultSslKeyStore.java
cwperks: That begs the question: where would be the best location to store it?

These files are not identical because I changed the constants to remove `plugins.security` and also removed any reference to `HTTP` since extension nodes are not HTTP servers. I agree it should be extracted, but to where?

Would that block this PR, what do you think about filing an issue to abstract the common code into a library that both repos can consume?

Docs/CERTIFICATE_GENERATION.md
cwperks: I included this in here for testing instructions to show how to test this change manually. Currently we have no tool for certificate generation, only documentation: https://opensearch.org/docs/latest/security/configuration/generate-certificates/

I don't think we should create a tool for certificate generation, but we should update the `install_demo_configuration.sh` tool to account for extensions. If we create a demo configuration with AD as an extension, then AD will need its own demo certs just like the node does in the demo configuration.

src/main/java/org/opensearch/sdk/ExtensionSettings.java
cwperks: That's a good idea, I will check it against a known list of settings keys. I did it this way for dev convenience since there are quite a few new settings introduced in this PR, but you're right that it shouldn't just accept anything as a setting key.

src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
```diff
@@ -65,7 +66,10 @@ public InitializeExtensionResponse handleExtensionInitRequest(InitializeExtensio
             extensionsRunner.setExtensionNode(extensionInitRequest.getExtension());
             // TODO: replace with sdkTransportService.getTransportService()
             TransportService extensionTransportService = extensionsRunner.getExtensionTransportService();
-            extensionTransportService.connectToNode(extensionsRunner.opensearchNode);
+            extensionTransportService.connectToNodeAsExtension(
```
cwperks, line 69: Without this change, it causes a fatal error to the opensearch node and the node fails with the error:

```
Likely root cause: OpenSearchException[Transport client authentication no longer supported.]
```

The fix for this is the companion PR in core that introduces `connectToNodeAsExtension`. Inside that function it injects `extension_unique_id` into the thread context of the connection validator. I'm still wrapping my head around this change, but it solves for the fatal error.

Docs/CERTIFICATE_GENERATION.md
```diff
@@ -0,0 +1,137 @@
...
+
+```
+#! /bin/bash
+
+openssl genrsa -out root-ca-key.pem 2048
+openssl req -new -x509 -sha256 -key root-ca-key.pem -subj "/C=US/ST=NEW YORK/L=BROOKLYN/O=OPENSEARCH/OU=SECURITY/CN=ROOT" -out root-ca.pem -days 730
+
+openssl genrsa -out extension-01-key-temp.pem 2048
+openssl pkcs8 -inform PEM -outform PEM -in extension-01-key-temp.pem -topk8 -nocrypt -v1 PBE-SHA1-3DES -out extension-01-key.pem
```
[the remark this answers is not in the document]
cwperks: If that's the case what do you think about opening a PR against the documentation website? https://opensearch.org/docs/latest/security/configuration/generate-certificates/

These commands are yanked from there and modified with different subjects

Docs/CERTIFICATE_GENERATION.md
```diff
@@ -0,0 +1,137 @@
...
+setting of `opensearch.yml`
+
+
+## Certificate Generation Script
+
+```
+#! /bin/bash
+
+openssl genrsa -out root-ca-key.pem 2048
+openssl req -new -x509 -sha256 -key root-ca-key.pem -subj "/C=US/ST=NEW YORK/L=BROOKLYN/O=OPENSEARCH/OU=SECURITY/CN=ROOT" -out root-ca.pem -days 730
```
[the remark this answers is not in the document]
cwperks: This is for testing manually so I think 730 days is fine, when an integration test framework is created then we should take the cert generation logic from the integration test framework of the security plugin.

src/main/java/org/opensearch/sdk/ExtensionSettings.java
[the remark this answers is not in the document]
cwperks: Changed to `securitySettings`

src/main/java/org/opensearch/sdk/ssl/util/SSLCertificateHelper.java
cwperks: Updated

Docs/CERTIFICATE_GENERATION.md
saratvemulapalli: That makes sense, thank you.

src/main/java/org/opensearch/sdk/ssl/DefaultSslKeyStore.java
saratvemulapalli: Sure definitely lets track this in an issue, I dont want to block our progress. 
This is operational pain but incremental progress is important :)

src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
cwperks: @saratvemulapalli I'm looking into this more today if this is what's blocking this PR. This was the only way I could figure out how to get the handshake working between a node and an extension with the security plugin installed. I will look more in detail to see how cross cluster works to see if this can more closely model cross cluster cases.

src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
saratvemulapalli: Thank you, that makes more sense.

src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
saratvemulapalli: Thanks @cwperks. It makes sense.

src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
saratvemulapalli: I've started making changes to move the identity into abstract piece of information. This should help us add more information for identity for extension and it encapsulates future changes.
See: https://github.com/opensearch-project/OpenSearch/pull/7302/files#diff-10ffb20227a1e7e024bc7a633c75b97ea7d2db8e59b6117de39e28b7b9e4c22a

cwperks: Thank you Sarat! wdyt about populating the threadcontext with the unique id from that protobuf when its received in core? Similar to how `_opendistro_security_remotecn` is expected to be in the threadcontext, or would there be a better place to keep track of where the transport request originated from that can be read by other plugins?

saratvemulapalli: Lets do it, Lets start with populating `uniqueId` from ExtensionIdentity.
I think this is the only way it would work, considering if Extension wants to talk to a plugin the threadcontext is expected.

saratvemulapalli:
Merged companion PR on OpenSearch, gradle check should pass.

cwperks:
@saratvemulapalli The gradle check passed now that the companion PR in core has been merged. Thank you!

saratvemulapalli:
@opensearch-project/opensearch-sdk-java looking for another set of eyes.

cwperks:
I will create an issue to move the classes in `org.opensearch.sdk.ssl` to a library somewhere that both the security plugin and the sdk can depend on and import.

src/main/java/org/opensearch/sdk/ssl/SSLConfigConstants.java
```diff
@@ -0,0 +1,217 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import org.opensearch.common.settings.Settings;
+
+/**
+ * Class containing constants for SSL
+ */
+public final class SSLConfigConstants {
```
peternied, line 21: I see that this and the keystore file contain a lot of functionality that is unused.  What do you think about removing unused functionality or at least making as much as possible package internal?  This might be a good candidate for a follow up issue too.

src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -76,6 +155,10 @@ public String getOpensearchPort() {
         return opensearchPort;
     }
 
+    public Map<String, String> getSecuritySettings() {
```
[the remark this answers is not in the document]
peternied: After reviewing everything else maybe this should be rolled into that follow up issue

peternied (approved):
I've left a number of comments that I think all bubble up to the notion of having a shared library for SSL. Great job getting this working end to end.

cwperks:
I created 2 follow-up issues around TLS to create integration tests and extract the common code with the security plugin into a library:

- https://github.com/opensearch-project/opensearch-sdk-java/issues/714
- https://github.com/opensearch-project/opensearch-sdk-java/issues/715

src/main/java/org/opensearch/sdk/ssl/SSLConfigConstants.java
cwperks: Most of the values in this class are used, but there are a lot of commented out Ciphers which is the same in the security plugin. The list of Ciphers used should always be matching between the 2 systems which underscores the need to extract these into a [library](https://github.com/opensearch-project/opensearch-sdk-java/issues/714). 

In `DefaultSslKeyStore` I commented out the function to `getOtherName` because it would have required the addition of a new dependency on the SDK and I was trying to avoid that. See https://github.com/opensearch-project/opensearch-sdk-java/pull/619#discussion_r1156034710.

I chose to keep the commented out code in to spark a conversation on here to see if its potentially something needed. This was added into the security plugin to support hot reload with certificates that contained `otherName`.

src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -30,6 +54,33 @@ public class ExtensionSettings {
     private String opensearchAddress;
     private String opensearchPort;
 
+    public static final Set<String> SECURITY_SETTINGS_KEYS = Set.of(
+        "path.home", // TODO Find the right place to put this setting
```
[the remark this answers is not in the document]
cwperks: I will create a follow-up issue. Similar to the opensearch `path.home` setting this would represent the directory where the extension is installed and shouldn't be within the security settings.

Docs/CERTIFICATE_GENERATION.md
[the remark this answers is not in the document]
owaiskazi19: https://github.com/opensearch-project/opensearch-sdk-java/pull/687 the framework is almost there

src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
owaiskazi19: @cwperks just trying to understand why `connectToNodeAsExtension` is required? Is it because we are injecting the unique_id in the thread context and since no authentication was before so we didn't face any problem with `connectToNode`?

src/main/java/org/opensearch/sdk/ssl/SSLNettyTransport.java
```diff
@@ -0,0 +1,328 @@
...
+        }
+
+        /**
+         *
+         * @param ch            the {@link Channel} which was registered.
+         * @throws Exception
+         */
+        @Override
+        protected void initChannel(Channel ch) throws Exception {
+            super.initChannel(ch);
```
owaiskazi19, line 145: I see security plugin has a dual mode here. Can you help me understand what's the significance of dual mode?

owaiskazi19 (commented):
Overall LGTM! Sorry for the delay. A quick question though. What if in the future we plan to move the security plugin to core, how will we handle the files/changes that we pulled in this PR directly from security plugin?

src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
cwperks: @owaiskazi19 You haven't faced this problem yet because you haven't been testing with the security plugin installed. The security plugin has logic that forbids transport requests from anything that isn't a node in the cluster and an extension is not a node in the cluster.

What I was doing with `connectToNodeAsExtension` and also with line 51 in this file is to identify all of these transport requests as coming from an extension and the extension is identified by its unique id. For some reason line 51 was not working with the handshake so I created `connectToNodeAsExtension` which also injects the unique id into the thread context of the connection validator. See the linked PR to the security plugin in the PR description, it checks if the experimental feature flag for extensions is enabled and then checks to see if a transport request comes from an extension so it can permit the request.

dbwiddis (dismissed):

src/main/java/org/opensearch/sdk/ssl/SSLNettyTransport.java
cwperks: Dual means that it supports encrypted and unencrypted traffic at the same time. To be honest, I have never used this feature of the security plugin before and it needs to be explicitly turned on. https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/ssl/transport/DualModeSSLHandler.java#L69-L75

cwperks:
@owaiskazi19 In either case with security as a plugin or security in core we would extract the same files that help setup TLS to a library that both can depend on.

saratvemulapalli (approved):
Thanks @cwperks for these changes!!
Getting this working E2E is great!

owaiskazi19 (approved):
LGTM! Thanks @cwperks for addressing the comments.

cwperks:
@owaiskazi19 I went through each comment marked unresolved and marked them each as resolved. There were a few items from Peter, namely around the placement of the `path.home` setting, that I plan to address in a future PR. I have opened up 2 issues around integration tests for this change and extracting common files to a library.

vibrantvarun:
Great work, @cwperks

cwperks:
Yes I will raise a manual PR for the 1.x backport. Understood about letting reviewers mark the PR as resolved, this PR was sitting for a while and I was trying to move it forward and use discretion with whether a comment was addressed. This should be the largest PR in size for a while since it added a lot of configurability around TLS, but future security related PRs will be much smaller comparatively.

cwperks:
@owaiskazi19 Opened up a manual backport PR: https://github.com/opensearch-project/opensearch-sdk-java/pull/718

PR #622, opened 2023-03-29
Map extension routes to permission-able action names

cwperks:
### Description

This PR relies on an an earlier PR (https://github.com/opensearch-project/opensearch-sdk-java/pull/619) to setup TLS for extensions. 

This PR also has a companion core and security PR

- Core - https://github.com/opensearch-project/OpenSearch/pull/6870
- Security - https://github.com/opensearch-project/security/pull/2753 ~~https://github.com/opensearch-project/security/pull/2601~~

Isolated changes can be viewed using the compare tool: https://github.com/cwperks/opensearch-sdk-java/compare/setup-extension-tls...ssl-and-handler-naming

This change modifies the extension settings by adding a new `shortName` setting (an abbreviation for an extension) that is used when creating a name for a route. `/hello` of the HelloWorld extension is mapped to the permission-able name `extensions:hw/greet`. The RouteHandler moves to core as part of this change, its needed by the security plugin to determine if a RestRequest is destined for an extension.

Endpoint registration has been updated to send the name of the route to core upon registration.

### Issues Resolved

Related to https://github.com/opensearch-project/security/issues/2589


owaiskazi19:
The changes overall LGTM @cwperks! I'm still trying to figure out the best place for these changes. Should it be in OpenSearch like we have [ExtensionRestRequest](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/extensions/rest/ExtensionRestRequest.java) there.

src/main/java/org/opensearch/sdk/ExtensionRouteHandlerFactory.java
```diff
@@ -0,0 +1,30 @@
...
+    private static ExtensionRouteHandlerFactory INSTANCE;
+
+    private String extensionShortName;
+
+    private ExtensionRouteHandlerFactory() { }
+
+    public static ExtensionRouteHandlerFactory getInstance() {
+        if(INSTANCE == null) {
+            INSTANCE = new ExtensionRouteHandlerFactory();
+        }
```
vibrantvarun, line 15: Are we trying to create SINGLETON object for INSTANCE here? if yes is it multithreaded safe?

src/main/java/org/opensearch/sdk/ExtensionRouteHandlerFactory.java
```diff
@@ -0,0 +1,30 @@
+package org.opensearch.sdk;
+
+import org.opensearch.OpenSearchException;
+
+public class ExtensionRouteHandlerFactory {
+    private static ExtensionRouteHandlerFactory INSTANCE;
```
owaiskazi19, line 6: Do you think we can use Guice here to achieve this?

src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -93,14 +93,23 @@ private ExtensionSettings() {
...
+     * @param shortName  The shortened name for the extension
      * @param hostAddress  The IP Address to bind this extension to.
      * @param hostPort  The port to bind this extension to.
      * @param opensearchAddress  The IP Address on which OpenSearch is running.
      * @param opensearchPort  The port on which OpenSearch is running.
      */
-    public ExtensionSettings(String extensionName, String hostAddress, String hostPort, String opensearchAddress, String opensearchPort) {
+    public ExtensionSettings(
+        String extensionName,
+        String shortName,
```
owaiskazi19, line 104:
```suggestion
        String shortExtensionName,
```

src/main/java/org/opensearch/sdk/rest/ExtensionRestPathRegistry.java
```diff
@@ -133,6 +133,24 @@ private void registerWithDeprecatedHandler(
...
+     * @param extensionRestHandler  The RestHandler to handle this route
+     */
+    public void registerHandler(Method method, String path, String name, ExtensionRestHandler extensionRestHandler) {
+        pathTrie.insertOrUpdate(
+            path,
+            new SDKMethodHandlers(path, extensionRestHandler, method),
+            (mHandlers, newMHandler) -> mHandlers.addMethods(extensionRestHandler, method)
+        );
+        String restPathWithName = restPathToString(method, path, name);
+        registeredPaths.add(restPathWithName);
```
owaiskazi19, line 151: We are missing adding paths to deprecated paths here. One way to handle this would be to merge both the `registerHandler` methods and add a check for `name`.

src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -120,20 +130,25 @@ public ExtensionSettings(String extensionName, String hostAddress, String hostPo
...
         this.securitySettings = securitySettings;
     }
 
     public String getExtensionName() {
         return extensionName;
     }
 
+    public String getShortName() {
+        return shortName;
+    }
```
owaiskazi19, line 150: This getter should have a test in `TestExtensionSettings.java`

src/main/java/org/opensearch/sdk/ExtensionRouteHandlerFactory.java
cwperks: Yes, the problem I was facing is that I needed the extensions shortName (abbreviation) inside the RestHandler classes like `RestHelloAction` in the sample extension. Instead of passing the value as a constructor arg or passing that extension runner into those classes, I thought it would be better to have singular place where its registered and can be used throughout.

Do you know how I can test if this is multi-threaded safe?

src/main/java/org/opensearch/sdk/ExtensionRouteHandlerFactory.java
cwperks: That would be great. I'll look into it. That main problem I was facing was that I needed to have the shortName of the extension available in the REST Handler (i.e. inside of RestHelloAction for the sample extension). The way I went about doing that was to create a singleton class that could be fetched throughout the SDK but if there is a better way to do it then I will switch to using that.

src/main/java/org/opensearch/sdk/ExtensionSettings.java
cwperks: I'll update it, I was torn between different names for this variable. One of the naming ideas was `extensionAbbr`, but I didn't know if was appropriate to use an abbreviation in a variable name signifying that this variable is an abbreviation. It felt like it was a crossword puzzle clue.

src/main/java/org/opensearch/sdk/ExtensionSettings.java
cwperks: This has been added to a few different tests to verify its behavior

src/main/java/org/opensearch/sdk/ExtensionSettings.java
cwperks: Updated

cwperks:
Can the maintainers take another look at this PR?

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -194,6 +197,14 @@ protected ExtensionsRunner(Extension extension) throws IOException {
         logger.info("SSL is " + sslText + " for transport");
         this.settings = settingsBuilder.build();
 
+        if (extensionSettings.getShortExtensionName() != null) {
+            // initialize ExtensionRouteHandlerFactory
+            ExtensionRouteHandlerFactory factory = ExtensionRouteHandlerFactory.getInstance();
+            if (!factory.isInitialized()) {
```
dbwiddis, line 203: I'm trying to think of a situation where this would ever be initialized before the ExtensionsRunner constructor was called.  Possible but someone would have to go out of their way to do it.   

Not sure we need a factory at all, and this is just one more reason this solution is overly complex.

src/main/java/org/opensearch/sdk/rest/ExtensionRestPathRegistry.java
```diff
@@ -57,20 +65,25 @@ public void registerHandler(ExtensionRestHandler restHandler) {
...
         if (extensionRestHandler instanceof ExtensionDeprecationRestHandler) {
-            registeredDeprecatedPaths.add(restPathToString(method, path));
+            registeredDeprecatedPaths.add(restPathToString(method, path, name));
             registeredDeprecatedPaths.add(((ExtensionDeprecationRestHandler) extensionRestHandler).getDeprecationMessage());
         } else {
-            registeredPaths.add(restPathToString(method, path));
+            if (name.isPresent()) {
+                registeredPaths.add(restPathToString(method, path, name));
+            } else {
+                registeredPaths.add(restPathToString(method, path, name));
```
dbwiddis, line 85: The method called in `else` is exactly the same as the method called in the `isPresent()` case. Why does this conditional exist?

src/main/java/org/opensearch/sdk/rest/ExtensionRestPathRegistry.java
```diff
@@ -171,9 +189,13 @@ public List<String> getRegisteredDeprecatedPaths() {
      *
      * @param method  the method.
      * @param path  the path.
+     * @param name  the name corresponding to this route.
      * @return A string appending the method and path.
      */
-    public static String restPathToString(Method method, String path) {
+    public static String restPathToString(Method method, String path, Optional<String> name) {
```
dbwiddis, line 195: Originally this was a more important method but I think it's been reduced to a cosmetic/logging/exception-message method as the rest paths are stored in the `pathTrie`.  So I think the plain space-delimited text may not be useful or may be more confusing than using some other custom output here.

src/main/java/org/opensearch/sdk/sample/helloworld/rest/RestHelloAction.java
```diff
@@ -50,13 +52,18 @@ public class RestHelloAction extends BaseExtensionRestHandler {
...
+    public RestHelloAction() {}
+
     @Override
     public List<RouteHandler> routeHandlers() {
         return List.of(
-            new RouteHandler(GET, "/hello", handleGetRequest),
-            new RouteHandler(POST, "/hello", handlePostRequest),
-            new RouteHandler(PUT, "/hello/{name}", handlePutRequest),
-            new RouteHandler(DELETE, "/goodbye", handleDeleteRequest)
+            new ExtensionRouteHandler("greet", GET, "/hello", handleGetRequest),
```
dbwiddis, line 63: This appears to be the only case where we are actually using the new `ExtensionRouteHandler` class.

This method signature is an override that only exists if we are extending `BaseExtensionRestHandler` so it seems to me we could add all this logic in that superclass.   The `BaseExtensionRestHandler` class could fetch the shortname from the settings and then just expose a `protected routePrefix(String name)` method that returns "shortname:name".  Then this line would just be `new RouteHandler(routePrefix("greet"), GET, ...)` and you'd save a ton of intermediate objects and initialization checking, etc.

dbwiddis (changes requested):
I think this is overly complex. Details below.

src/main/java/org/opensearch/sdk/ExtensionRouteHandlerFactory.java
dbwiddis: >  if there is a better way to do it then I will switch to using that.

Extensions are always initialized with their settings. `getExtensionSettings()` must be implemented on the `Extension` interface.  (The `BaseExtension` class offers the convenience of either a settings object or yml file.)

Any method on the main`Extension` class ought to be able to call `getExtensionSettings().getShortExtensionName()` to access this value without any factory/singletons.

In the AD Extension we usually pass the `ExtensionsRunner` class to every REST handler, and we have a getter on the runner for the extension.  So all the AD REST handlers can do `runner.getExtension().getExtensionSettings().getShortExtensionName()`.

src/main/java/org/opensearch/sdk/rest/ExtensionRestPathRegistry.java
cwperks: Removed this conditional

cwperks:
@dbwiddis I removed the ExtensionRouteHandler and ExtensionRouteHandlerFactory in favor of your suggestion of including the `routePrefix` static method in the `BaseExtensionRestHandler`.

cwperks:
Closing in favor of https://github.com/opensearch-project/opensearch-sdk-java/pull/827

PR #623, opened 2023-03-29
Generate RestRequest to pass to Rest Handlers

dbwiddis:
Companion PRs (this one must be merged after Core and before AD):
 - Core: https://github.com/opensearch-project/OpenSearch/pull/6874
 - AD: https://github.com/opensearch-project/anomaly-detection/pull/848

### Description

Extensions do not use all the internal components of a `RestRequest` and initial implementation double-purposed the `ExtensionsRestRequest` object used in transport as the object that Extensions would need to handle.  However, increasing use of the methods on this object are ending up duplicating a lot of code.  It makes a lot more sense to just recreate a `RestRequest` object from the parameters passed in this request.

### Issues Resolved

Part of [SDK #431](https://github.com/opensearch-project/opensearch-sdk-java/issues/431).


src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
```diff
@@ -21,6 +21,7 @@
 import org.opensearch.extensions.rest.ExtensionRestRequest;
```
owaiskazi19, line 21: Do we need this import/file anymore?

dbwiddis: Nope, it was in a stray javadoc so didn't get auto-cleaned by spotless like everything else.

owaiskazi19: What about `ExtensionRestRequest`file? Since we are using `RestRequest` should we get delete it as a part of this PR?

dbwiddis: No, we still need that as part of transport, we're just removing the "dual use" nature of it.

dbwiddis: We still need some parts of it, particularly the parsed `params()` that were generated by `RestController` using the registered rest action named params.

owaiskazi19: Shouldn't we document the usage of both the requests then? When to use `ExtensionRestRequest` and `RestRequest` in the migration doc?

dbwiddis: For migration, using "RestRequest" isn't a change at all, it will minimize the diff. And the ExtensionRestRequest is documented [here](https://github.com/opensearch-project/opensearch-sdk-java/blob/main/PLUGIN_MIGRATION.md#replace-bytesrestresponse-with-extensionrestresponse).  (The lines above it are being removed as part of this PR.)

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -185,8 +185,10 @@ protected ExtensionsRunner(Extension extension) throws IOException {
         this.customNamedXContent = extension.getNamedXContent();
         // save custom namedWriteable
         this.customNamedWriteables = extension.getNamedWriteables();
-        // initialize NamedXContent Registry. Must happen after getting extension namedXContent
+        // initialize NamedXContent Registry.
         this.sdkNamedXContentRegistry = new SDKNamedXContentRegistry(this);
+        // initialize RestRequest Handler. Must happen after instantiating SDKNmaedXContentRegistry
```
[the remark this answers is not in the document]
ryanbogan: LGTM other than this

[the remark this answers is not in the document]
dbwiddis: Typo? What typo? ;)

joshpalis (approved):

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
owaiskazi19: Anything documented about `ExtensionRestRequest` or plugins won't need it at all for the migration and it will be required only by SDK moving forward?

dbwiddis: > Anything documented about `ExtensionRestRequest` or plugins won't need it at all for the migration and it will be required only by SDK moving forward?

That is correct.  The sequence:
 - user submits a HttpRequest
 - Rest Controller parses this into a RestRequest including named params (tricky!)
 - we use ExtensionRestRequest to extract the parts we need out of RestRequest and send to SDK
 - SDK reassembles a new RestRequest which is mostly like the old one and is further used by extensions without any change to existing code
 - The only tricky thing needed is that users can't just send back a BytesRestResponse like usual; that is wrapped in the ExtensionRestResponse in order to track the "consumed parameters" of the request.

saratvemulapalli (dismissed):
The change makes sense. 
I wonder how much of these dependencies we will inherit from OpenSearch.

owaiskazi19 (dismissed):
Merging this in! We can continue the discussion in the comments/thread

dbwiddis:
> Merging this in! We can continue the discussion in the comments/thread

#605 is still in progress and we can address any lingering questions there as well.

owaiskazi19:
```
> Task :compileJava

> Task :processResources
> Task :classes
/home/runner/work/opensearch-sdk-java/opensearch-sdk-java/src/test/java/org/opensearch/sdk/TestBaseExtensionRestHandler.java:152: error: cannot find symbol
                    uri.append(uri.isEmpty() ? '?' : '&').append(param.getKey()).append('=').append(param.getValue());

dbwiddis:
>                     uri.append(uri.isEmpty() ? '?' : '&').append(param.getKey()).append('=').append(param.getValue());

Turns out `isEmpty()` is JDK15+, which is why it passed locally.  Going back to the old style `.length() == 0`.

Fixed and rebased to main.

owaiskazi19 (approved):

joshpalis (approved):

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
dbwiddis: > 1. In the future can we eliminate the ExtensionRestRequest completely and use RestRequest since we will be having security for the headers too? This will help us to ease the pain of creating another RestRequest in SDK

We need some method of passing the method, uri, content type headers, content bytes, and params over transport.  You can call it whatever you want, and maybe even change to protobuf, but those bits need to get from point A to point B somehow.

> 2. Again, in the future can we just send back BytesRestResponse and parse it in SDK to eliminate `ExtensionRestResponse`?

You need some way of communicating "consumed params" along with the bytes or, if you want to just ignore the whole consumed params thing, you could.

dbwiddis:
> The backport to `1.x` failed:

I'll handle this

dbwiddis:
Probably need backport https://github.com/opensearch-project/opensearch-sdk-java/pull/628 merged before trying to backport this.

Issue #624, opened 2023-03-29
[FEATURE] Add getExecutorBuilders extension point to SDK BaseExtension interface


joshpalis:
### Is your feature request related to a problem?

In the `AnomalyDetectorJobRunner`, when the job runner's `runJob()` method is invoked by Job Scheduler, AD creates a `Runnable` [here](https://github.com/opensearch-project/anomaly-detection/blob/fd2fd5d5f63c9004b9f265e3c1e4475282447da8/src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java#L148) that executes the job when invoked. The runnable is then submitted to this threadpool executor [here](https://github.com/opensearch-project/anomaly-detection/blob/fd2fd5d5f63c9004b9f265e3c1e4475282447da8/src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java#L202) for the threadpool named `ad-threadpool`.

The executor builder for this particular threadpool is instantiated [here](https://github.com/opensearch-project/anomaly-detection/blob/fd2fd5d5f63c9004b9f265e3c1e4475282447da8/src/main/java/org/opensearch/ad/AnomalyDetectorPlugin.java#L833) within the `getExecutorBuilders` extension point. 

Within node.java, the executor builders for each plugin is retrieved [here](https://github.com/opensearch-project/OpenSearch/blob/5eed61e263d9bdfbd64b2b677d4ef6933085dac5/server/src/main/java/org/opensearch/node/Node.java#LL479C48-L479C48), which is then used to instantiate the threadpool [here](https://github.com/opensearch-project/OpenSearch/blob/5eed61e263d9bdfbd64b2b677d4ef6933085dac5/server/src/main/java/org/opensearch/node/Node.java#LL482C15-L482C15).

### What solution would you like?
The SDK's `BaseExtension` interface should have a method called `getExecutorBuilders` in which we return a list of ExecutorBuilders from an extension. These executor builders should then be retrieved from the extension during `ExtensionsRunner` initialization and used to instantiate the threadpool [here](https://github.com/opensearch-project/opensearch-sdk-java/blob/e13aed8eaaf17f448a9783b8fcb6a3f20ef6d5bb/src/main/java/org/opensearch/sdk/ExtensionsRunner.java#L179).

Issue #626, opened 2023-03-29
Migrate AnomalyResultAction/TransportAction to SDK using the SDKClient


joshpalis:
### Is your feature request related to a problem?
Part of https://github.com/opensearch-project/opensearch-sdk-java/issues/383

After the `AnomalyDetectorJobRunner` is executed by Job Scheduler, AD creates a `AnomalyResultRequest` and invokes the `AnomalyResultAction` [here](https://github.com/opensearch-project/anomaly-detection/blob/fd2fd5d5f63c9004b9f265e3c1e4475282447da8/src/main/java/org/opensearch/ad/AnomalyDetectorJobRunner.java#L307). 

To complete the start detector workflow, it is necessary to :

- [x] Add the `AnomalyResultAction` Action Handler to the `getActions` extension point
- [x] Add the `RCFResultAction` Action handler to the `getActions` extension point
- [x] Add the `EntityResultAction` Action handler to the `getActions` extension point
- [x] Add the `ADResultBulkAction` Action handler to the `getActions` extension point
- [x] Integrate `AnomalyResultTransportAction` with the SDKClient/ClusterService
- [x] Integrate `EntityResultAction` with the SDKClient/ClusterService (multi-entity detector)
- [x] Integrate `RCFResultAction/TransportAction` with the SDKClient/ClusterService (single-entity detector)
- [x] Integrate `ProfileTransportAction` with SDKClient/ClusterService (used to update realtime high cardinality detector)
- Integrate the following components in `createComponents` extension point needed for the `AnomalyResultTransportAction`, `EntityResultTransportAction` :
   - [x] NodeStateManager
   - [x] FeatureManager
   - [x] ModelManager
   - [x] ADCircutBreakerService
   - [x] ADStats
   - [x] ADTaskManager
   - [x] IndexNameExpressionResolver
   - [x] EntityColdStartWorker
   - [x] MultiEntityResultHandler
   - [x] ResultWriteWorker
   - [x] CheckPointReadWorker
   - [x] ColdEntityWorker
- [x] Fix any affected test classes

## Blockers for multi-entity detector results
- [x] Implement IndexingPressure as a created component (https://github.com/opensearch-project/opensearch-sdk-java/issues/655)
[...]

PR #627, opened 2023-03-30
Adds getExecutorBuilders and indexNameExpressionResolver to extensionsRunner

joshpalis:
### Description
implements the `getExecutorBuilders` extension point within `ExtensionsRunner` and also adds the `indexNameExpressionResolver` to the `extensionRunner` to be used in `createComponents`

The `executorBuilders` extension point will be implemented as part of [start detector](https://github.com/opensearch-project/opensearch-sdk-java/issues/383).

The `indexNameExpressionResolver` will be utilized as part of the [AnomalyResultAction/TransportAction](https://github.com/opensearch-project/opensearch-sdk-java/issues/626)

### Issues Resolved
https://github.com/opensearch-project/opensearch-sdk-java/issues/624

### DIFF
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -154,7 +154,6 @@ public class ExtensionsRunner {
         new ExtensionsIndicesModuleNameRequestHandler();
     private final ExtensionsRestRequestHandler extensionsRestRequestHandler = new ExtensionsRestRequestHandler(extensionRestPathRegistry);
     private final ExtensionActionRequestHandler extensionsActionRequestHandler;
-    private final List<ExecutorBuilder<?>> executorBuilders;
     private final AtomicReference<RunnableTaskExecutionListener> runnableTaskListener;
     private final IndexNameExpressionResolver indexNameExpressionResolver;
 
@@ -184,9 +183,9 @@ public class ExtensionsRunner {
             .put(TransportSettings.PORT.getKey(), extensionSettings.getHostPort())
             .build();
 
-        this.executorBuilders = extension.getExecutorBuilders(settings);
+        final List<ExecutorBuilder<?>> executorBuilders = extension.getExecutorBuilders(settings);
 
-        runnableTaskListener = new AtomicReference<>();
+        this.runnableTaskListener = new AtomicReference<>();
         this.threadPool = new ThreadPool(settings, runnableTaskListener, executorBuilders.toArray(new ExecutorBuilder[0]));
         this.indexNameExpressionResolver = new IndexNameExpressionResolver(this.threadPool.getThreadContext());
         this.taskManager = new TaskManager(settings, threadPool, Collections.emptySet());
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -154,6 +154,7 @@ public class ExtensionsRunner {
         new ExtensionsIndicesModuleNameRequestHandler();
     private final ExtensionsRestRequestHandler extensionsRestRequestHandler = new ExtensionsRestRequestHandler(extensionRestPathRegistry);
     private final ExtensionActionRequestHandler extensionsActionRequestHandler;
+    private final List<ExecutorBuilder<?>> executorBuilders;
     private final AtomicReference<RunnableTaskExecutionListener> runnableTaskListener;
     private final IndexNameExpressionResolver indexNameExpressionResolver;
 
@@ -183,7 +184,7 @@ public class ExtensionsRunner {
             .put(TransportSettings.PORT.getKey(), extensionSettings.getHostPort())
             .build();
 
-        final List<ExecutorBuilder<?>> executorBuilders = extension.getExecutorBuilders(settings);
+        this.executorBuilders = extension.getExecutorBuilders(settings);
 
         runnableTaskListener = new AtomicReference<>();
         this.threadPool = new ThreadPool(settings, runnableTaskListener, executorBuilders.toArray(new ExecutorBuilder[0]));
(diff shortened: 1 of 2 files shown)


owaiskazi19:
Let's add tests for this new extension point

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -176,7 +182,12 @@ protected ExtensionsRunner(Extension extension) throws IOException {
             .put(TransportSettings.BIND_HOST.getKey(), extensionSettings.getHostAddress())
             .put(TransportSettings.PORT.getKey(), extensionSettings.getHostPort())
             .build();
-        this.threadPool = new ThreadPool(settings);
+
+        final List<ExecutorBuilder<?>> executorBuilders = extension.getExecutorBuilders(settings);
+
+        runnableTaskListener = new AtomicReference<>();
+        this.threadPool = new ThreadPool(settings, runnableTaskListener, executorBuilders.toArray(new ExecutorBuilder[0]));
```
owaiskazi19, line 189: Instead of passing it like `new ExecutorBuilder[0]`, create a global variable above and pass it here

owaiskazi19: What exactly `runnableTaskListener` is doing? I see it is just initialized and passed

joshpalis: I have just emulated what is done in Node.java [here](https://github.com/opensearch-project/OpenSearch/blob/5eed61e263d9bdfbd64b2b677d4ef6933085dac5/server/src/main/java/org/opensearch/node/Node.java#L482) to add the executor builders to the thread pool

joshpalis:
@owaiskazi19 The `getExecutorBuilders` extension point is not new, it has been added already, this PR just utilizes the extension point within the ExtensionRunner. See some test cases [here](https://github.com/joshpalis/opensearch-sdk-java/blob/e13aed8eaaf17f448a9783b8fcb6a3f20ef6d5bb/src/test/java/org/opensearch/sdk/TestExtensionInterfaces.java#L46). But I will add tests for the getters that I have added

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
dbwiddis: > Instead of passing it like `new ExecutorBuilder[0]`, create a global variable above and pass it here

The format @joshpalis has is standard for creating an array from a list.  (JDK 8+ has a fancier format availbale, `ExecutorBuilder[]::new` but it's essentially the same.)
 - The list is required by the implemented extension point
 - The array is required by the varags `ThreadPool` constructor.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -176,7 +182,12 @@ protected ExtensionsRunner(Extension extension) throws IOException {
             .put(TransportSettings.BIND_HOST.getKey(), extensionSettings.getHostAddress())
             .put(TransportSettings.PORT.getKey(), extensionSettings.getHostPort())
             .build();
-        this.threadPool = new ThreadPool(settings);
+
+        final List<ExecutorBuilder<?>> executorBuilders = extension.getExecutorBuilders(settings);
+
+        runnableTaskListener = new AtomicReference<>();
```
dbwiddis, line 188: (Style Preference) I really like to use `this.foo` in Constructors when setting final fields.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
```diff
@@ -203,6 +214,7 @@ protected ExtensionsRunner(Extension extension) throws IOException {
         modules.add(b -> {
             b.bind(ExtensionsRunner.class).toInstance(this);
             b.bind(Extension.class).toInstance(extension);
+            b.bind(IndexNameExpressionResolver.class).toInstance(indexNameExpressionResolver);
```
dbwiddis, line 217: (Style Preference) I like to group things together with breaks in between them to somewhat match the same order as they are declared/assigned above. It makes maintaining code easier when you can scan down the list and insert things at the same point.  This looks odd grouped together with the two top level classes.  Even if it's in the right order it should be perhaps after that whitespace break on the line below.

dbwiddis (approved):
LGTM. Some purely personal preference style suggestions.

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
dbwiddis: > What exactly `runnableTaskListener` is doing? I see it is just initialized and passed

Looks like it's a reference to a [TaskResourceTrackingService](https://github.com/opensearch-project/OpenSearch/blob/37b37c5b518c5bdb2dfdc5250117cfd39394edeb/server/src/main/java/org/opensearch/node/Node.java#L1224) object which in turn looks like it tracks thread memory allocations.

dbwiddis: > Instead of passing it like `new ExecutorBuilder[0]`, create a global variable above and pass it here

@owaiskazi19 and @joshpalis you may be interested in why we use `[0]` here: https://shipilev.net/blog/2016/arrays-wisdom-ancients/

ryanbogan (approved):

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
dbwiddis: @joshpalis I think you misunderstood the comment.  The executorBuilders is fine as a local variable. I meant to add this to the runnableTaskListener.

owaiskazi19 (approved):
LGTM! Thanks @joshpalis

src/main/java/org/opensearch/sdk/ExtensionsRunner.java
joshpalis: Oh I misread, let me go ahead and fix this :)

ryanbogan (approved):

owaiskazi19 (approved):

PR #637, opened 2023-04-01
Adding the ExtensionActionUtil class

### LINKED ISSUE #607: [FEATURE] Add util class for parsing ExtensionAction bytes

[Don't Repeat Yourself](https://en.wikipedia.org/wiki/Don't_repeat_yourself)

While reviewing [JS #349](https://github.com/opensearch-project/job-scheduler/pull/349) I noticed a lot of repetition of some of the same code as in #588.

This is all general code for the common pattern we've chosen of appending a request class name as a string to its parameters in bytes.

Copy these repeated features to a class in the same package as the ExtensionAction / ExtensionTransportAction / ExtensionActionRequest and related classes.  They might fit well in an existing class (ExtensionAction possibly) or in a new class (ExtensionActionUtil).  These include:
 - the UNIT_SEPARATOR byte constant
 - a method for finding the position of that separator in the byte array
 - methods for splitting the class string and param bytes
 - perhaps even a method to instantiate the class with the params

Basically, all the functions needed to do this:
https://github.com/opensearch-project/opensearch-sdk-java/blob/ebc684ae223e87f1e05712e086fc1dab911a3b5f/src/main/java/org/opensearch/sdk/handlers/ExtensionActionRequestHandler.java#L85-L96

Repeating ourselves.

This somewhat doubles down on using Writeable and if we're considering switching to protobuf, then we may consider a different approach.

Kuanysh-kst:
### Description
In this PR, I add methods to the ExtensionActionUtil class to reduce code repetition

### Issues Resolved
#607 


src/main/java/org/opensearch/sdk/action/ExtensionActionUtil.java
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * ExtensionActionUtil - a class for creating and processing remote requests using byte arrays.
+ */
+public class ExtensionActionUtil {
+    private static final Logger logger = LogManager.getLogger(ExtensionActionUtil.class);
+
+    /**
+     * The Unicode UNIT SEPARATOR used to separate the Request class name and parameter bytes
+     */
+    private static final byte UNIT_SEPARATOR = (byte) '\u001F';
```
dbwiddis, line 32: Let's make this public. It's a useful/standard unicode constant.

src/main/java/org/opensearch/sdk/action/ExtensionActionUtil.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * @param  requestBytes is a byte array containing the request data, used by the "createActionRequest"
+     * method to create an "ActionRequest" object, which represents the request model to be processed on the server.
+     * @return an "Action Request" object, which represents the request model for processing on the server,
+     * and is created using the request data stored in the byte array provided in the "requestBytes" parameter.
+     */
+    public static ActionRequest createActionRequest(ExtensionActionRequest requestBytes) {
+        int nullPos = indexOf(requestBytes.getRequestBytes());
```
dbwiddis, line 57: When I first wrote this I used a 0 (null) byte and named it accordingly.  After we changed it to a non-null delimiter it seems to have stuck.  Can we name this something like `delimPos`?

src/main/java/org/opensearch/sdk/action/ExtensionActionUtil.java
```diff
@@ -0,0 +1,82 @@
...
+                Arrays.copyOfRange(requestBytes.getRequestBytes(), nullPos + 1, requestBytes.getRequestBytes().length)
+            );
+            actionRequest = (ActionRequest) constructor.newInstance(requestByteStream);
+        } catch (Exception e) {
+            logger.debug("No request class [" + requestClassName + "] is available: " + e.getMessage());
+        }
+        return actionRequest;
+    }
+
+    private static int indexOf(byte[] bytes) {
```
dbwiddis, line 74: The usual `indexOf()` methods in Java tend to take the delimiter so it logically reads "index of the delimiter".  So it's not the best name here... how about `delimPos()` or `findDelimPos()`?

dbwiddis (commented):
Code LGTM so far, with a few suggestions.

I think there's an opportunity for more methods here, though.  

For example, the reverse of `createActionRequest` for example, where you give it a subclass `ActionRequest` and it returns the `ExtensionActionRequest`.  There may be more duplicates.

dbwiddis:
One more comment, @Kuanysh-kst ... looking at the commits on this PR, they go back to your original MapperExtension submission.  You've kept up to date with merge commits, but it's much cleaner if when you start work on a new feature, you go to your main branch, catch it up (`git pull upstream main`) and then re-branch from there to start work on a new feature.

Kuanysh-kst:
I tried to find where to add a label , I looked near the Label , but there's nothing , maybe I don't have enough rights to add labels

dbwiddis:
One more comment. When you create the class on OpenSearch side you'll also need a test class for it!

Kuanysh-kst:
Could it be that I'm getting an error due to different versions of the Constructor class in the sdk and Opensearch 
SDK:
`   
 @CallerSensitive
    public Constructor<T> getConstructor(Class<?>... parameterTypes) throws NoSuchMethodException, SecurityException {
        SecurityManager sm = System.getSecurityManager();
        if (sm != null) {
            this.checkMemberAccess(sm, 0, Reflection.getCallerClass(), true);
        }

        return getReflectionFactory().copyConstructor(this.getConstructor0(parameterTypes, 0));
    }
`
OpenSearch:
`    @CallerSensitive
    public Constructor<T> getConstructor(Class<?>... parameterTypes)
        throws NoSuchMethodException, SecurityException
    {
        @SuppressWarnings("removal")
        SecurityManager sm = System.getSecurityManager();
        if (sm != null) {
            checkMemberAccess(sm, Member.PUBLIC, Reflection.getCallerClass(), true);
        }
        return getReflectionFactory().copyConstructor(
            getConstructor0(parameterTypes, Member.PUBLIC));
    }
`

Kuanysh-kst:
Could it be that I'm getting an error due to different versions of the Constructor class in the sdk and Opensearch

dbwiddis:
> Could it be that I'm getting an error due to different versions of the Constructor class in the sdk and Opensearch

I'm missing the context of this question, can you ask it on the PR with the issue?   The "Constructor" class itself is part of Java.  The class you are recreating from its name needs to be on the classpath on both sides.... so if it's an SDK path you can't construct it on OpenSearch.

PR #643, opened 2023-04-03
Added support for headers for ExtensionRestRequest

### LINKED ISSUE #431: [FEATURE] Add support for headers for ExtensionRestRequest

nassipkali:
### Description
Adds SDKRestRequest, SDKHttpRequest

Companion PR: https://github.com/opensearch-project/OpenSearch/pull/6826

### Issues Resolved
#431


dbwiddis:
> @dbwiddis first option mentioned in previous pr is more complicated because protected constructor of RestRequest requires also to fill HttpRequest. We can create sdkHttpRequest instead sdkRestRequest. sdkHttpRequest can replace anonymous HttpRequest class.

We still need both.   In particular we really need the `params()` in the `RestRequest()` and the only way to get them in there from the static HttpRequest is by hacking at the uri (which I did in the test class).  Take a look at `TestBaseExtensionRestHandler.createTestRestRequest()` in particular the `uri()` method.  That works for combining `path` + `params` into a uri, but that's not the original uri.

But we should be able to use an anonymous `RestRequest` subclass instead of the static method to create the `RestRequest`.  Change `RestRequest.request(null, new HttpRequest() { ... }, null);` to `new RestRequest(xContentRegistry, params, path, headers, new HttpRequest() { ... } , null);`.

nassipkali:
> > @dbwiddis first option mentioned in previous pr is more complicated because protected constructor of RestRequest requires also to fill HttpRequest. We can create sdkHttpRequest instead sdkRestRequest. sdkHttpRequest can replace anonymous HttpRequest class.
> 
> We still need both. In particular we really need the `params()` in the `RestRequest()` and the only way to get them in there from the static HttpRequest is by hacking at the uri (which I did in the test class). Take a look at `TestBaseExtensionRestHandler.createTestRestRequest()` in particular the `uri()` method. That works for combining `path` + `params` into a uri, but that's not the original uri.
> 
> But we should be able to use an anonymous `RestRequest` subclass instead of the static method to create the `RestRequest`. Change `RestRequest.request(xContentRegistry, new HttpRequest() { ... }, null);` to `new RestRequest(xContentRegistry, params, path, headers, new HttpRequest() { ... existing overrides ... } , null) {};`.

https://github.com/opensearch-project/OpenSearch/blob/bcbb561b51c60776257e68786fcc411eebf4fab4/server/src/main/java/org/opensearch/rest/RestRequest.java#LL170C42-L170C42
Constuctor gets params from uri. Is RestRequest subclass needed in future? If yes then I try implement that class

nassipkali:
@dbwiddis I can add path to ExtensionRestRequest. If I keep path, uri, params in ExtensionRestRequest then I can create instance of RestRequest subclass without overhead(converting uri to path, getting params from uri)

dbwiddis:
> @dbwiddis I can add path to ExtensionRestRequest. If I keep path, uri, params in ExtensionRestRequest then I can create instance of RestRequest subclass without overhead(converting uri to path, getting params from uri)

The "params from uri" happens in OpenSearch inside the RestController.  It has already:
 - Decoded params from the URI ( such as the name/blah pair in `/restcommand/foo?name=blah`)
 - Decoded named params from registered REST actions (such as the id/123 pair in a registered rest path `/detector/{id}` when a user types `/detector/123`.

It's that second part that we do not have the ability to decode in the Extension, which is why we must pass the `params()` map in the `ExtensionRestRequest` and use it to recreate the `RestRequest` with those same params.   What I did in the test class was a hack around the way we should do it.

Since we have the params, we don't _really_ need the uri, since the path is the interesting part.  But since we are recreating the whole HttpRequest it doesn't hurt to include both (by putting the uri where it belongs, the user can get either the path or the uri).  It looks like you've handled these in your PR, so I'll go review it now!

src/main/java/org/opensearch/sdk/SDKHttpRequest.java
```diff
@@ -0,0 +1,144 @@
...
+ */
+public class SDKHttpRequest implements HttpRequest {
+    private final RestRequest.Method method;
+    private final String uri;
+    private final BytesReference content;
+    private final Map<String, List<String>> headers;
+    private final HttpVersion httpVersion;
+
+    /**
+    * Instantiates this class with a copy of request
```
dbwiddis, line 33: Can we be a bit more clear that this is a copy of an `ExtensionRestRequest`? You might even use `@link` either here or in the param tag.

src/main/java/org/opensearch/sdk/SDKHttpRequest.java
```diff
@@ -0,0 +1,144 @@
...
+        this.uri = request.uri();
+        this.content = request.content();
+        this.headers = request.headers();
+        this.httpVersion = request.httpVersion();
+    }
+
+    /**
+    * This method returns request method
+    * @return A method of request
+    */
```
dbwiddis, line 48: You do not need to include any javadocs when there is an `@Override` annotation, as it inherits the documentation from the superclass.  

Delete this javadoc, and all the ones which are overrides that actually return data.  If you want to include a javadoc for the overrides that return null, you should indicate that instead of pretending it does something. 

Something like this, perhaps.
```java
/**
 * Not implemented.
 * @return null
 */
```
or for the `releaseAndCopy()` methods:
```java
/**
 * Not implemented. Does nothing.
 * @return null
 */
```

src/main/java/org/opensearch/sdk/handlers/ExtensionsRestRequestHandler.java
```diff
@@ -77,73 +69,18 @@ public RestExecuteOnExtensionResponse handleRestExecuteOnExtensionRequest(Extens
...
-                return null;
-            }
-        }, null);
+        SDKHttpRequest sdkHttpRequest = new SDKHttpRequest(request);
+        SDKRestRequest sdkRestRequest = new SDKRestRequest(
+            sdkNamedXContentRegistry.getRegistry(),
+            request.params(),
+            request.path(),
+            request.headers(),
+            sdkHttpRequest,
```
dbwiddis, line 78: Suggestion: use `new SDKHttpRequest(request)` here, and you won't need the temporary variable on line 72.

dbwiddis (changes requested):
This looks great!  

I've made a few suggestions for improvements, mostly in the documentation.

One more request:  our `org.opensearch.sdk` package is getting cluttered and we should move stuff out of it when we can.  Please create a new `org.opensearch.sdk.rest` package and put your two new SDK classes into it.

Might also be a good home for these classes (we can address in a later PR):
 - `ExtensionRestHandler`
 - `BaseExtensionRestHandler` (will break dependent extensions like AD but we can synchronize fix)
 - `RouteHandler` (will break dependent extensions like AD but we can synchronize fix)
 - `ExtensionRestPathRegistry`

dbwiddis (approved):
LGTM!

Tests should pass following merge of https://github.com/opensearch-project/OpenSearch/pull/6826

joshpalis (dismissed):

src/main/java/org/opensearch/sdk/rest/SDKHttpRequest.java
```diff
@@ -0,0 +1,124 @@
...
+    @Override
+    public BytesReference content() {
+        return content;
+    }
+
+    @Override
+    public Map<String, List<String>> getHeaders() {
+        return headers;
+    }
+
```
owaiskazi19, line 64: We should add tests for `SDKHttpRequest` and test these getters too.

dbwiddis: > We should add tests for `SDKHttpRequest` and test these getters too for better code coverage.

Good point, @owaiskazi19 .

The HttpRequest is embedded internally into the RestRequest, so the existing uses of `createTestRestRequest()` inside the `TestBaseExtensionRestHandler` instantiate both of these.   We should probably:
 - create a new test class `TestSDKRestRequest`
 - Move that static `createTestRestRequest()` method there so other test methods that need it can call it
 - add some tests of all the (implemented) getters.  Most tests of the SDKHttpRequest will come from getters on SDKRestRequest, for example the `RestRequest` `method()` and `uri()` methods just delegate to the internal `HttpRequest`.

src/test/java/org/opensearch/sdk/TestSDKRestRequest.java
```diff
@@ -0,0 +1,84 @@
...
+
+public class TestSDKRestRequest extends OpenSearchTestCase {
+    @Test
+    public void TestSDKRestRequestMethods() {
+        RestRequest.Method expectedMethod = Method.GET;
+        String expectedUri = "foobar?foo=bar&baz=42";
+        String expectedPath = "foo";
+        Map<String, String> expectedParams = Map.ofEntries(entry("foo", "bar"), entry("baz", "42"));
+        Map<String, List<String>> expectedHeaders = Map.ofEntries(entry("foo", Arrays.asList("hello", "world")));
+        XContentType exptectedXContentType = null;
```
dbwiddis, line 38: We should be able to test this better: if we pass as the headers a "Content-Type" of "application/json" then the `RestRequest` constructor does this:
```java
        final XContentType xContentType;
        try {
            xContentType = parseContentType(headers.get("Content-Type"));
        } catch (final IllegalArgumentException e) {
            throw new ContentTypeHeaderException(e);
        }
```

This also means that we don't really need to pass the xcontent type as part of the ExtensionRestRequest any more, as long as we're passing the XContent header.

src/test/java/org/opensearch/sdk/TestSDKRestRequest.java
```diff
@@ -0,0 +1,84 @@
...
+        BytesReference expectedContent = new BytesArray("bar");
```
dbwiddis, line 39: If we pass actual JSON here (`"{\"foo\":\"bar\"}`" then we can actually test the extracted result of the `contentParser()` method from the RestRequest:
```java
Map<String, Object> source = request.contentParser().map();
assertEquals("bar", (String) source.get("foo")); // I think?
```

dbwiddis (changes requested):
Looks good, some suggestions to improve it!

src/test/java/org/opensearch/sdk/TestSDKRestRequest.java
dbwiddis: > Should I remove xcontent type from ExtensionRestRequest?

I'm ok with leaving it for now, even if we don't use it.  You might add a comment that it's not needed if we are sure the content header is passed, and It's a good "backup" if we somehow lose the header.  We can make the final call when @saratvemulapalli moves those requests to protobuf.

nassipkali:
@dbwiddis `Map<String, Object> source = request.contentParser().map();`. Can I use there `mapStrings()`? mapStrings returns `Map<String, String` and doesnt need cast/convert Object to String

dbwiddis:
> @dbwiddis `Map<String, Object> source = request.contentParser().map();`. Can I use there `mapStrings()`? mapStrings returns `Map<String, String>` and doesnt need cast/convert Object to String

Yep!

nassipkali:
@dbwiddis When snapshot that contains latest commits of OpenSearch is will be used? My PR from OpenSearch side is merged but this PR checks didn't see changes from OpenSearch side.

dbwiddis (dismissed):

dbwiddis:
OK, new snapshot is being read but there's a missing method signature we're trying to read:

> /home/runner/work/opensearch-sdk-java/opensearch-sdk-java/src/main/java/org/opensearch/sdk/rest/SDKHttpRequest.java:42: error: cannot find symbol
> 
>         this.httpVersion = request.httpVersion();

nassipkali:
> OK, new snapshot is being read but there's a missing method signature we're trying to read:
> 
> > /home/runner/work/opensearch-sdk-java/opensearch-sdk-java/src/main/java/org/opensearch/sdk/rest/SDKHttpRequest.java:42: error: cannot find symbol
> > ```
> >     this.httpVersion = request.httpVersion();
> > ```

Yes, there should be protocolVersion()

dbwiddis:
> Yes, there should be protocolVersion()

Just found that myself :)

Should be a quick fix and we can merge this!

dbwiddis:
@nassipkali are you going to be fixing that method?

nassipkali:
> @nassipkali are you going to be fixing that method?
 I have a problem with git push. I made change from github site

dbwiddis:
> @dbwiddis I have a problem with git push. I made change from github web editor

Lots of different errors in the build now.

Locally you should be able to pull the changes down from your branch.

dbwiddis:
@nassipkali I'm going to go ahead and fix up the issues here, since the SDK build is now failing other PRs.

One challenge with these companion PRs is finding out what's broken before we merge the companion.  You should be using `./gradlew publishToMavenLocal` on the OpenSearch PR branch and then running `./gradlew check` on this PR branch to verify that things work.

owaiskazi19 (dismissed):

dbwiddis (approved):

owaiskazi19 (approved):

Issue #656, opened 2023-04-06
OpenSearch does not pickup extensions feature flag when enabled via `run.gradle` file


DarshitChanpura:
Following the Dev guide, when I tried to run OpenSearch with extensions enabled via "run.gradle," OpenSearch didn't pick up on this setting, so the experimental feature flag wasn't turned on.

I was able to work around this by passing the feature flag as command line argument. _See below_
```
./bin/opensearch -E opensearch.experimental.feature.extensions.enabled=true
```
```
```

OpenSearch: 3.0.0
OS: Mac
Arch: aarch64 M1

dbwiddis:
> when I tried to run OpenSearch with extensions enabled via "run.gradle," OpenSearch didn't pick up on this setting

Could you be more specific on what changes you made in run.gradle and how you were running OpenSearch?  It may just be unclear documentation.

DarshitChanpura:
> Could you be more specific on what changes you made in run.gradle and how you were running OpenSearch? It may just be unclear documentation.

Yes I modified run.gradle to add the feature flag.
<table>
<tr>
<td> Before </td> <td> After </td>
</tr>
<tr>
<td>

```
testClusters {
  runTask {
    testDistribution = 'archive'
    if (numZones > 1) numberOfZones = numZones
    if (numNodes > 1) numberOfNodes = numNodes
  }
}
```

</td>
<td>

```
testClusters {
  runTask {
    testDistribution = 'archive'
    if (numZones > 1) numberOfZones = numZones
    if (numNodes > 1) numberOfNodes = numNodes
   systemProperty 'opensearch.experimental.feature.extensions.enabled', 'true'
  }
}
```

</td>

I tried to assemble and run at-least 5 times but the feature flag was not picked up

DarshitChanpura:
Steps I followed:

1. Assemble OS local distribution
2. Copy that distribution to another location
3. cd to that location/opensearch-3.0.0-SNAPSHOT
4. run using `./bin/opensearch`

This didn't work for me.

However, when I modify `gradle/run.gradle` and run using `./gradlew run` the feature flag is picked up.

The steps in the dev-guide section that mentions `To run OpenSearch from a compiled binary:` didn't work for me.

PR #657, opened 2023-04-06
Update dev guide to add another option to run OpenSearch with Extensions feature-flag enable

DarshitChanpura:
### Description
Option 1, which is adding the feature flag in `gradle/run.gradle` did not work for me, but passing the feature flag as command line argument did. Hence, I'm adding alternate option, Option 2,  to run OpenSearch with Extensions feature-flag enabled in case someone runs into same issue as me.

### Issues Resolved
Related to #656 


owaiskazi19:
@DarshitChanpura thanks for the PR. There's an option 3 as well.
You can uncomment the flag [here](https://github.com/opensearch-project/OpenSearch/blob/main/distribution/src/config/opensearch.yml#L116) and run `./bin/opensearch`. Probably you can add in this PR.

DEVELOPER_GUIDE.md
```diff
@@ -75,6 +76,20 @@ testClusters {
...
+- `./gradlew run -Dopensearch.experimental.feature.extensions.enabled=true` when running using gradle in developer mode
+
+##### Option 3
+Enable the experimental feature flag by setting it to true in `opensearch.yml`.
+- cd in to your local distribution build for OpenSearch.
+- now cd into config folder and open `opensearch.yml` in your local editor
+- Search for `opensearch.experimental.feature.extensions.enabled`, uncomment and set it to `true`
+- Run OpenSearch using 
+  - `./bin/opensearch` when running from a local distribution
+  - `./gradlew run` when running using gradle in developer mode
```
owaiskazi19, line 91: I don't think so we can run `./gradlew run` for this option. For gradle run we need to add the feature flag in run.gradle which is already covered in option 1

dbwiddis: I still haven't seen an example of the not-working run.gradle file.  That's how I'm running it and it works.

DarshitChanpura: true

DarshitChanpura: @dbwiddis I think I know where I got confused. After modifying run.gradle, I tried running the assembled version with `./bin/opensearch` as mentioned here thinking that the feature flag will be copied over. But that is not the case. run.gradle modification works only with `./gradlew run` and not `./bin/opensearch`

In order to run it with `./bin/opensearch` you need to either configure the feature flag via opensearch.yml or via command-line argument

owaiskazi19 (approved):

joshpalis (approved):

PR #666, opened 2023-04-09
Handle Deprecated and Replaced Routes in Extension Rest Handlers

### LINKED ISSUE #665: [FEATURE] Add deprecated and replaced route handling

Many (most?) of the Anomaly Detection APIs we are migrating include replaced routes relating to OpenDistro.  We initially only implemented `routes()` under the assumption "new capability we don't need the deprecated handlers".  However, since we will be releasing Extensions in the OpenSearch 2.x line we need to support deprecated routes.

Deprecated routes are just routes with a wrapper that adds logging.

Replaced routes are a combination of new and deprecated routes with an auto-generated log message for the deprecation.

Implement identical deprecated/replaced handling as the RestController.

Not supporting backwards compatibility, or doing some more convoluted handling of this than simply replicating existing code.

This started as an attempt to fix #654 which turned out to be not a bug, but since the work is done, might as well submit it, eh?  Lots of benefit for migration in any case!

dbwiddis:
Companion PR on OpenSearch must be merged first: https://github.com/opensearch-project/OpenSearch/pull/7064
Gradle Check passes with companion PR published to maven local.

Merging this PR will break AD rest handlers due to the package change of the handler interface/abstract class to `.rest` , but after it's merged they should be able to be migrated much closer to the original plugin format (minimize the diff!)

### Description

Adds the ability to handle Deprecated and Replaced Routes to the ExtensionRestHandler.  This will permit (nearly) identical syntax for plugin migration, and provide backwards compatibility to extension users.
 - Routes were already registered.
 - Deprecated routes are just a wrapper around Routes, which add a log message.
 - Replaced routes are a fancy user interface to simultaneously register Routes and Deprecated Routes.

The code used was copied from the corresponding `RestController` methods, with changes adapted to Extensions.

Moved more classes into `o.o.sdk.rest` in the process since we're breaking AD and should do all the changes at once.

### Issues Resolved

Fixes #665.  


src/main/java/org/opensearch/sdk/rest/DeprecatedRouteHandler.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.opensearch.extensions.rest.ExtensionRestResponse;
+import org.opensearch.rest.RestHandler.DeprecatedRoute;
+import org.opensearch.rest.RestRequest;
+import org.opensearch.rest.RestRequest.Method;
+
+/**
+ * A subclass of {@link DeprecatedRoute} that includes a handler method for that route.
+ */
+public class DeprecatedRouteHandler extends DeprecatedRoute {
```
owaiskazi19, line 22: Can we have this class as a nested class in `ExtensionRestHandler`?

src/main/java/org/opensearch/sdk/rest/ExtensionDeprecationRestHandler.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.opensearch.common.logging.DeprecationLogger;
+import org.opensearch.extensions.rest.ExtensionRestResponse;
+import org.opensearch.rest.DeprecationRestHandler;
+import org.opensearch.rest.RestRequest;
+
+/**
+ * {@code ExtensionDeprecationRestHandler} provides a proxy for any existing {@link Extension RestHandler} so that usage of the handler can be logged using the {@link DeprecationLogger}.
+ */
+public class ExtensionDeprecationRestHandler implements ExtensionRestHandler {
```
owaiskazi19, line 22: We can have this class also as a nested class in `ExtensionRestHandler`

src/test/java/org/opensearch/sdk/rest/TestBaseExtensionRestHandler.java
```diff
@@ -0,0 +1,267 @@
...
+            "/new/put/foo",
+            "/new/put/foo",
+            Collections.emptyMap(),
+            Collections.emptyMap(),
+            null,
+            new BytesArray("bar".getBytes(StandardCharsets.UTF_8)),
+            "",
+            null
+        );
+        ExtensionRestResponse response = handler.handleRequest(successfulRequestOld);
```
owaiskazi19, line 161: Can we test here whether the old request has been replaced to the new one? Mostly by checking the URI path?

owaiskazi19 (commented):
Overall LGTM  with few suggestions

src/main/java/org/opensearch/sdk/rest/DeprecatedRouteHandler.java
dbwiddis: I'm on the fence about this one.  If we want users to use it they need to see it.  Whatever we do with this we should do with `RouteHandler`.

src/main/java/org/opensearch/sdk/rest/ExtensionDeprecationRestHandler.java
dbwiddis: Sure, this one makes sense.

src/test/java/org/opensearch/sdk/rest/TestBaseExtensionRestHandler.java
dbwiddis: There is no actual "replacement" happening.  A replaced route is just syntactic sugar for simultaneously registering a deprecated route and a route both pointing to the same handler, which is exactly what's been tested.  We could add another test of the deprecation message on the old handler.

dbwiddis: Added new tests in the registry that tested these routes were added to both deprecated and normal lists and tested the deprecation message too. :)

src/test/java/org/opensearch/sdk/rest/TestExtensionRestPathRegistry.java
```diff
@@ -0,0 +1,186 @@
...
+        assertTrue(registeredDeprecatedPaths.contains("PUT /old/put/foo"));
+        index = registeredDeprecatedPaths.indexOf("PUT /old/put/foo");
+        assertTrue(index >= 0);
+        assertEquals("[PUT /old/put/foo] is deprecated! Use [PUT /new/put/foo] instead.", registeredDeprecatedPaths.get(index + 1));
+
+        assertTrue(registeredDeprecatedPaths.contains("POST /old/foo"));
+        index = registeredDeprecatedPaths.indexOf("POST /old/foo");
+        assertTrue(index >= 0);
+        assertEquals("[POST /old/foo] is deprecated! Use [POST /new/foo] instead.", registeredDeprecatedPaths.get(index + 1));
+    }
```
owaiskazi19, line 180: This test explains the whole workflow!

owaiskazi19 (approved):
Thanks for incorporating the feedback. LGTM!

vibrantvarun:
BaseRestExtensionHandler does not have latest changes for unrecognized method. 

@dbwiddis

vibrantvarun (approved):
Great work @dbwiddis . LGTM 🎉

PR #668, opened 2023-04-10
Add ClusterStateRequest parameter to cluster state transport request

### LINKED ISSUE #354: [FEATURE] Add ability to pass ClusterStateRequest object when requesting Cluster State

dbwiddis:
Companion PR on OpenSearch: https://github.com/opensearch-project/OpenSearch/pull/7066

### Description

Adds the ability to send a `ClusterStateRequest` parameter when requesting Cluster State, to limit the values returned (and reduce transport bandwidth).  To get the old behavior, use `new ClusterStateRequest().all()` as the parameter.

### Issues Resolved

Fixes #354 (caveat: we should decide whether to add a call to this method in the SDKClient (which requires passing the TransportService around). For now it is accessible to extensions via `ExtensionsRunner.getSdkTransportService()`.

### DIFF
--- a/src/main/java/org/opensearch/sdk/SDKTransportService.java
+++ b/src/main/java/org/opensearch/sdk/SDKTransportService.java
@@ -18,6 +18,8 @@ import java.util.stream.Collectors;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.opensearch.action.admin.cluster.state.ClusterStateRequest;
+import org.opensearch.cluster.ClusterState;
 import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.extensions.ExtensionsManager;
 import org.opensearch.extensions.action.RegisterTransportActionsRequest;
@@ -27,6 +29,7 @@ import org.opensearch.sdk.api.ActionExtension.ActionHandler;
 import org.opensearch.sdk.action.RemoteExtensionActionRequest;
 import org.opensearch.sdk.action.SDKActionModule;
 import org.opensearch.sdk.handlers.AcknowledgedResponseHandler;
+import org.opensearch.sdk.handlers.ClusterStateResponseHandler;
 import org.opensearch.sdk.handlers.ExtensionActionResponseHandler;
 import org.opensearch.transport.TransportService;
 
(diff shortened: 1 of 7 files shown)


joshpalis (approved):

saratvemulapalli (approved):

src/main/java/org/opensearch/sdk/SDKTransportService.java
```diff
@@ -106,6 +109,32 @@ public RemoteExtensionActionResponse sendRemoteExtensionActionRequest(RemoteExte
...
+     * @param request a ClusterStateRequest object defining the information to be retrieved
+     * @return The requested cluster state. Only the parts of the cluster state that were requested are included in the returned {@link ClusterState} instance.
+     */
+    public ClusterState sendClusterStateRequest(ClusterStateRequest request) {
+        logger.info("Sending Cluster State request to OpenSearch");
+        ClusterStateResponseHandler clusterStateResponseHandler = new ClusterStateResponseHandler();
+        try {
+            transportService.sendRequest(
+                opensearchNode,
+                ExtensionsManager.REQUEST_EXTENSION_CLUSTER_STATE,
```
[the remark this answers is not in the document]
dbwiddis: Yeah, I thought so.   

Also, I'll be doing another PR for GetFieldMappings which will follow this template, and we (you) may want to make a more generic protobuf action for all the TransportActions (with associated ActionRequest/ActionResponse) with similar casing to this... they all implement Writeable.

Or we could adapt the ProxyAction code to do this. I started down that road but it got pretty spaghetti pretty fast.

src/test/java/org/opensearch/sdk/TestSDKTransportService.java
```diff
@@ -101,4 +107,43 @@ public void testRegisterTransportAction() {
...
+        ClusterStateRequest request = new ClusterStateRequest().clear().indices("foo", "bar");
+        sdkTransportService.sendClusterStateRequest(request);
+        verify(transportService, times(1)).sendRequest(
+            any(),
+            eq(ExtensionsManager.REQUEST_EXTENSION_CLUSTER_STATE),
+            clusterStateRequestCaptor.capture(),
+            any(ClusterStateResponseHandler.class)
+        );
+        assertArrayEquals(new String[] { "foo", "bar" }, clusterStateRequestCaptor.getValue().indices());
+    }
```
owaiskazi19, line 148: Do we have a test already for `ClusterStateRequest().all()`?

dbwiddis: > Do we have a test already for `ClusterStateRequest().all()`?

That's just part of the OpenSearch class.  It sets some booleans to true.  It's the opposite of `clear()` (which sets them to false.  There is a set of ClusterStateRequestTests which randomly set and test the booleans but I don't see one that explicitly tests that they're all set to true (or false). :)

owaiskazi19 (approved):

dbwiddis:
Based on feedback about sending cluster state over transport, I'm closing this PR in favor of targeted API requests for the information we need.

PR #670, opened 2023-04-10
Adding unrecognized method in BaseExtensionRestHandler

vibrantvarun:
### Description
Adding unrecognized method in BaseExtensionRestHandler for Extensions Architecture. The method calls unrecognized method from BaseRestHandler in OpenSearch.

### DIFF
--- a/src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
+++ b/src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
@@ -136,6 +136,15 @@ public abstract class BaseExtensionRestHandler implements ExtensionRestHandler {
         }
     }
 
+    /**
+     * Returns a String message of the detail of any unrecognized error occurred
+     *
+     * @param request The request that caused the exception
+     * @param invalids invalid strings
+     * @param candidates candidates involved
+     * @param detail
+     * @return an String that contains the message.
+     */
     public static final String unrecognized(RestRequest request, Set<String> invalids, Set<String> candidates, String detail) {
         StringBuilder message = new StringBuilder(
             String.format(Locale.ROOT, "request [%s] contains unrecognized %s%s: ", request.path(), detail, invalids.size() > 1 ? "s" : "")
--- a/src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
+++ b/src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import org.apache.lucene.search.spell.LevenshteinDistance;
 import org.apache.lucene.util.CollectionUtil;
 import org.opensearch.common.collect.Tuple;
+import org.opensearch.rest.BaseRestHandler;
 import static org.opensearch.rest.RestStatus.INTERNAL_SERVER_ERROR;
 import static org.opensearch.rest.RestStatus.NOT_FOUND;
 
@@ -136,13 +137,13 @@ public abstract class BaseExtensionRestHandler implements ExtensionRestHandler {
     }
 
     /**
-     * Returns a String message of the detail of any unrecognized error occurred
+     * Returns a String message of the detail of any unrecognized error occurred. The string is intended for use in error messages to be returned to the user.
      *
      * @param request The request that caused the exception
-     * @param invalids invalid strings
-     * @param candidates candidates involved
-     * @param detail
-     * @return an String that contains the message.
+     * @param invalids Strings from the request which were unable to be understood.
+     * @param candidates A set of words that are most likely to be the valid strings determined invalid, to be suggested to the user.
+     * @param detail The parameter contains the details of the exception.
+     * @return a String that contains the message.
      */
     protected final String unrecognized(
         final RestRequest request,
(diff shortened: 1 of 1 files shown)


joshpalis (dismissed):

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
```diff
@@ -127,4 +135,58 @@ protected ExtensionRestResponse createJsonResponse(RestRequest request, RestStat
...
+     * @param detail
+     * @return an String that contains the message.
+     */
+    public static final String unrecognized(RestRequest request, Set<String> invalids, Set<String> candidates, String detail) {
+        StringBuilder message = new StringBuilder(
+            String.format(Locale.ROOT, "request [%s] contains unrecognized %s%s: ", request.path(), detail, invalids.size() > 1 ? "s" : "")
+        );
+        boolean first = true;
+
+        for (Iterator var7 = invalids.iterator(); var7.hasNext(); first = false) {
```
owaiskazi19, line 154: Should be 
```suggestion
        for (final String invalid : invalids) {
```

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
```diff
@@ -127,4 +135,58 @@ protected ExtensionRestResponse createJsonResponse(RestRequest request, RestStat
...
+                    message.append("[").append((String) keys.get(0)).append("]");
+                } else {
+                    message.append("any of ").append(keys.toString());
+                }
+
+                message.append("?");
+            }
+        }
+
+        return message.toString();
```
owaiskazi19, line 190: I don't thin so this method is copy/pasted correctly [here](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/rest/BaseRestHandler.java#L131)

owaiskazi19 (changes requested):
Added comments

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
vibrantvarun: The variable first is always initialized as false in every iteration. If I accept you commit suggestion then I have to add another line which will initialized first as false for every iteration.  Therefore my suggestion would be to not change anything.

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
vibrantvarun: ACK

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
owaiskazi19: See the next comment. Copy/paste the method exactly. That will override this change

vibrantvarun: Got it. Pushed the updated changes.
Thanks

owaiskazi19:
@dbwiddis are you fine with this change?

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
```diff
@@ -127,4 +134,58 @@ protected ExtensionRestResponse createJsonResponse(RestRequest request, RestStat
             return new ExtensionRestResponse(request, status, fieldName + ": " + responseStr);
         }
     }
+
+    /**
+     * Returns a String message of the detail of any unrecognized error occurred
```
dbwiddis, line 139: Suggestion: add a sentence indicating it is intended for use in error messages to be returned to a user.

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
```diff
@@ -127,4 +134,58 @@ protected ExtensionRestResponse createJsonResponse(RestRequest request, RestStat
...
+     *
+     * @param request The request that caused the exception
+     * @param invalids invalid strings
+     * @param candidates candidates involved
+     * @param detail
```
dbwiddis, line 144: This needs a description of what it actually is.  Looking at line 154 it appears to be "a noun describing the category or type of invalid strings to be passed".    Also apparently it is supposed to be singular.   TLDR, I couldn't tell from reading this javadoc what I am supposed to do with this method.

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
```diff
@@ -127,4 +134,58 @@ protected ExtensionRestResponse createJsonResponse(RestRequest request, RestStat
             return new ExtensionRestResponse(request, status, fieldName + ": " + responseStr);
         }
     }
+
+    /**
+     * Returns a String message of the detail of any unrecognized error occurred
+     *
+     * @param request The request that caused the exception
+     * @param invalids invalid strings
+     * @param candidates candidates involved
```
dbwiddis, line 143: Are candidates "involved"?  I don't know what this means. Looking at the code it appears to be using levenshtein distance to provide a "did you mean" functionality.  So say so here.  "A set of words that are most likely to be the valid strings determined invalid, to be suggested to the user."

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
```diff
@@ -127,4 +134,58 @@ protected ExtensionRestResponse createJsonResponse(RestRequest request, RestStat
             return new ExtensionRestResponse(request, status, fieldName + ": " + responseStr);
         }
     }
+
+    /**
+     * Returns a String message of the detail of any unrecognized error occurred
+     *
+     * @param request The request that caused the exception
+     * @param invalids invalid strings
```
dbwiddis, line 142: "Strings from the request which were unable to be understood."

dbwiddis (changes requested):
Ugh.   So if we keep this, we need much better javadoc.  See specifics below.

But I am really not fond of duplicating code.  Is there any possible way of creating a thin wrapper method that delegates to the original?

```java
    // add this method to both BaseRestHandler and BaseExtensionRestHandler
    /**
     * Insert javadocs here
     */
    protected final String unrecognized(
        final RestRequest request,
        final Set<String> invalids,
        final Set<String> candidates,
        final String detail
    ) {
        return unrecognizedStrings(request, invalids, candidates, detail);
    }

    // change signature of existing method in BaseRestHandler
    /**
     * Insert javadocs here
     */
    public static String unrecognizedStrings(
        final RestRequest request,
        final Set<String> invalids,
        final Set<String> candidates,
        final String detail
    ) {
        // current unrecognized() code
    }
```

Or, this works in `BaseExtensionRestHandler` without changing the signature in `BaseRestHandler`.  Ugly and I wouldn't do it, but it's possible. Inheritance is wierd.
```java
    /**
     * Insert javadocs here
     */
    protected final String unrecognized(
        final RestRequest request,
        final Set<String> invalids,
        final Set<String> candidates,
        final String detail
    ) {        
        return new BaseRestHandler() {
            @Override
            public String getName() {
                return null;
            }

            @Override
            protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) throws IOException {
                return null;
            }            
            
            public String unrecognizedStrings(
                final RestRequest request,
                final Set<String> invalids,
                final Set<String> candidates,
                final String detail
                ) {
                return super.unrecognized(request, invalids, candidates, detail);                
            }
        }.unrecognizedStrings(request, invalids, candidates, detail);
    }
```

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
vibrantvarun: Updated

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
vibrantvarun: Updated

src/main/java/org/opensearch/sdk/BaseExtensionRestHandler.java
vibrantvarun: updated

dbwiddis (approved):
LGTM! Thanks for the refactor!

owaiskazi19 (approved):
Much cleaner!

Issue #589, opened 2023-03-20
Modify the gradle plugin to also read from extensions


ryanbogan:
Currently, the OpenSearch gradle plugin only reads from plugins, but in order to write integration tests for opensearch-sdk-java it needs to also read from extensions.

Action items for this issue:

`testclusters` folder:
- [x] Create file similar to `ExtensionsSettings` that can hold extension information 

`OpenSearchNode`:
- [x] Create method to take extensions and write to extensions.yml
- [x] Within this method, add a line in `config/jvm.options` to enable the extensions feature flag

SDK `build.gradle`:
- [x] Create integTest task, which contains `dependsOn(helloWorld)` at the beginning (This will run the extension before integration tests start)
- [x] Set up test clusters

Issue #144, opened 2022-09-16
[FEATURE] Create independent checks for extensions


owaiskazi19:
### Is your feature request related to a problem?
Currently, the checks like `licenseHeaders`, `dependencyLicenses`, `forbiddenApisTest`, `thirdPartyAudit` and `loggerUsageCheck` are fetch from OpenSearch. Since for extensions we are removing `opensearchplugin` gradle plugin, these checks should be either part of SDK or the extension.

### What solution would you like?
_A clear and concise description of what you want to happen._

### What alternatives have you considered?
_A clear and concise description of any alternative solutions or features you've considered._

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

ryanbogan:
With the integration test framework, we are now importing the gradle plugin, which contains these tests. Currently, the tests are intentionally disabled since some of them are failing.  The new action item for this issue should be to enable those tests and make the necessary changes to ensure they pass.

PR #687, opened 2023-04-19
Add integration test framework

ryanbogan:
### Description
Adds an integration test framework, utilizing the changes made in https://github.com/opensearch-project/OpenSearch/pull/7235.  The integration test task starts the helloWorld extension, spins up a test single-node cluster to initialize the extensions, and then kills the cluster and the extension process running in the background. There are currently no tests written for this framework, so the task will fail with an error temporarily.

Note: Must be merged after https://github.com/opensearch-project/OpenSearch/pull/7235

### Issues Resolved
https://github.com/opensearch-project/opensearch-sdk-java/issues/589


dblock:
Is this the first time a consumer of opensearch-sdk-java would be bringing in the gradle tasks from OpenSearch core? That feels like undesirable. The alternative would be to (re)implement a simpler version of the functionality of test cluster setup in the SDK.

saratvemulapalli:
> Is this the first time a consumer of opensearch-sdk-java would be bringing in the gradle tasks from OpenSearch core? That feels like undesirable. The alternative would be to (re)implement a simpler version of the functionality of test cluster setup in the SDK.

We definitely could implement one in SDK. We've taken this path to:
- Not duplicate code for test cluster setup
- Plugins migrating to extensions have all the features from the framework
- Seamless integration test support with `opensearch-build`

I definitely understand we'd like to keep the SDK as independent as possible, we'd definitely take it as a follow up.

saratvemulapalli:
@ryanbogan whats needed here to get this moving?

ryanbogan:
@saratvemulapalli Working on fixing merge conflicts now.  I also want to test it out with the new protobuf additions but I'll have everything finalized by EOD

ryanbogan:
I'm not sure why gradle check is failing for a dependency conflict.  We have the same version as OpenSearch and everything is passing on my local.

build.gradle
```diff
@@ -11,6 +11,26 @@
...
 import java.nio.file.Files
 import org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestFramework
+import org.opensearch.gradle.test.RestIntegTestTask
+import org.opensearch.gradle.testclusters.ExtensionsProperties
+
+buildscript {
+    ext {
+        opensearch_version = System.getProperty("opensearch.version", "3.0.0-SNAPSHOT")
+        isSnapshot = "true" == System.getProperty("build.snapshot", "true")
+        buildVersionQualifier = System.getProperty("build.version_qualifier", "")
```
vibrantvarun, line 21: Just curious, what is the significance of  buildVersionQualifier and why "" has been set here?

build.gradle
```diff
@@ -77,10 +137,14 @@ repositories {
...
 
+configurations.all {
+    resolutionStrategy.force("jakarta.json:jakarta.json-api:2.1.1")
+}
+
 dependencies {
 
     def opensearchVersion = "3.0.0-SNAPSHOT"
-    def log4jVersion = "2.20.0"
+    def log4jVersion = "2.17.1"
```
vibrantvarun, line 147: Just for my understanding why log4j has been degraded?

ryanbogan: This is to match the version in OpenSearch

build.gradle
ryanbogan: It's an identifier that can potentially be added on to a version constant.  Here, it isn't inherently set to "".  That is the default value for the variable in case the System property can't be found or is empty.

build.gradle
```diff
@@ -22,21 +42,49 @@ plugins {
...
 
+apply from: 'gradle/formatting.gradle'
+apply plugin: 'maven-publish'
+apply plugin: 'opensearch.opensearchplugin'
+apply plugin: 'opensearch.testclusters'
+
+opensearchplugin {
+    name 'opensearch-sdk-java'
+    description 'OpenSearch Software Development Kit'
+    classname 'Sample Class'
```
joshpalis, line 53: This class name should be the fully qualified class name of the main class, take Job Scheduler's [build.gradle](https://github.com/opensearch-project/job-scheduler/blob/ad33680eb6c2ea69ebc463cdbbee7e2a396d3e05/build.gradle#L58) for example

build.gradle
```diff
@@ -118,6 +185,12 @@ dependencies {
     testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:${junit5Version}")
     testImplementation("org.opensearch.test:framework:${opensearchVersion}")
     testRuntimeOnly("org.junit.platform:junit-platform-launcher:${junitPlatform}")
+
+    configurations.all {
+        resolutionStrategy.force("jakarta.json:jakarta.json-api:2.1.1")
```
joshpalis, line 190: just a nit, lets keep the version numbers defined at the start of the dependencies section. Makes maintaining/upgrading dependencies easier in the future.

joshpalis (dismissed):

joshpalis (approved):

build.gradle
```diff
@@ -99,6 +160,7 @@ dependencies {
     api("org.opensearch.client:opensearch-java:${opensearchVersion}")
     implementation("org.opensearch.plugin:transport-netty4-client:${opensearchVersion}")
     implementation("io.netty:netty-all:${nettyVersion}")
+    implementation("jakarta.inject:jakarta.inject-api:2.0.0")
```
owaiskazi19, line 163: Can we pull the version from `jakartaVersion` i.e. 2.1.1?
```suggestion
    implementation("jakarta.inject:jakarta.inject-api:${jakartaVersion}")
```

build.gradle
```diff
@@ -182,3 +257,67 @@ diffCoverageReport {
...
+    }
+
+    // The --debug-jvm command-line option makes the cluster debuggable; this makes the tests debuggable
+    if (System.getProperty("test.debug") != null) {
+        jvmArgs '-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005'
+    }
+}
+
+task closeTestExtension (type: Exec) {
+    commandLine 'bash', '-c', "kill \$(lsof -i:4532)"
```
owaiskazi19, line 302: We should run the test cluster on a different port other than 4532 because that's the port we have to run SDK or other extensions. This will kill SDK process as well.

build.gradle
```diff
@@ -182,3 +257,67 @@ diffCoverageReport {
...
+}
+
+tasks.named("integTest").configure { finalizedBy(closeTestExtension) }
+
+testClusters.integTest {
+    extension(new ExtensionsProperties("hello-world", "opensearch-sdk-java-1", "127.0.0.1", "4532", "1.0", "3.0.0", "3.0.0"))
+    testDistribution = "ARCHIVE"
+    // Cluster shrink exception thrown if we try to set numberOfNodes to 1, so only apply if > 1
+    if (_numNodes > 1) numberOfNodes = _numNodes
```
owaiskazi19, line 311: Will this condition ever work because we have defined `_numNodes = 1` above?

owaiskazi19 (changes requested):
Did an initial pass. Design LGTM. Added comments.

build.gradle
ryanbogan: Eventually, we will be pulling `_numNodes` from a SystemProperty and defaulting it to one.  This check will be necessary once that happens.

build.gradle
ryanbogan: That makes sense.

build.gradle
joshpalis: We can set `_numNodes` like so : `def _numNodes = findProperty('numNodes') as Integer ?: 1`

build.gradle
```diff
@@ -22,21 +42,49 @@ plugins {
...
 
+dependencyLicenses.enabled = false
+thirdPartyAudit.enabled = false
+loggerUsageCheck.enabled = false
+forbiddenApisMain.enabled = false
+forbiddenApisTest.enabled = false
+filepermissions.enabled = false
+forbiddenPatterns.enabled = false
+testingConventions.enabled = false
+validateNebulaPom.enabled = false
```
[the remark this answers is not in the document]
ryanbogan: These are checks that we didn't have before on the SDK, but come in as part of the gradle plugin IIRC.

build.gradle
ryanbogan: I changed the definition to read from the system property now

build.gradle
```diff
@@ -182,3 +257,67 @@ diffCoverageReport {
         failOnViolation = true
     }
 }
+
+task integTest(type: RestIntegTestTask) {
+    description = "Run tests against a cluster"
+    testClassesDirs = sourceSets.test.output.classesDirs
```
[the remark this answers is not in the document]
ryanbogan: I was just going to have a separate folder for tests inside of the test folder.  Changing this results in a lot of secondary errors.

build.gradle
ryanbogan: Looks like it's failing

ryanbogan: I could potentially try a resolutionStrategy too if we want to go that route

build.gradle
```diff
@@ -182,3 +257,67 @@ diffCoverageReport {
...
+}
+
+task closeTestExtension (type: Exec) {
+    commandLine 'bash', '-c', "kill \$(lsof -i:4532)"
+}
+
+tasks.named("integTest").configure { finalizedBy(closeTestExtension) }
+
+testClusters.integTest {
+    extension(new ExtensionsProperties("hello-world", "opensearch-sdk-java-1", "127.0.0.1", "4532", "1.0", "3.0.0", "3.0.0"))
```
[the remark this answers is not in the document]
owaiskazi19: Let's read these properties from a file rather keeping it static over here.

build.gradle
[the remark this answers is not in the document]
ryanbogan: https://github.com/opensearch-project/opensearch-sdk-java/issues/144

build.gradle
```diff
@@ -22,21 +42,50 @@ plugins {
 }
 
 
+apply from: 'gradle/formatting.gradle'
+apply plugin: 'maven-publish'
+apply plugin: 'opensearch.opensearchplugin'
+apply plugin: 'opensearch.testclusters'
+
+opensearchplugin {
```
joshpalis, line 50: seems that replacing the application gradle plugin with the opensearch gradle plugin breaks `./gradlew run` and does not execute the configured main class of the opensearch gradle plugin.

CC : @owaiskazi19

joshpalis (changes requested):

DEVELOPER_GUIDE.md
```diff
@@ -221,6 +221,16 @@ Run tests :
 ```
 ./gradlew clean build test
 ```
+
+Run integration tests :
+```
+./gradlew clean build integTest
+```
+
+## Launching and debugging from an IDE
```
owaiskazi19, line 230: Can you add the entry in index?

build.gradle
owaiskazi19: @ryanbogan can verify the above? We had `application` plugin before to run the SDK with defined `MainClassName`.

joshpalis: we cannot have both plugins as `application` and `opensearchplugin` both define a `run` task

owaiskazi19: @joshpalis I didn't say we can have both :slightly_smiling_face:. Since I added `application` plugin I'm aware of that. That comment was just to explain the usage of `application`.

ryanbogan: Yeah we ran into this issue yesterday - I'm not sure why it's not seeing the class name to run the extension.

build.gradle
owaiskazi19: Are we facing any issue with `2.20.0`? SDK should be independent of OpenSearch.

build.gradle
owaiskazi19: I tried a couple of approaches which didn't help either:
1. Created a main file under `org.opensearch.sdk` instead of using HelloWorldExtension.
2. Included/excluded few dependencies which are added in this PR.
3. Tried debugger for `main` class. `opensearchplugin` doesn't invoke the process at all.

@dbwiddis @saratvemulapalli any idea here?

src/main/resources/sample/helloworld-settings.yml
```diff
@@ -1,10 +1,7 @@
 # This name is only used as a response string during init and is purely cosmetic
 extensionName: hello-world
-# The listening port is bound to the interface with this IP. Set to 0.0.0.0 to listen on all interfaces
-hostAddress: 0.0.0.0
```
joshpalis, line 4: Im curious as to how this modification passed the checks. Won't this modification break this test, since the expected extension hostAddress setting for this test is 0.0.0.0: https://github.com/opensearch-project/opensearch-sdk-java/blob/579c3020cc2775df1cf014b67ad6ef8f61b27766/src/test/java/org/opensearch/sdk/sample/helloworld/TestHelloWorldExtension.java#L105

ryanbogan: That must have happened as part of a merge - I didn't change the address at all

joshpalis: let's also keep the comments on this file

ryanbogan: Fixed

src/main/java/org/opensearch/sdk/sample/helloworld/HelloWorldExtension.java
```diff
@@ -79,6 +79,7 @@ public List<Setting<?>> getSettings() {
      */
     public static void main(String[] args) throws IOException {
         // Execute this extension by instantiating it and passing to ExtensionsRunner
+        System.out.println("Hi");
```
joshpalis, line 82: Do we still want to keep this line?

build.gradle
```diff
@@ -22,21 +42,51 @@ plugins {
 }
 
 
+apply from: 'gradle/formatting.gradle'
+apply plugin: 'maven-publish'
+apply plugin: 'java'
+apply plugin: 'opensearch.opensearchplugin'
+apply plugin: 'opensearch.testclusters'
+
+opensearchplugin {
```
owaiskazi19, line 51: Can you create an issue on creating a new gradle plugin `opensearchextension` which shouldn't relay on plugin.descriptor property file? That's the reason for `./gradlew run` not working because of these changes.

ryanbogan: But how would that work when we need to run plugins and extensions at the same time?

owaiskazi19: SDK would inherit `opensearchextension` gradle plugin and we need to figure out a way to run it either by creating a new file like the descriptor property or by completely removing it.

ryanbogan: So you're saying we would have to inherit both an extensions plugin and the current plugin for gradle?

ryanbogan: In that scenario, `opensearchextension` would have to contain all the current code to run plugins, except for the plugin-descriptor part, and contain the additions I made to run extensions

owaiskazi19: Plugins will still keep using `opensearchplugin`.

owaiskazi19: `opensearchextension` should consider all these use cases.

ryanbogan: I'll create an issue for the gradle run failure that we're seeing, but we should have further discussion about the solution.

ryanbogan: https://github.com/opensearch-project/opensearch-sdk-java/issues/789

joshpalis (approved):
LGTM, thanks for the changes Ryan.

owaiskazi19:
The path for the integTest folder should be under a directory called `test` -> `src/test/java/org/opensearch/sdk/test/integTest/TransportCommunicationIT.java`.

PR #697, opened 2023-04-25
Fix javadocs and improve javadoc CI

dbwiddis:
### Description

Actually fixes the build, as my previous fix in #696 was incorrect.  You don't need to include params in a linked method but you should leave off the parentheses in that case.

Also updated our CI to fail fast on javadoc checks to prevent these errors from slipping in.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,12 +10,19 @@ jobs:
   spotless:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v3
       - name: Spotless Check
         run: ./gradlew spotlessCheck
+  javadoc:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Javadoc CheckStyle
+        run: ./gradlew checkstyleMain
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -10,12 +10,19 @@ jobs:
...
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v3
       - name: Spotless Check
         run: ./gradlew spotlessCheck
+  javadoc:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Javadoc CheckStyle
+        run: ./gradlew checkstyleMain
```
owaiskazi19, line 21: Should we create a separate task for Checkstyle and not have it under javaDoc?

dbwiddis: The only thing we are using checkstyle for right now is javadocs. So for now I think it's fine here.  

If we decide to use checkstyle for any other code hygiene stuff I'm fine with splitting it out, but I'd prefer to wait until we actually do that (e.g., star imports).

joshpalis (approved):

owaiskazi19 (approved):

dbwiddis:
I'll do the backport.

Issue #378, opened 2023-01-29
Migrate Delete Detector results to SDK using SDKRestClient


dbwiddis:
### Is your feature request related to a problem?
 
Part of https://github.com/opensearch-project/OpenSearch/issues/5224
 
### What solution would you like?
 
Migrate the Delete Detector results Rest API to Extensions SDK using SDKRestClient.
 
Use the implementation of #353 as a reference.  Add new issues for any steps you skip.
 
Implement the following API endpoints:
 - [ ] `DELETE _plugins/_anomaly_detection/detectors/results`

owaiskazi19:
API: DELETE localhost:9200/_extensions/_ad-extension/detectors/results

```
{
  "query": {
    "bool": {
      "filter": [
        {
          "term": {
            "detector_id": {
              "value": "XqAwu4cBQBVk3lUtpDvr"
            }
          }
        }
      ]
    }
  }
}
```

Response:
```
{
    "took": 304,
    "timed_out": false,
    "total": 20,
    "updated": 0,
    "created": 0,
    "deleted": 20,
    "batches": 1,
    "version_conflicts": 0,
    "noops": 0,
    "retries": {
        "bulk": 0,
        "search": 0
    },
    "throttled_millis": 0,
    "requests_per_second": -1.0,
    "throttled_until_millis": 0,
    "failures": []
}
```

PR #708, opened 2023-04-27
Adds DeleteByQuery Request for SDKRestClient

owaiskazi19:
### Description
Required for Delete Detector Results to query OpenSearch using HLRC.

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-sdk-java/issues/378

### DIFF
--- a/src/main/java/org/opensearch/sdk/SDKClient.java
+++ b/src/main/java/org/opensearch/sdk/SDKClient.java
@@ -83,6 +83,8 @@ import org.opensearch.client.opensearch.OpenSearchAsyncClient;
 import org.opensearch.client.opensearch.OpenSearchClient;
 import org.opensearch.client.transport.OpenSearchTransport;
 import org.opensearch.client.transport.rest_client.RestClientTransport;
+import org.opensearch.index.reindex.BulkByScrollResponse;
+import org.opensearch.index.reindex.DeleteByQueryRequest;
 
 import javax.net.ssl.SSLEngine;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -83,6 +83,8 @@
 import org.opensearch.client.opensearch.OpenSearchClient;
 import org.opensearch.client.transport.OpenSearchTransport;
 import org.opensearch.client.transport.rest_client.RestClientTransport;
+import org.opensearch.index.reindex.BulkByScrollResponse;
+import org.opensearch.index.reindex.DeleteByQueryRequest;
```
dbwiddis, line 87: Unusual that these are the only imports that aren't `o.o.client.*`.   Any idea why?

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -464,6 +466,17 @@ public void delete(DeleteRequest request, ActionListener<DeleteResponse> listene
...
         }
 
+        /**
+         * Deletes a document from the index based on the query.
+         *
+         * @param request The delete by query request
+         * @param listener A listener to be notified with a result
+         *
+         */
+        public void deleteByQueryAsync(DeleteByQueryRequest request, ActionListener<BulkByScrollResponse> listener) {
```
dbwiddis, line 476: This is also different than all the others by adding `Async` to it.  Is this how it's called from extensions now?

dbwiddis (commented):
LGTM (will approve after questions answered) but I'm really interested in why this request looks so different to all the others.

src/main/java/org/opensearch/sdk/SDKClient.java
joshpalis: Delete/UpdateByQueryRequest classes are part of the Reindex module

dbwiddis (approved):

joshpalis (approved):

src/main/java/org/opensearch/sdk/SDKClient.java
owaiskazi19: Since this has been merged. I'll take care of it in the next one.

PR #730, opened 2023-05-05
Update OpenSearch host and port on SDKRestClient after init

### LINKED ISSUE #729: [BUG] OpenSearch and Extension host and port are required before initializing either
### What is the bug?

As currently designed:
 - 🐔 we must know the Extension host and port (`extensions.yml` file) and the extension must be running prior to starting OpenSearch.
 - 🥚 when initializing an Extension, we initialize and inject the `SDKRestClient` which requires the host and port of OpenSearch. 

This works fine when we're testing everything on `localhost` but fails in the real world use case where you are starting up a cluster using automation and don't know in advance its eventual ELB hostname when starting the extension.

Attempt to use `opensearch-cluster-cdk` to spin up an OpenSearch cluster attached to an Extension on an EC2 instance in the same account.

1. Extension does not need OpenSearch host/port on initialization
2. When OpenSearch connects to extension, it passes its connection information
3. Extension then knows how to connect to OpenSearch

I commented about this [here](https://github.com/opensearch-project/opensearch-sdk-java/issues/206#issuecomment-1304234524).

This will absolutely be required to be handled in a future case involving stateless (AWS Lambda/Azure function/OCI function) connections.

One possible solution is to not inject the RestClient; just inject the SDKClient and initialize the RestClient internally after the extension has been initialized (similar pattern we use for NamedXContentRegistry which, I might note, is tightly integrated with RESTClient).

Another solution looks like there are potentially some workarounds with the low level `RestClient` which looks designed for this use case: 
 - The constructor using `HttpHost` varargs has javadoc "You can use this if you do not have metadata up front about the nodes."
 -  There is a `setNodes(Collection<Node> nodes)` method

dbwiddis:
### Description

Updates the host and port on the RestHighLevelClient following Extension initialization.

This is a combination "quick fix" workaround for the way we've already integrated the deprecated client for migration, plus an update of the client's ExtensionSettings which should work just fine for new instantiations of the Java Clients (which we should not be injecting, see #612).

### Issues Resolved

Fixes #729 

### DIFF
--- a/src/main/java/org/opensearch/sdk/SDKClient.java
+++ b/src/main/java/org/opensearch/sdk/SDKClient.java
@@ -12,6 +12,7 @@ package org.opensearch.sdk;
 import java.io.Closeable;
 import java.io.IOException;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
@@ -61,6 +62,7 @@ import org.opensearch.client.ClusterAdminClient;
 import org.opensearch.client.ClusterClient;
 import org.opensearch.client.GetAliasesResponse;
 import org.opensearch.client.IndicesClient;
+import org.opensearch.client.Node;
 import org.opensearch.client.RequestOptions;
 import org.opensearch.client.Requests;
 import org.opensearch.client.RestClient;
@@ -83,6 +85,7 @@ import org.opensearch.client.opensearch.OpenSearchAsyncClient;
 import org.opensearch.client.opensearch.OpenSearchClient;
 import org.opensearch.client.transport.OpenSearchTransport;
 import org.opensearch.client.transport.rest_client.RestClientTransport;
+import org.opensearch.common.transport.TransportAddress;
 import org.opensearch.index.reindex.BulkByScrollResponse;
 import org.opensearch.index.reindex.DeleteByQueryRequest;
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -125,6 +128,22 @@ public void initialize(Map<ActionType, TransportAction> actions) {
...
+     * Update the ExtensionSettings with a new OpenSearch address and port
+     * @param address the TransportAddress associated with the OpenSearchNode
+     */
+    public void updateOpenSearchNodeSettings(TransportAddress address) {
+        // Update the settings for future initialization of new clients
+        this.extensionSettings.setOpensearchAddress(address.getAddress());
+        this.extensionSettings.setOpensearchPort(Integer.toString(address.getPort()));
+        // Update the settings on the already-initialized SDKRestClient (Deprecated -- for migration use)
+        if (this.sdkRestClient != null) {
+            sdkRestClient.getRestHighLevelClient()
```
owaiskazi19, line 141: Let's keep it consistent with `this.sdkRestClient`

src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -288,7 +313,7 @@ public void doCloseJavaClients() throws IOException {
      * @throws IOException if closing the highLevelClient fails
      */
     public void doCloseHighLevelClient() throws IOException {
-        if (sdkRestClient != null) {
+        if (this.sdkRestClient != null) {
             sdkRestClient.close();
```
owaiskazi19, line 317: Same as above

owaiskazi19 (dismissed):
LGTM for the quick fix! Let's create an issue to address the overall design to handle the initialization.

owaiskazi19 (approved):

saratvemulapalli (approved):

PR #736, opened 2023-05-07
Added method for create OK status response with empty JSON content

### LINKED ISSUE #721: [FEATURE] Reduce boilerplate for empty JSON "OK" responses

Success for many REST requests results in a response with status OK (200) and an empty JSON object (`{}`).

There is lots of boilerplate code generating this.  Plugin code looks like:
```java
XContentBuilder contentBuilder = r.toXContent(channel.newBuilder().startObject(), ToXContent.EMPTY_PARAMS);
contentBuilder.endObject();
channel.sendResponse(new BytesRestResponse(RestStatus.OK, contentBuilder));
```
In extensions we are migrating this to:
```java
XContentBuilder contentBuilder = r.toXContent(JsonXContent.contentBuilder().startObject(), ToXContent.EMPTY_PARAMS);
contentBuilder.endObject();
ExtensionRestResponse extensionRestResponse = new ExtensionRestResponse(request, RestStatus.OK, contentBuilder);
```
Each `r` in the above snippets is a different sort of Response object (usually); but all implement the `ToXContent` interface.

In addition, we also have this code repeated often, using the above `contentBuilder` instance:
```java
ExtensionRestResponse extensionRestResponse = new ExtensionRestResponse(request, RestStatus.OK, contentBuilder);
```

Create some constants/methods in the `BaseExtensionRestHandler` to reduce or eliminate the boilerplate.  I believe (this is untested code) that something like the following would create the `XContentBuilder` object:
```java
import org.opensearch.core.xcontent.ToXContent;

public static XContentBuilder EMPTY_JSON_XCONTENT = (new ToXContent() {
    @Override  
    toXContent(XContentBuilder builder, Params params) throws IOException {
        builder.endObject();
        return builder;
    }).toXContent(JsonXContent.contentBuilder().startObject(), ToXContent.EMPTY_PARAMS);
}
```
(Note that you could probably move `startObject()` next to `endObject()`; I only copied the above to stay somewhat true to the original code.)

Once that is created, you could also have a canned "ok" response, possibly:
```java
protected ExtensionRestResponse okNoContent(RestRequest request) {
     return new ExtensionRestResponse(request, RestStatus.OK, EMPTY_JSON_XCONTENT);
}
```

The `RestResponse` (and by extension (literally in more than one way) the `ExtensionRestResponse` class) permits you to just specify the content type, and the content as a `String`, so this could also work, possibly faster, and we wouldn't need the whole builder.  While this would work, I don't like hardcoding constants like this so prefer the above approach.
```java
protected ExtensionRestResponse okNoContent(RestRequest request) {
     return new ExtensionRestResponse(request, RestStatus.OK, "application/json; charset=UTF-8", "{}");
}
```

Coming from https://github.com/opensearch-project/anomaly-detection/pull/881/files#r1181795885 and https://github.com/opensearch-project/anomaly-detection/pull/881/files#r1181857497

nassipkali:
### Description
Added method for create OK status response with empty JSON content

### Issues Resolved
#721


src/main/java/org/opensearch/sdk/rest/BaseExtensionRestHandler.java
```diff
@@ -206,6 +206,16 @@ protected final String unrecognized(
...
     }
 
+    /**
+    * Creates a new plain text response with OK status and empty JSON content
+    *
+    * @param request the REST request being responded to
+    * @return ExtensionRestResponse with OK status response
+    */
+    protected ExtensionRestResponse okNoContent(RestRequest request) {
+        return new ExtensionRestResponse(request, RestStatus.OK, "application/json; charset=UTF-8", "{}");
```
dbwiddis, line 216: Can we make that JSON content string a constant (or two constants, one for the type and one for the charset)?  Either define our own or see if it's already included as a constant in an existing dependency.

For example we use Apache http core so [this constant](https://javadoc.io/static/org.apache.httpcomponents.core5/httpcore5/5.0.2/org/apache/hc/core5/http/ContentType.html#APPLICATION_JSON) (v5) [(and this for 1.x backport)](https://www.javadoc.io/static/org.apache.httpcomponents/httpcore/4.4.6/org/apache/http/entity/ContentType.html#APPLICATION_JSON) is available.  

Also, we might not even need the charset as [JSON is UTF-8 by default](https://stackoverflow.com/a/9254967/1161484).

dbwiddis: Looking at the Base Extension Rest Handler again, I see we already have a `createJSONResponse()` method:
```java
    protected ExtensionRestResponse createJsonResponse(RestRequest request, RestStatus status, String fieldName, String responseStr) {
        try {
            return new ExtensionRestResponse(
                request,
                status,
                JsonXContent.contentBuilder().startObject().field(fieldName, responseStr).endObject()
            );
        } catch (IOException e) {
            // This Should Never Happen (TM)
            // If we messed up JSON code above, just send plain text
            return new ExtensionRestResponse(request, status, fieldName + ": " + responseStr);
        }
    }
```
May want to just name your method `createEmptyJsonResponse` and allow the user to pass the OK param.

src/main/java/org/opensearch/sdk/rest/BaseExtensionRestHandler.java
```diff
@@ -36,6 +36,10 @@
  * Provides convenience methods to reduce boilerplate code in an {@link ExtensionRestHandler} implementation.
  */
 public abstract class BaseExtensionRestHandler implements ExtensionRestHandler {
+    /**
+     * Constant for JSON content type
+     */
+    public static final String JSON_CONTENT_TYPE = "application/json";
```
dbwiddis, line 42: Please use the built-in constant from Apache HTTP core.
```
import static org.apache.hc.core5.http.ContentType.APPLICATION_JSON;

public static final String JSON_CONTENT_TYPE = APPLICATION_JSON.toString();
```

dbwiddis (dismissed):
LGTM but would prefer to use a known constant to create the string... does the same thing!

Great test case in this PR!

src/main/java/org/opensearch/sdk/rest/BaseExtensionRestHandler.java
dbwiddis: Or, `APPLICATION_JSON.withCharset(StandardCharsets.UTF_8).toString()` ... I like this one. 

```java
System.out.println(org.apache.hc.core5.http.ContentType.APPLICATION_JSON.withCharset(StandardCharsets.UTF_8).toString());
```
> application/json; charset=UTF-8

ryanbogan (approved):

Issue #732, opened 2023-05-05
[PROPOSAL] Redesign the SDKClient


dbwiddis:
### What are you proposing?

We should redesign the SDKClient from scratch, without referring to the existing client which is trying to do too many things and doing some of them poorly.

Our original vision (if we ever had one) for the SDKClient has morphed over time into an increasingly unmaintainable pile of spaghetti code.  It looks like we started wanting to use it just as a wrapper for the OpenSearch Client for Java (`opensearch-java`), but then:
 - We pivoted to use RestHighLevelClient for migration and added that
 - We tried to handle all the existing plugin calls but are using a different Client Interface than the plugins
 - We added support for transport actions to parallel what NodeClient does
 - We added support for remote/proxy transport actions on other extensions
 - We added support for RequestOptions (headers) 

Core capabilities we need:
 - Handle transport actions (`client.execute()`)
 - Provide a bridge for migration, internally using RestHighLevelClient calls and attempting to use the exact signature as existing clients including the same request/response classes

### What users have asked for this feature?

#612, #729, #430, #616, TBD issue on serverless

### What problems are you trying to solve?

Replace existing usages of NodeClient implementations in plugins, specifically
 - Transport Actions (`client.execute()`) internal to the extension
 - Index, Cluster, and Admin client actions that need to be handled by a REST client
 - Support for headers/authentication associated with the REST client

### What is the developer experience going to be?

For migrating existing plugins, a minimal diff.

For developing new plugins, a set of clearly documented tools providing needed capabilities.

### Why should it be built? Any reason not to?

Properly handle initialization
Properly handle resource usage/leakage
Properly handle multithreading
Permit DRY

### What will it take to execute?

Re-thinking/re-scoping what the object does and does not do.

### Any remaining open questions?

 - Should we be making the Java Clients part of the SDK API at all or just documenting how to use them? #612 
 - Should we be initializing the (deprecated) RestClient as a singleton and how does that fit into the initialization sequence? #729 
 - How do we avoid resource leaks if we do initialize it ourself?  We have a close method but that's pretty useless on our singleton.  Can we follow a try-with-resources model for all usages of the client?
 - Should we try to use the same client Request/Response classes that plugins do with the NodeClient, which differ slightly in some cases with the RestHighLevelClient?  (Would require a lot of translating one request class to another)

dblock:
I am a big fan of deferring client responsibilities to opensearch-java and taking a runtime dependency on it.

[...]

PR #745, opened 2023-05-09
Replaced implementation() by api() for opensearch libs

### LINKED ISSUE #609: [BUG] Transitive Dependencies are not exposed to dependent projects
### What is the bug?

I am working on a "from scratch" setup.  I have declared a dependency only to `opensearch-sdk-java` 2.0.0-SNAPSHOT.

When attempting to import a class in `opensearch` 3.0.0-SNAPSHOT package `org.opensearch.extensions`, compilation fails.

1. Create a new project that depends only on `opensearch-sdk-java`, like [this one](https://github.com/dbwiddis/CRUDExtension)
2. Import a class from a transitive dependency such as OpenSearch.
3. Observe compile time failure

Dependency resolution shows the import is `runtime` only:
```
$ mvn dependency:tree
[INFO] CRUDExtension:CRUDExtension:jar:0.0.1-SNAPSHOT
[INFO] \- org.opensearch.sdk:opensearch-sdk-java:jar:2.0.0-SNAPSHOT:compile
[INFO]    +- org.opensearch:opensearch:jar:3.0.0-SNAPSHOT:runtime
```

Adding the transitive dependency manually resolves this problem, but should not be required.

Developers should not be required to declare dependencies themselves; transitive compile dependencies should resolve automatically.

According to [Gradle Docs](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation)
> The key difference between the standard Java plugin and the Java Library plugin is that the latter introduces the concept of an API exposed to consumers. A library is a Java component meant to be consumed by other components. It’s a very common use case in multi-project builds, but also as soon as you have external dependencies.
>
> The plugin exposes two [configurations](https://docs.gradle.org/current/userguide/declaring_dependencies.html#sec:what-are-dependency-configurations) that can be used to declare dependencies: api and implementation. The api configuration should be used to declare dependencies which are exported by the library API, whereas the implementation configuration should be used to declare dependencies which are internal to the component.

This suggests that this line in `build.gradle`:

```groovy
implementation("org.opensearch:opensearch:${opensearchVersion}")
```

should be:

```groovy
api("org.opensearch:opensearch:${opensearchVersion}")
```

This is likely true for other transitive dependencies such as `log4j-api` and the OpenSearch clients.

nassipkali:
### Description
Replaced implementation() by api() for opensearch libs

### Issues Resolved
Fixes #609 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,7 +13,7 @@ import java.nio.file.Files
 import org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestFramework
 
 plugins {
-    id 'java'
+    id 'java-library'
     id "com.diffplug.spotless" version "6.18.0" apply false
     id 'jacoco'
     id "com.form.diff-coverage" version "0.9.5"
@@ -90,13 +90,13 @@ dependencies {
     def junitPlatform = "1.9.3"
     def jaxbVersion = "2.3.1"
 
-    implementation("org.opensearch:opensearch:${opensearchVersion}")
+    api("org.opensearch:opensearch:${opensearchVersion}")
     implementation("org.apache.logging.log4j:log4j-api:${log4jVersion}")
     implementation("org.apache.logging.log4j:log4j-core:${log4jVersion}")
     implementation("org.apache.logging.log4j:log4j-jul:${log4jVersion}")
-    implementation("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
-    implementation("org.opensearch.client:opensearch-rest-client:${opensearchVersion}")
-    implementation("org.opensearch.client:opensearch-java:${opensearchVersion}")
+    api("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
+    api("org.opensearch.client:opensearch-rest-client:${opensearchVersion}")
+    api("org.opensearch.client:opensearch-java:${opensearchVersion}")
     implementation("org.opensearch.plugin:transport-netty4-client:${opensearchVersion}")
     implementation("io.netty:netty-all:${nettyVersion}")
     testCompileOnly("junit:junit:${junit4Version}") {
--- a/build.gradle
+++ b/build.gradle
@@ -94,7 +94,7 @@ dependencies {
     implementation("org.apache.logging.log4j:log4j-api:${log4jVersion}")
     implementation("org.apache.logging.log4j:log4j-core:${log4jVersion}")
     implementation("org.apache.logging.log4j:log4j-jul:${log4jVersion}")
-    api("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
+    implementation("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
     api("org.opensearch.client:opensearch-rest-client:${opensearchVersion}")
     api("org.opensearch.client:opensearch-java:${opensearchVersion}")
     implementation("org.opensearch.plugin:transport-netty4-client:${opensearchVersion}")


nassipkali:
@dbwiddis I tested it on CRUDExtension and it removed requirement of defining opensearch, clients in dependencies manually. But still needs add to dependencies a lot of apache-lucene libraries

nassipkali:
@dbwiddis Like this:
```xml
<dependency>
	<groupId>org.apache.lucene</groupId>
	<artifactId>lucene-core</artifactId>
	<version>9.6.0-SNAPSHOT</version>
</dependency>
<dependency>
	<groupId>org.apache.lucene</groupId>
	<artifactId>lucene-analysis-common</artifactId>
	<version>9.6.0-SNAPSHOT</version>
</dependency>
<dependency>
	<groupId>org.apache.lucene</groupId>
	<artifactId>lucene-backward-codecs</artifactId>
	<version>9.6.0-SNAPSHOT</version>
</dependency>
<dependency>
	<groupId>org.apache.lucene</groupId>
	<artifactId>lucene-grouping</artifactId>
	<version>9.6.0-SNAPSHOT</version>
</dependency>
<dependency>
	<groupId>org.apache.lucene</groupId>
	<artifactId>lucene-highlighter</artifactId>
	<version>9.6.0-SNAPSHOT</version>
</dependency>
<dependency>
	<groupId>org.apache.lucene</groupId>
	<artifactId>lucene-join</artifactId>
	<version>9.6.0-SNAPSHOT</version>
</dependency>
<dependency>
	<groupId>org.apache.lucene</groupId>
	<artifactId>lucene-memory</artifactId>
	<version>9.6.0-SNAPSHOT</version>
</dependency>
<dependency>
	<groupId>org.apache.lucene</groupId>
	<artifactId>lucene-misc</artifactId>
	<version>9.6.0-SNAPSHOT</version>
</dependency>
<dependency>
	<groupId>org.apache.lucene</groupId>
	<artifactId>lucene-queries</artifactId>
	<version>9.6.0-SNAPSHOT</version>
</dependency>
<dependency>
	<groupId>org.apache.lucene</groupId>
	<artifactId>lucene-queryparser</artifactId>
	<version>9.6.0-SNAPSHOT</version>
</dependency>
<dependency>
	<groupId>org.apache.lucene</groupId>
	<artifactId>lucene-sandbox</artifactId>
	<version>9.6.0-SNAPSHOT</version>
</dependency>
<dependency>
	<groupId>org.apache.lucene</groupId>
	<artifactId>lucene-spatial-extras</artifactId>
	<version>9.6.0-SNAPSHOT</version>
</dependency>
<dependency>
	<groupId>org.apache.lucene</groupId>
	<artifactId>lucene-spatial3d</artifactId>
	<version>9.6.0-SNAPSHOT</version>
</dependency>
<dependency>
	<groupId>org.apache.lucene</groupId>
	<artifactId>lucene-suggest</artifactId>
	<version>9.6.0-SNAPSHOT</version>
</dependency>
```

build.gradle
```diff
@@ -90,13 +90,13 @@ dependencies {
...
+    api("org.opensearch:opensearch:${opensearchVersion}")
     implementation("org.apache.logging.log4j:log4j-api:${log4jVersion}")
     implementation("org.apache.logging.log4j:log4j-core:${log4jVersion}")
     implementation("org.apache.logging.log4j:log4j-jul:${log4jVersion}")
-    implementation("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
-    implementation("org.opensearch.client:opensearch-rest-client:${opensearchVersion}")
-    implementation("org.opensearch.client:opensearch-java:${opensearchVersion}")
+    api("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
+    api("org.opensearch.client:opensearch-rest-client:${opensearchVersion}")
+    api("org.opensearch.client:opensearch-java:${opensearchVersion}")
```
dbwiddis, line 99: As a note to other reviewers, there are open issues to remove these clients completely: #612 and #732.  So this is appropriate temporarily but I hope eventually these lines disappear entirely.

dbwiddis (changes requested):
See below.

dbwiddis:
> @dbwiddis I tested it on CRUDExtension and it removed requirement of defining opensearch, clients in dependencies manually. But still needs add to dependencies a lot of apache-lucene libraries

Those libraries should come in via OpenSearch core, though, not the RestHighLevelClient... I think?

nassipkali:
@dbwiddis What does Validate Gradle Wrapper check? After last commit that check failed

dbwiddis (approved):
LGTM!

build.gradle
```diff
@@ -13,7 +13,7 @@ import java.nio.file.Files
 import org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestFramework
 
 plugins {
-    id 'java'
+    id 'java-library'
```
owaiskazi19, line 16: Any specific reason to change the plugin here?

nassipkali: > Any specific reason to change the plugin here?

api() doesn't work if java-library not imported.

build.gradle
```diff
@@ -90,13 +90,13 @@ dependencies {
     def junitPlatform = "1.9.3"
     def jaxbVersion = "2.3.1"
 
-    implementation("org.opensearch:opensearch:${opensearchVersion}")
+    api("org.opensearch:opensearch:${opensearchVersion}")
```
owaiskazi19, line 93: `api` tends to make the internal libraries as public whereas `implementation` won't let us access the internal libraries by considering them `private`. I don't understand the reason for this change.

owaiskazi19: Okay. I see #609

build.gradle
dbwiddis: `java` is actually deprecated, `java-library` is the new thing if you have a project intended to be consumed by other projects (which the SDK does).

owaiskazi19 (approved):

Issue #761, opened 2023-05-17
[BUG] Single-node OpenSearch crashes when connecting to an extension on a remote node


dbwiddis:
### What is the bug?

The OpenSearch ExtensionsManager / TransportService does not know its own IP or hostname when initializing extensions.  

We are passing `TransportService.getLocalNode()` to the extension which works fine in local testing on "localhost" and effectively passes the listening Tranport port (9300) but the "local node" is not intended for this purpose, it just creates a short-cut for Transport calls to the local node.  So an extension thinks OpenSearch lives here:
```text
publish_address {127.0.0.1:9300}, bound_addresses {[::1]:9300}, {127.0.0.1:9300}
```

When initializing an extension, it receives the incoming transport just fine, and can respond in the same transport channel, but attempting to send additional transport requests back (like registering REST requests, settings, etc.) fails.   And after some recent changes to support SSL, the resulting exception actually crashes.  The crashing can be fixed.  The underlying failure to communicate is a much gnarlier problem.
```text
[2023-05-16T17:49:39,636][ERROR][o.o.e.ExtensionsManager  ] [runTask-0] Extension initialization failed
org.opensearch.transport.RemoteTransportException: [hello-world][172.31.26.75:4532][internal:discovery/extensions]
Caused by: org.opensearch.transport.ConnectTransportException: [runTask-0][127.0.0.1:9300] connect_exception
...
Caused by: java.io.IOException: Connection refused: 127.0.0.1/127.0.0.1:9300
```
Getting the IP is not trivial:
 - We see an IP from the transport connection but it has gone through NAT and is useless.  
 - Similarly, the OpenSearch node itself doesn't know it's public IP address, either, just a private IP address

Further complicating matters, using the `opensearch-cluster-cdk` to set up the OpenSearch cluster in a "standard" environment produces a single hostname which resolves to the load balancer, and doesn't actually connect to the same extension node that is starting up the extension.

We could hard-code the IP address, but this is dynamic, created during cluster startup, so the extension doesn't know it and if we have multiple cluster manager nodes we aren't exactly sure of it either; but the extension must be started up before OpenSearch so it can be connected to (see #729).  And I'd need to make sure the extension is running in the same private subnet to even access it (possibly a way I can hack around this for performance testing but clearly not a good solution).

### How can one reproduce the bug?

1. Start up an extension on an EC2 node.  (Cloning SDK and running `./gradlew HelloWorld` works.)
[...]

PR #763, opened 2023-05-18
Add default Opensearch host and port if init comes from localhost

dbwiddis:
### Description

Under some circumstances (I believe a single node "cluster") the TransportService identifies its local node at localhost rather than its IP address.  This confuses the ExtensionsInitRequestHandler which has a completely different understanding of localhost, causing it to try to connect to itself on the OpenSearch port.

Since we include an Opensearch host and port in our ExtensionSettings, this uses those as a backup connection string in the event that the information received from OpenSearch is unreliable.

### Issues Resolved

Helps workaround #761 but the problems go deeper than this.

### DIFF
--- a/src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
+++ b/src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
@@ -12,14 +12,22 @@ package org.opensearch.sdk.handlers;
 import org.apache.logging.log4j.LogManager;
 
 import org.apache.logging.log4j.Logger;
+import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.common.transport.TransportAddress;
 import org.opensearch.discovery.InitializeExtensionRequest;
 import org.opensearch.discovery.InitializeExtensionResponse;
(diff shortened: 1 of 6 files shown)


joshpalis (approved):

joshpalis (approved):

src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
```diff
@@ -46,14 +54,33 @@ public ExtensionsInitRequestHandler(ExtensionsRunner extensionsRunner) {
...
+            sourceHost = extensionsRunner.getSettings().get(OPENSEARCH_HOST_SETTING);
+            try {
+                int sourcePort = Integer.valueOf(extensionsRunner.getSettings().get(OPENSEARCH_PORT_SETTING));
+                sourceNode = new DiscoveryNode(
+                    sourceNode.getName(),
+                    new TransportAddress(InetAddress.getByName(sourceHost), sourcePort),
+                    sourceNode.getVersion()
+                );
+            } catch (NumberFormatException | UnknownHostException e) {
+                // If configured defaults don't work, just keep original source node
```
dbwiddis, line 73: I think missing testing a no-op catch block is acceptable. :)

joshpalis (approved):

ryanbogan (approved):

PR #768, opened 2023-05-20
Refactored all sendFooRequest(TransportService transportService) call…

### LINKED ISSUE #585: [FEATURE] Refactor all sendFooRequest(TransportService transportService) calls into SDKTransportService

While `SDKClient` handles REST calls to OpenSearch, the Proxy action implementation (see #525) requires transport action calls to OpenSearch, and thus require `TransportService` and `OpenSearchNode` objects.

The TransportService is initialized as part of the `run()` method _after_ instantiation.

The `OpenSearchNode` isn't initialized until the ExtensionsRunner connects to OpenSearch.

Chicken, meet egg.

We need some sort of holder class that can be sent to the actions during ExtensionsRunner initialization and dependency injection.  As part of work on #525 I've currently written an `SDKTransportService` class that works nicely for this purpose.  With a setter, it can hold the (initialized) TransportService and OpenSearchNode (and UniqueId, too!) and greatly simplify these calls, particularly:
 - the only reason `transportService` is included as a parameter is for unit tests.  That can be removed and tests can access the setter on the wrapper class to serve the same purpose.
 - there's no need to pass additional parameters like the OpenSearchNode or UniqueId that some of these calls need.
 - we could completely eliminate the `getExtensionTransportService()` getter, as well as getters for the OpenSearchNode and UniqueId, which could be moved to this class (they are rarely used outside these transport calls).

I tried some other existing classes as the wrapper
 - SDKActionModule and ExtensionsRunner suffered from the same chicken and egg problem.  
 - SDKClient worked, but I like the separation of purpose, SDKClient for REST, SDKTransportService for transport

Posting this request now, but it will touch a lot of classes (including tests) and would create big conflicts if overlapping in-progress work.  It should wait for completion of:
 - PR (TBD) implementing #525 which will create this class
 - PR (TBD) implementing #17 which is working with initialization classes and would conflict with any new changes

nassipkali:
### Description
Refactored all sendFooRequest(TransportService transportService) calls into SDKTransportService

### Issues Resolved
#585

### DIFF
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -19,7 +19,6 @@ import org.opensearch.action.ActionType;
 import org.opensearch.action.support.TransportAction;
 import org.opensearch.cluster.ClusterState;
 import org.opensearch.cluster.metadata.IndexNameExpressionResolver;
-import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.common.io.stream.NamedWriteableRegistry;
 import org.opensearch.extensions.proto.ExtensionRequestProto;
 import org.opensearch.extensions.rest.ExtensionRestRequest;
@@ -96,16 +95,10 @@ public class ExtensionsRunner {
     // Set when initialization is complete
     private boolean initialized = false;
 
-    private String uniqueId;
     /**
      * This field is initialized by a call from {@link ExtensionsInitRequestHandler}.
      */
-    public DiscoveryNode opensearchNode;
     private DiscoveryExtensionNode extensionNode;
-    /**
-     * This field is initialized by a call from {@link ExtensionsInitRequestHandler}.
-     */
-    private TransportService extensionTransportService = null;
 
     // The routes and classes which handle the REST requests
     private final ExtensionRestPathRegistry extensionRestPathRegistry = new ExtensionRestPathRegistry();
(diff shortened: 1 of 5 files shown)


dbwiddis (commented):
Good start here, but the main intent is to actually move all the methods themselves into the SDKTransportService class.

nassipkali:
@dbwiddis I removed argument capture tests because mock() can't applied to subclass(SDKTransportService).

src/main/java/org/opensearch/sdk/handlers/ClusterSettingsResponseHandler.java
```diff
@@ -13,16 +13,14 @@
...
 import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.transport.TransportException;
 import org.opensearch.transport.TransportResponseHandler;
-import org.opensearch.transport.TransportService;
 
 import java.io.IOException;
 
 /**
- * This class handles the response from OpenSearch to a {@link ExtensionsRunner#sendClusterSettingsRequest(TransportService)} call.
+ * This class handles the response from OpenSearch to a {@link org.opensearch.sdk.SDKTransportService#sendClusterSettingsRequest()} call.
```
dbwiddis, line 23: This is fine, but I think it would be more readable if you made `SDKTransportService` an import instead of putting the fully qualified class name here (and a few other places in this PR)

src/main/java/org/opensearch/sdk/SDKTransportService.java
```diff
@@ -107,6 +127,195 @@ public RemoteExtensionActionResponse sendRemoteExtensionActionRequest(RemoteExte
...
+    private void sendGenericRequestWithExceptionHandling(
+        ExtensionRequestProto.RequestType requestType,
+        String orchestratorNameString,
+        TransportResponseHandler<? extends TransportResponse> responseHandler
+    ) {
+        logger.info("Sending " + requestType + " request to OpenSearch");
+        try {
+            transportService.sendRequest(opensearchNode, orchestratorNameString, new ExtensionRequest(requestType), responseHandler);
+        } catch (Exception e) {
+            logger.info("Failed to send " + requestType + " request to OpenSearch", e);
```
dbwiddis, line 186: We should probably change these (a few others in this class) to `.error()` log level.

dbwiddis (changes requested):
Looks good, a merge conflict and a few improvements to make!

dbwiddis (approved):

owaiskazi19 (approved):
LGTM

PR #799, opened 2023-06-01
Revert "Add integration test framework"

owaiskazi19:
Reverts opensearch-project/opensearch-sdk-java#687


joshpalis (approved):
We can remove the backport 1.x label since the original backport isn't merged yet

PR #827, opened 2023-06-20
Register new routes via SDK as named routes

DarshitChanpura:
### Description
Continuation of #622 
This change allows registering extension routes as NamedRoute (new, deprecated, replaced), and updates sample helloworld extension to conform to the new route registration mechanism to optionally add legacy permission actions (i.e. `cluster:admin/opensearch/.....`)

Related PRs:
- Blocked by: https://github.com/opensearch-project/OpenSearch/pull/7957
- https://github.com/opensearch-project/security/pull/2753


DarshitChanpura:
Gradle check is failing as it needs core artifact with changes from this: https://github.com/opensearch-project/OpenSearch/pull/7957 merged

DEVELOPER_GUIDE.md
```diff
@@ -162,18 +162,20 @@ To **run OpenSearch from Gradle**, follow these steps:
...
+"hostAddress":"127.0.0.1",
+"port":"4532",
+"version":"1.0",
+"opensearchVersion":"3.0.0",
+"minimumCompatibleVersion":"3.0.0",
+"dependencies":[{"uniqueId":"test1","version":"2.0.0"},{"uniqueId":"test2","version":"3.0.0"}]
 }'
 ```
 
+Note: If Security plugin is initialized in OpenSearch, use admin credentials to send extension initialization request.
```
dbwiddis, line 177:
```suggestion
Note: If the Security plugin is initialized in OpenSearch, use admin credentials to send the extension initialization request.
```

dbwiddis (commented):
LGTM with a few documentation nits. 

Also didn't comment inline, but I'm not sure rewriting temporary Deprecated and Replaced route handlers is the right approach.  I understand we will eventually incorporate this functionality into the NamedRoutes, correct?

DarshitChanpura:
@dbwiddis I've removed references to named handlers for deprecated and replaced routes.

dbwiddis (dismissed):

PLUGIN_MIGRATION.md
```diff
@@ -67,14 +67,37 @@ XContentParser parser = XContentType.JSON
...
+```
+private Function<RestRequest, RestResponse> uriHandler = () -> {};
+public List<NamedRoute> routes() {
+    return ImmutableList.of(
+            new NamedRoute.Builder().method(GET).path("/uri").uniqueName("extension:uri").handler(uriHandler).build()
+        );
+}
+```
+
+You can optionally also add `actionNames()` to this route. These should correspond to any current actions defined as permissions in roles.
```
cwperks, line 97: Do you think its worth mentioning here that `actionNames` are useful for plugins being converted to extensions and wanting to keep backward compatibility with existing reserved roles that they have defined?

src/main/resources/sample/helloworld-settings.yml
```diff
@@ -1,4 +1,5 @@
 extensionName: hello-world
+shortExtensionName: hw
```
cwperks, line 2: Is `shortExtensionName` being used? I know we had talked about a separate values for extension unique id and an abbreviation for an extension, but is the extension unique id now also being treated as an abbreviation that would be prefixed to all action names? i.e. The route `_extensions/_hw/greet` is assigned the permissionable action name `hw:greet` and `hw` is the extension unique id

src/main/java/org/opensearch/sdk/rest/ExtensionRestPathRegistry.java
```diff
@@ -171,9 +190,18 @@ public List<String> getRegisteredDeprecatedPaths() {
...
      * @return A string appending the method and path.
      */
-    public static String restPathToString(Method method, String path) {
-        return method.name() + " " + path;
+    public static String restPathToString(Method method, String path, String name, Set<String> actionNames) {
+        StringBuilder sb = new StringBuilder(method.name() + " " + path + " ");
+        if (name != null && !name.isEmpty()) {
+            sb.append(name).append(" ");
+        }
+        if (actionNames != null) {
```
cwperks, line 202: Is it enforced anywhere that `actionNames` cannot contain spaces? If space is being used as a delimiter than an `actionName` must not contain a space

src/main/resources/sample/helloworld-settings.yml
cwperks: Ok, I see now that extension name !== extension unique id. That sounds fine to me.

cwperks (commented):
Thank you @DarshitChanpura . Took a pass and this looks good to me. There's one reference to `shortExtensionName` that looks like it can be removed.

src/main/java/org/opensearch/sdk/sample/helloworld/rest/RestHelloAction.java
```diff
@@ -50,24 +55,49 @@ public class RestHelloAction extends BaseExtensionRestHandler {
...
+    public List<NamedRoute> routes() {
         return List.of(
-            new RouteHandler(GET, "/hello", handleGetRequest),
-            new RouteHandler(POST, "/hello", handlePostRequest),
-            new RouteHandler(PUT, "/hello/{name}", handlePutRequest),
-            new RouteHandler(DELETE, "/goodbye", handleDeleteRequest)
+            new NamedRoute.Builder().method(GET)
+                .path("/hello")
+                .handler(handleGetRequest)
+                .uniqueName(routePrefix("greet"))
```
cwperks, line 69: One last comment, it looks like an extensions developer has the ability to pass any name as unique name and conventionally they would opt to use this `routePrefix`. I'm wondering if it should be stronger than that and that all names for routes with an extension must be prefixed with the extension's name/abbreviation?

cwperks (approved):
This PR looks good to me. I left a few comments in a review earlier, but none that require to be addressed before this PR is merged.

src/main/java/org/opensearch/sdk/rest/ExtensionRestPathRegistry.java
dbwiddis: I had this same question and [yes, it is](https://github.com/DarshitChanpura/OpenSearch/blob/bea26f97c66070581070764fe4790d2f0579c8e7/server/src/main/java/org/opensearch/rest/NamedRoute.java#L27).

CREATE_YOUR_FIRST_EXTENSION.md
```diff
@@ -164,28 +164,28 @@ import org.opensearch.sdk.rest.BaseExtensionRestHandler;
 public class CrudAction extends BaseExtensionRestHandler {
 
     @Override
-    protected List<RouteHandler> routeHandlers() {
+    protected List<NamedRoute> routes() {
```
dbwiddis, line 167: This needs to be `public` because of the public method it's inheriting.  Don't hold up PR approval for this, I'll address this in a follow-up doc change.

src/main/resources/sample/helloworld-settings.yml
DarshitChanpura: removed the reference.

CREATE_YOUR_FIRST_EXTENSION.md
DarshitChanpura: Done.

PLUGIN_MIGRATION.md
DarshitChanpura: done.

src/main/java/org/opensearch/sdk/sample/helloworld/rest/RestHelloAction.java
DarshitChanpura: I think we should leave this responsibility to extensions developer. We can maybe in a future iteration enforce strict checks to validate the prefixes with extension name.

DarshitChanpura:
@dbwiddis Would you please re-review? 

The build seems to broken due to this change: https://github.com/opensearch-project/OpenSearch/commit/52a5e3f6e0ca599e3193807134ea42660ecdd195

DarshitChanpura:
> @dbwiddis Would you please re-review?
> 
> The build seems to broken due to this change: [opensearch-project/OpenSearch@52a5e3f](https://github.com/opensearch-project/OpenSearch/commit/52a5e3f6e0ca599e3193807134ea42660ecdd195)

Added a fix.

dbwiddis:
> > @dbwiddis Would you please re-review?
> > The build seems to broken due to this change: [opensearch-project/OpenSearch@52a5e3f](https://github.com/opensearch-project/OpenSearch/commit/52a5e3f6e0ca599e3193807134ea42660ecdd195)
> 
> Added a fix.

Yep, I'll re-review and thank you for fixing that!  I was playing around off and on fixing up the sample extension that this documents, and encountered that same problem. I just deleted the offending lines. ;)

BTW, updated sample at https://github.com/dbwiddis/CRUDExtension I am working on updating the docs starting with your version in this PR

dbwiddis (dismissed):
LGTM.  

I have updated the "first extension" that the CREATE_YOUR_FIRST_EXTENSION doc describes and will be updating that doc file in a separate PR shortly.   See https://github.com/dbwiddis/CRUDExtension

build.gradle
```diff
@@ -189,16 +189,20 @@ dependencies {
...
+            force("com.fasterxml.jackson.core:jackson-core:${jacksonDatabindVersion}")
+            force("com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:${jacksonDatabindVersion}")
+            force("com.fasterxml.jackson.dataformat:jackson-dataformat-smile:${jacksonDatabindVersion}")
+            force("com.fasterxml.jackson.dataformat:jackson-dataformat-cbor:${jacksonDatabindVersion}")
+            force("org.apache.logging.log4j:log4j-api:${log4jVersion}")
+            force("org.apache.logging.log4j:log4j-core:${log4jVersion}")
+            force("org.apache.logging.log4j:log4j-jul:${log4jVersion}")
+            force("org.slf4j:slf4j-api:${slf4jVersion}")
+            force("org.apache.httpcomponents.core5:httpcore5:5.2.2")
+            force("org.apache.httpcomponents.core5:httpcore5-h2:5.2.2")
```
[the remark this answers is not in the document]
DarshitChanpura: done.

DarshitChanpura:
@dbwiddis @owaiskazi19 Would you please review this again?

src/main/java/org/opensearch/sdk/rest/ExtensionRestPathRegistry.java
```diff
@@ -85,7 +104,7 @@ private void registerHandler(Method method, String path, ExtensionRestHandler ex
     private void registerAsDeprecatedHandler(Method method, String path, ExtensionRestHandler handler, String deprecationMessage) {
         assert (handler instanceof ExtensionDeprecationRestHandler) == false;
 
-        registerHandler(method, path, new ExtensionDeprecationRestHandler(handler, deprecationMessage, deprecationLogger));
+        registerHandler(method, path, null, null, new ExtensionDeprecationRestHandler(handler, deprecationMessage, deprecationLogger));
```
owaiskazi19, line 107: Nit: Instead of passing null here we can have a separate method for  `registerHandler` without `name` and `actionNames.`

owaiskazi19 (approved):

dbwiddis (approved):

dbwiddis:
@DarshitChanpura can you please create a backport PR?

joshpalis:
@DarshitChanpura Can you confirm that the hello world extension can be started with these new changes. I am seeing errors when running `./gradlew helloWorld`

```
19:15:52.052 [main] DEBUG org.opensearch.threadpool.ThreadPool - created thread pool: name [write], size [8], queue size [10k]
19:15:52.052 [main] DEBUG org.opensearch.threadpool.ThreadPool - created thread pool: name [snapshot], core [1], max [4], keep alive [5m]
19:15:52.053 [main] DEBUG org.opensearch.threadpool.ThreadPool - created thread pool: name [search_throttled], size [1], queue size [100]
Exception in thread "main" OpenSearchException[Invalid route name specified. The route name may include the following characters 'a-z', 'A-Z', '0-9', ':', '/', '*', '_' and be less than 250 characters]
        at org.opensearch.rest.NamedRoute.<init>(NamedRoute.java:143)
        at org.opensearch.rest.NamedRoute$Builder.build(NamedRoute.java:112)
        at org.opensearch.sdk.sample.helloworld.rest.RestHelloAction.routes(RestHelloAction.java:71)
        at org.opensearch.sdk.rest.ExtensionRestPathRegistry.registerHandler(ExtensionRestPathRegistry.java:43)
        at org.opensearch.sdk.ExtensionsRunner.<init>(ExtensionsRunner.java:240)
        at org.opensearch.sdk.ExtensionsRunner.run(ExtensionsRunner.java:583)
        at org.opensearch.sdk.sample.helloworld.HelloWorldExtension.main(HelloWorldExtension.java:82)
```

dbwiddis:
> @DarshitChanpura Can you confirm that the hello world extension can be started with these new changes. I am seeing errors when running `./gradlew helloWorld`

The problem is the route prefix code:
```java
protected static String routePrefix(String name) {
    return extensionName + ":" + name;
}
```
It uses the extension name without checking it; the hello world extension name (per the .yml file) is 
```yaml
extensionName: hello-world
```
The `-` is not allowed.

DarshitChanpura:
Aah. I've tested it using the short hand `hw` and didn't run into this error. I can create a PR to rename this to `hello_world` which will work.

owaiskazi19:
> Aah. I've tested it using the short hand `hw` and didn't run into this error. I can create a PR to rename this to `hello_world` which will work.

The correct way of adding a route would be `helloWorld` rather with a `-` or `_`. Can you rename it with the same @DarshitChanpura?

DarshitChanpura:
@dbwiddis @joshpalis This PR should resolve the issue. https://github.com/opensearch-project/opensearch-sdk-java/pull/868

Issue #815, opened 2023-06-08
[BUG] Extension should not bind to its hostname


dbwiddis:
### What is the bug?

During initialization, an extension binds to its port on the network interface associated with the host name:
https://github.com/opensearch-project/opensearch-sdk-java/blob/5fed5642c9f728d9d67099a0a38cba58a419d968/src/main/java/org/opensearch/sdk/ExtensionsRunner.java#L162

It is possible for an extension to be listening on multiple interfaces. In particular, the localhost interface will be different than an externally-facing one; it's also possible to have a different interface associated with private and public IP addresses.  These may not be known in advance when starting up the extension.

### How can one reproduce the bug?

1. Start up an extension such as Hello World on a remote node with the default settings (127.0.0.1).
2. Attempt to connect to the extension on its configured port.

### What is the expected behavior?

Connection established on its listening port.

### Do you have any additional context?

On OpenSearch, there are separate settings for `BIND_HOST` and `PUBLISH_HOST`.  There is no assigned host/IP for `BIND_HOST`:
```java
    public static final Setting<List<String>> BIND_HOST = listSetting(
        "transport.bind_host",
        HOST,
        Function.identity(),
        Setting.Property.NodeScope
    );
```
where
```java
    public static final Setting<List<String>> HOST = listSetting(
        "transport.host",
        emptyList(),
        Function.identity(),
        Setting.Property.NodeScope
    );
```

The extension should probably set `BIND_HOST` to this same `HOST` (essentially an empty list), or bind to "0.0.0.0" which listens on all network interfaces.  We could/should also set `PUBLISH_HOST` to the existing value from the settings.

tarunsamanta2k20:
I'd like to work on this issue

dbwiddis:
Hey @tarunsamanta2k20 are you still working on this?

tarunsamanta2k20:
@dbwiddis could you please suggest me how to do it?

tarunsamanta2k20:
@dbwiddis I am interested to work on it. Can I do pairing for this issue?

dbwiddis:
The line of code you need to fix is documented above.  Change `BIND_HOST` to `PUBLISH_HOST` in that line, and add a new similar line for `BIND_HOST` which uses the builder's `putList(String setting, List<String> values)` method with an empty list.

tarunsamanta2k20:
I couldn't set the second argument like that as it expects a path, not a list.
**Collections.emptyList()**

[...]

PR #830, opened 2023-06-21
fix(bug): Extension should not bind to its hostname

tarunsamanta2k20:
Closes: #815 

What is the bug?

During initialization, an extension binds to its port on the network interface associated with the host name:

[opensearch-sdk-java/src/main/java/org/opensearch/sdk/ExtensionsRunner.java](https://github.com/opensearch-project/opensearch-sdk-java/blob/5fed5642c9f728d9d67099a0a38cba58a419d968/src/main/java/org/opensearch/sdk/ExtensionsRunner.java#L162)

```
Line 162 in [5fed564](https://github.com/opensearch-project/opensearch-sdk-java/commit/5fed5642c9f728d9d67099a0a38cba58a419d968)
```

It is possible for an extension to be listening on multiple interfaces. In particular, the localhost interface will be different than an externally-facing one; it's also possible to have a different interface associated with private and public IP addresses. These may not be known in advance when starting up the extension.
How can one reproduce the bug?

    Start up an extension such as Hello World on a remote node with the default settings (127.0.0.1).
    Attempt to connect to the extension on its configured port.

What is the expected behavior?

Connection established on its listening port.
Do you have any additional context?

On OpenSearch, there are separate settings for BIND_HOST and PUBLISH_HOST. There is no assigned host/IP for BIND_HOST:

    public static final Setting<List<String>> BIND_HOST = listSetting(
        "transport.bind_host",
        HOST,
        Function.identity(),
        Setting.Property.NodeScope
    );

where

    public static final Setting<List<String>> HOST = listSetting(
        "transport.host",
        emptyList(),
        Function.identity(),
        Setting.Property.NodeScope
    );

The extension should probably set BIND_HOST to this same HOST (essentially an empty list), or bind to "0.0.0.0" which listens on all network interfaces. We could/should also set PUBLISH_HOST to the existing value from the settings.

### DIFF
--- a/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
+++ b/src/main/java/org/opensearch/sdk/ExtensionsRunner.java
@@ -159,7 +159,8 @@ public class ExtensionsRunner {
         ExtensionSettings extensionSettings = extension.getExtensionSettings();
         Settings.Builder settingsBuilder = Settings.builder()
             .put(NODE_NAME_SETTING, extensionSettings.getExtensionName())
-            .put(TransportSettings.BIND_HOST.getKey(), extensionSettings.getHostAddress())
+            .put(TransportSettings.PUBLISH_HOST.getKey(), extensionSettings.getHostAddress())
+            .putList(TransportSettings.BIND_HOST.getKey(), Collections.emptyList())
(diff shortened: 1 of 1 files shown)


dbwiddis (approved):
LGTM!

joshpalis (approved):

ryanbogan (approved):

Issue #857, opened 2023-07-04
[BUG] Build fails due to transitive dependencies


dbwiddis:
### What is the bug?

Following merge of https://github.com/opensearch-project/OpenSearch/pull/8434, transitive dependencies between OpenSearch and `opensearch-java` conflict for Apache Http dependencies.
```
Execution failed for task ':compileJava'.
> Could not resolve all dependencies for configuration ':compileClasspath'.
   > Conflict(s) found for the following module(s):
       - org.apache.httpcomponents.client5:httpclient5 between versions 5.2.1 and 5.1.4
       - org.apache.httpcomponents.core5:httpcore5 between versions 5.2.2, 5.1.5 and 5.2
       - org.apache.httpcomponents.core5:httpcore5-h2 between versions 5.2.2, 5.1.5 and 5.2
```

### How can one reproduce the bug?

Build latest SDK against latest Opensearch and OpenSearch Client for Java

### What is the expected behavior?

Code compiles

### Do you have any additional context?

We need to either exclude the transitive dependency or force our own version.

owaiskazi19:
This should be fixed by: https://github.com/opensearch-project/opensearch-java/pull/554

dbwiddis:
Yeah, and we'll break again next time one updates before the other.

cwperks:
I'm still getting this issue even after the merge of the change in `opensearch-java` and publishing the latest changes to maven local.

```
> ./gradlew :dependencyInsight --configuration compileClasspath --dependency org.apache.httpcomponents.core5:httpcore5

org.apache.httpcomponents.core5:httpcore5:5.2.2
+--- org.apache.httpcomponents.core5:httpcore5-h2:5.2.2
|    +--- org.opensearch.client:opensearch-rest-client:3.0.0-SNAPSHOT
|    |    +--- compileClasspath
|    |    \--- org.opensearch.client:opensearch-rest-high-level-client:3.0.0-SNAPSHOT
|    |         \--- compileClasspath
|    +--- org.opensearch.client:opensearch-java:3.0.0-SNAPSHOT
|    |    \--- compileClasspath
|    \--- org.apache.httpcomponents.client5:httpclient5:5.2.1 (requested org.apache.httpcomponents.core5:httpcore5-h2:5.2)
|         +--- org.opensearch.client:opensearch-rest-client:3.0.0-SNAPSHOT (*)
[...]

PR #861, opened 2023-07-06
Build fails due to transitive dependencies

reta:
### Description
Build fails due to transitive dependencies

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-sdk-java/issues/857

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -154,7 +154,10 @@ dependencies {
     implementation("org.apache.logging.log4j:log4j-jul:${log4jVersion}")
     implementation("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
     api("org.opensearch.client:opensearch-rest-client:${opensearchVersion}")
-    api("org.opensearch.client:opensearch-java:${opensearchVersion}")
+    api("org.opensearch.client:opensearch-java:${opensearchVersion}") {
+      exclude group : 'org.apache.httpcomponents.core5'
+      exclude group : 'org.apache.httpcomponents.client5'
+    }
     implementation("org.opensearch.plugin:transport-netty4-client:${opensearchVersion}")
     implementation("io.netty:netty-all:${nettyVersion}")
     implementation("jakarta.inject:jakarta.inject-api:${jakartaInjectVersion}")


build.gradle
```diff
@@ -154,7 +154,10 @@ dependencies {
     implementation("org.apache.logging.log4j:log4j-jul:${log4jVersion}")
     implementation("org.opensearch.client:opensearch-rest-high-level-client:${opensearchVersion}")
     api("org.opensearch.client:opensearch-rest-client:${opensearchVersion}")
-    api("org.opensearch.client:opensearch-java:${opensearchVersion}")
+    api("org.opensearch.client:opensearch-java:${opensearchVersion}") {
```
reta, line 157: The Apache HttpClient5 are coming from `opensearch-rest-client` and  `opensearch-java`, to disambiguate any conflicts, I would suggest to rely on `opensearch-rest-client`

owaiskazi19: Looks like `httpclient5:5.2.1` is bringing `httpcore5:5.2` instead of `5.2.2` causing the conflict in the first place.
```
org.apache.httpcomponents.core5:httpcore5:5.2 -> 5.2.2
\--- org.apache.httpcomponents.client5:httpclient5:5.2.1
     +--- org.opensearch.client:opensearch-rest-client:3.0.0-SNAPSHOT
     |    +--- runtimeClasspath
     |    \--- org.opensearch.client:opensearch-rest-high-level-client:3.0.0-SNAPSHOT
     |         \--- runtimeClasspath
     \--- org.opensearch.client:opensearch-java:3.0.0-SNAPSHOT
          \--- runtimeClasspath
```

reta: This is quite strange, I will double check that but the deps are consistent (in Gradle at least):

```
  api "org.apache.httpcomponents.client5:httpclient5:${versions.httpclient5}"
  api "org.apache.httpcomponents.core5:httpcore5:${versions.httpcore5}"
  api "org.apache.httpcomponents.core5:httpcore5-h2:${versions.httpcore5}"
```

owaiskazi19: > Btw, why do we need 2 clients here?

That's a good question. When we started building SDK we initially had java client which was causing a bunch of issues for the migration of a plugin to an extension. Later, on we moved to HLRC.
There's an open issue https://github.com/opensearch-project/opensearch-sdk-java/issues/732 discussing about redesigning the SDKClient and just have HLRC here. Java Client is something extension can include in their own build.gradle.

reta: Just for the sake of it (you know that), `opensearch-java` is what we recommend and it also has no dependencies on any modules from OpenSearch core (in 3.x release line)

owaiskazi19: Yes, that would be the right client to go about it. It's just all the plugins are written based on NodeClient, converting those parsers and others would be a task while migrating from a plugin to an extension.

dbwiddis (approved):

build.gradle
dbwiddis: > Just for the sake of it (you know that), `opensearch-java` is what we recommend and it also has no dependencies on any modules from OpenSearch core (in 3.x release line)

Yep, long discussion in #294 as Owais mentioned, and also specific plans in #302.

The idea to remove the clients as an API dependency is associated with multithreading and resource leakage concerns.

owaiskazi19 (approved):

owaiskazi19:
Thanks @reta!

reta:
Folks, backport to `1.x` may not be need, the `opensearch-java` does not bring Apache HttpClient5 / HttpCore5 by default - this is opt-in only

PR #866, opened 2023-07-10
[Backport 1.x] Register new routes via SDK as named routes (#827) and Makes route prefix setting routeNamePrefix optional #868

DarshitChanpura:
### Description
Manually Backports #827 & #868

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #868, opened 2023-07-10
Makes route prefix setting `routeNamePrefix` optional

DarshitChanpura:
### Description

### Issues Resolved
https://github.com/opensearch-project/opensearch-sdk-java/pull/827#issuecomment-1629622060


owaiskazi19 (approved):

joshpalis (approved):

dbwiddis (changes requested):
While renaming will "fix" the sample, it doesn't solve the underlying problem:
1. There is no documentation to users on what is a valid name (and the OpenSearch error message isn't helpful enough to debug the problem without referencing the source code)
2. It disallows the existing names for every plugin, many of which we would migrate, e.g., `opensearch-alerting`, `opensearch-anomaly-detection`, `opensearch-knn`, `opensearch-ml`, etc.

The name here isn't actually used for anything programmatic.  It's purely decorative and is only used as the response to the initialization command where it is displayed in a log message, and as the "node name".  Node names in OpenSearch use hyphens.

Our test cases still use `sample-extension` as well, and don't fail on an illegal name there.  If we're really set on disallowing a hyphen those need to validate the names.

I strongly prefer one of the following two options:

1. Simply remove the `routePrefix()` method and replace it with a constant in each REST handler, or a local method in a REST handler that uses an allowed prefix
2. If you keep the route prefix method, add a `setRoutePrefix()` method that validates it's a legal prefix.

DarshitChanpura:
> While renaming will "fix" the sample, it doesn't solve the underlying problem:
> 
> 1. There is no documentation to users on what is a valid name (and the OpenSearch error message isn't helpful enough to debug the problem without referencing the source code)
> 2. It disallows the existing names for every plugin, many of which we would migrate, e.g., `opensearch-alerting`, `opensearch-anomaly-detection`, `opensearch-knn`, `opensearch-ml`, etc.
> 
> The name here isn't actually used for anything programmatic. It's purely decorative and is only used as the response to the initialization command where it is displayed in a log message, and as the "node name". Node names in OpenSearch use hyphens.
> 
> Our test cases still use `sample-extension` as well, and don't fail on an illegal name there. If we're really set on disallowing a hyphen those need to validate the names.
> 
> I strongly prefer one of the following two options:
> 
> 1. Simply remove the `routePrefix()` method and replace it with a constant in each REST handler, or a local method in a REST handler that uses an allowed prefix
> 2. If you keep the route prefix method, add a `setRoutePrefix()` method that validates it's a legal prefix.

I agree. However routePrefix only comes into play if an extension developer decides to use it while registering a route.

i.e. HelloWorld extension can also be registered without calling routePrefix method.
```
new NamedRoute.Builder().method(GET)
                .path("/hello")
                .handler(handleGetRequest)
                .uniqueName(routePrefix("greet"))   -----> can be written as `.uniqueName("greet")` OR `.uniqueName("helloWorld:greet")`
                .legacyActionNames(Set.of("cluster:admin/opensearch/hw/greet"))
                .build()
```

This would still work. `routePrefix` is completely optional

DarshitChanpura:
Added a validation check when extensionName is set to be later used by `routePrefix`

joshpalis (approved):

DarshitChanpura:
@dbwiddis I've update the routePrefix setting to be completely optional in extension-settings.yml

DEVELOPER_GUIDE.md
```diff
@@ -303,6 +303,10 @@ The artifact will include extension settings for the sample Hello World extensio
   opensearchPort: 9200
 ```
 
+You can optionally add `routeNamePrefix:` as a value to the yml. This setting allows you to prefix all you registered NamedRoute names.
```
dbwiddis, line 306:
```suggestion
You can optionally add `routeNamePrefix:` as a value to the yml. This setting allows you to prefix all your registered NamedRoute unique names.
```

DEVELOPER_GUIDE.md
```diff
@@ -303,6 +303,10 @@ The artifact will include extension settings for the sample Hello World extensio
...
+The value must be alphanumeric and can contain `:`, `/`, `*`, `_` in the name.
+
```
dbwiddis, line 308:
```suggestion
```

src/main/java/org/opensearch/sdk/rest/BaseExtensionRestHandler.java
```diff
@@ -76,17 +80,31 @@ protected List<ReplacedRouteHandler> replacedRouteHandlers() {
...
-        BaseExtensionRestHandler.extensionName = extensionName;
+    /**
+     * Sets the route prefix that can be used to prepend route names
+     * @param prefix the prefix to be used
+     */
+    public void setRouteNamePrefix(String prefix) {
+        // we by-pass null assignment as routeNamePrefixes are not mandatory
+        if (prefix != null && !prefix.matches(VALID_ROUTE_PREFIX_PATTERN)) {
+            throw new OpenSearchException(
+                "Invalid extension name specified. The extension name may include the following characters"
```
dbwiddis, line 91:
```suggestion
                "Invalid route prefix specified. The prefix may include the following characters"
```

src/main/java/org/opensearch/sdk/rest/BaseExtensionRestHandler.java
```diff
@@ -76,17 +80,31 @@ protected List<ReplacedRouteHandler> replacedRouteHandlers() {
...
-     * @param name The human-readable name for a route registered by this extension
-     * @return Returns a name prepended with the extension's name
+     * Generates a name for the handler prepended with the route prefix
+     * @param routeName The human-readable name for a route registered by this extension
+     * @return Returns a name conditionally prepended with the valid route prefix
      */
-    protected static String routePrefix(String name) {
-        return extensionName + ":" + name;
+    protected String addRouteNamePrefix(String routeName) {
+        if (routeNamePrefix == null || routeNamePrefix.isBlank()) {
```
dbwiddis, line 104: I think there's a function in `Strings` that does this check

src/main/java/org/opensearch/sdk/ExtensionSettings.java
```diff
@@ -127,9 +130,11 @@ public ExtensionSettings(
         String hostPort,
         String opensearchAddress,
         String opensearchPort,
+        String routeNamePrefix,
```
dbwiddis, line 133: Personal preference, I like to have all the constructers with fewer args call the one with more args, adding in the defaults. But I can see that wasn't done here so I'm fine with this as-is.

src/main/java/org/opensearch/sdk/rest/BaseExtensionRestHandler.java
```diff
@@ -41,7 +43,9 @@
  */
 public abstract class BaseExtensionRestHandler implements ExtensionRestHandler {
 
-    private static String extensionName;
+    private static final String VALID_ROUTE_PREFIX_PATTERN = "^[a-zA-Z0-9:/*_]*$";
```
dbwiddis, line 46: Is the equivalent constant in OpenSearch `public` so we can use it here instead of duplicating it?

dbwiddis (commented):
I like this option! Thanks for putting up with my pickiness.

A few nits but will approve when they're taken care of.

src/main/java/org/opensearch/sdk/rest/BaseExtensionRestHandler.java
DarshitChanpura: That constant is `VALID_ACTION_PREFIX` defined in TransportService. This same pattern is defined in NamedRoute but it isn't public.

DEVELOPER_GUIDE.md
DarshitChanpura: there were two new-lines, removed the one below your suggestion.

DarshitChanpura:
@joshpalis @owaiskazi19 @dbwiddis Would you please re-review this?

dbwiddis (dismissed):

DEVELOPER_GUIDE.md
```diff
@@ -303,6 +303,9 @@ The artifact will include extension settings for the sample Hello World extensio
   opensearchPort: 9200
 ```
 
+You can optionally add `routeNamePrefix:` as a value to the yml. This setting allows you to prefix all your registered NamedRoute names.
```
cwperks, line 306: @DarshitChanpura Is this going back to the notion of an abbreviation for an extension? i.e. Anomaly Detection's abbreviation is `ad`? If that's the case, then can we call this `extensionAbbreviation` or similar? 

What does it mean for this to be optional? For security, if an admin wants to grant full permissions to a user to use ad extensions would that be granting them `ad:*` cluster_permissions or something different?

Could the prefix be the abbreviation if it is set and otherwise take the name of the extension? Maybe support for both `ad:*` and `anomaly-detection:*`? FYI I think `:`, `/` and `*` should be forbidden in the abbreviation. These are used as delimiters in (transport) action names

DarshitChanpura: They can use this as a prefix or define their own when registering routes. I'll change the permission to be only alphanumeric and can include '-' and '_'. I prefer to call it a routeNamePrefix because that is what it does. It won't always be an abbreviation. Maybe in future iterations we enforce wrapping all routes with this prefix, but as of now it is left upto the developers whether they would like to use it or not.

dbwiddis: >  I'll change the permission to be only alphanumeric and can include '-' and '_'. 

The `-` will run afoul of the NamedRoute check.

src/main/java/org/opensearch/sdk/rest/BaseExtensionRestHandler.java
```diff
@@ -41,7 +44,9 @@
  */
 public abstract class BaseExtensionRestHandler implements ExtensionRestHandler {
 
-    private static String extensionName;
+    private static final String VALID_ROUTE_PREFIX_PATTERN = "^[a-zA-Z0-9-_]*$";
```
dbwiddis, line 47: 1. Having a hyphen in the prefix is precisely the bug we are trying to fix here.
2. A hyphen inside square brackets in a regex must be the first or last character, or escaped.

src/main/java/org/opensearch/sdk/rest/BaseExtensionRestHandler.java
```diff
@@ -76,17 +81,31 @@ protected List<ReplacedRouteHandler> replacedRouteHandlers() {
...
+    /**
+     * Sets the route prefix that can be used to prepend route names
+     * @param prefix the prefix to be used
+     */
+    public void setRouteNamePrefix(String prefix) {
+        // we by-pass null assignment as routeNamePrefixes are not mandatory
+        if (prefix != null && !prefix.matches(VALID_ROUTE_PREFIX_PATTERN)) {
+            throw new OpenSearchException(
+                "Invalid route name prefix specified. The prefix may include the following characters"
+                    + " 'a-z', 'A-Z', '0-9', ':', '/', '*', '_'"
```
dbwiddis, line 93: This would still disallow the hyphen (which is what we desire)

dbwiddis, line 93: This error message should match the valid route prefix pattern.

src/main/java/org/opensearch/sdk/rest/BaseExtensionRestHandler.java
DarshitChanpura: Hmm. when i test this regex pattern  ^[a-zA-Z0-9-_]*$ on regex101.com. It matches patterns as expected.

Matches: `hello-world123`, `hellWorld123`, `hello_world`
Doesn't match: `hello.world`, `hello/world`

DarshitChanpura: I will remove hyphen for it to be a subset of `^[a-zA-Z0-9:/*_]*$`

dbwiddis: > Hmm. when i test this regex pattern ^[a-zA-Z0-9-_]*$ on regex101.com. It matches patterns as expected.

Technically that is undefined behavior, so Regex101 is free do do whatever it wants to (which may or may not match what the JDK does).  From [the spec](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html):
> The interpretation of range expressions where the ending range point is also the starting range point of a subsequent range expression (for example, "[a-m-o]" ) is undefined.

dbwiddis (approved):

cwperks:
If this is about adding support for hyphen then why not update the pattern [here](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/rest/NamedRoute.java#L27)?

DarshitChanpura:
> @DarshitChanpura I'm still confused, isn't this PR purpose is to add hyphen to the extension name? Since we want to support other plugins like: opensearch-alerting, opensearch-anomaly-detection, opensearch-knn, opensearch-ml.

We are not meddling with extension name any more. This PR talks about a new optional setting `routeNamePrefix` which has to be alphanumeric and can contain `_` only. This lifts the restriction from the extension name setting.

joshpalis (approved):

DarshitChanpura:
Cherry-picking this manually into https://github.com/opensearch-project/opensearch-sdk-java/pull/866

Issue #867, opened 2023-07-10
[BUG] Integration Test Framework broken


saratvemulapalli:
### What is the bug?
```
./gradlew clean build integTest
```
Integ Test framework is broken for extensions and SDK.

- Fix Init sequence for extension integTest
- Fix all tests in `TransportCommunicationIT.java`
- Add new integration tests for HelloWorld extension.
- Add this integration test as part of CI (Github actions)

minalsha:
@joshpalis  is the IT framework fixed now?

PR #874, opened 2023-07-11
Initial Hello World Extension integration test

joshpalis:
### Description
Sets `./gradle check` to depend on `integTest` task. Adds an integration test to send an initialization request from OpenSearch to the hello world extension.

Additionally this PR removes the TransportCommunicationIT test class as this will need to be redone.

### Issues Resolved
https://github.com/opensearch-project/opensearch-sdk-java/issues/867


ryanbogan (dismissed):

ryanbogan:
Looks like it's failing spotless

ryanbogan (approved):

dbwiddis:
So are we testing the other APIs or is this just a "set it up" and we will add it later?

joshpalis:
> So are we testing the other APIs or is this just a "set it up" and we will add it later?

I have the tests written already for each of the REST apis that the hello world extension exposes. Currently the `NamedRoute` modifications have made it impossible to add these tests in as the extension response is not returned.

Therefore I have only opted to add in the initial set up to ensure that the extension initialization is successful, since this API is unaffected by the `NamedRoute` changes

src/test/java/org/opensearch/sdk/sample/helloworld/rest/TestHelloWorldIT.java
```diff
@@ -0,0 +1,134 @@
...
+
+            assertEquals(RestStatus.ACCEPTED, restStatus(response));
+            Map<String, Object> responseMap = entityAsMap(response);
+            String initializationResponse = (String) responseMap.get("success");
+            assertEquals("A request to initialize an extension has been sent.", initializationResponse);
+            initialized = true;
+
+            // Wait for extension settings/rest handlers to complete registration before any subsequent requests
+            Thread.sleep(TimeUnit.SECONDS.toMillis(5));
+        }
```
owaiskazi19, line 63: Instead in setUp, add it as a test since we just have one integ test right now.

joshpalis: I disagree. It is necessary for the extension initialization to occur prior to the execution of any tests, therefore it should be in the `setUp` method as it is the only mechanism to enforce test execution order.

owaiskazi19: 1. We generally add common code in `setUp` which is needed for all the tests. Since we don't have any test available except the initialization there's no point of having setUp right now. For sure, it is needed later when we add more test.
2. setUp should not have a test. We should not have any assert statements in that.

Let's add a test for extension initialization, remove setUp(add a TODO to move the initialization to setUp once we have more integ test)

joshpalis: sure sounds good, ill go ahead and make this change

owaiskazi19 (approved):
LGTM!

ryanbogan (approved):
LGTM! Just a thought going forward (not in this PR): we should create an abstract class or interface that contains protected methods for a lot of the supporting methods required for future integration testing.  For example, initializing the extension, toHttpEntity, makeRequest, etc.

ryanbogan (approved):

owaiskazi19 (approved):

ryanbogan (approved):

dbwiddis (approved):
LGTM!  Great job.

PR #879, opened 2023-07-12
Add support for extension_dn

MaciejMierzwa:
### Description
Add support for extension domain name, similar to how it's done for security plugin. More about that in the task: [[Extensions] Add support for extension_dns list in extensions.yml file#2730](https://github.com/opensearch-project/security/issues/2730) 

### Issues Resolved
[[Extensions] Add support for extension_dns list in extensions.yml file#2730](https://github.com/opensearch-project/security/issues/2730)

Currently the implementation looks like this:
1. During extension initialization via REST call, DNs are passed in the request body: 
```
curl -XPOST -k -u user:password "https://localhost:9200/_extensions/initialize" -H "Content-Type:application/json" --data '{ "name":"hello-world", "uniqueId":"hello-world2", "hostAddress":"127.0.0.1", "port":"4500", "version":"1.0", "opensearchVersion":"3.0.0", "minimumCompatibleVersion":"3.0.0", "dependencies":[{"uniqueId":"test1","version":"2.0.0"},{"uniqueId":"test2","version":"3.0.0"}], "extension_dn": "CN=extension-0.example.com,OU=node,O=node,L=test,DC=de" }'
```
2. Request gets parsed and forwarded to extension node, extension_dn get's sent with InitializeExtensionRequest object
3. At this point on the extension side after SSL handshake the DNs get compared, error is thrown if it's incorrect

PR on openserch-core side: 

I'm in the process of writing tests. Main issue I'm facing right now is that I'm wondering if the overall process looks good. Is the validation logic on the right side? Maybe core side should wait for extension response and then accept/reject call?

### DIFF
--- a/src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
+++ b/src/main/java/org/opensearch/sdk/handlers/ExtensionsInitRequestHandler.java
@@ -17,7 +17,10 @@ import org.opensearch.discovery.InitializeExtensionRequest;
 import org.opensearch.discovery.InitializeExtensionResponse;
 import org.opensearch.sdk.ExtensionsRunner;
 import org.opensearch.sdk.SDKTransportService;
+import org.opensearch.sdk.ssl.DefaultSslKeyStore;
 import org.opensearch.transport.TransportService;
+
+import java.nio.file.Path;
 
 import static org.opensearch.sdk.ExtensionsRunner.NODE_NAME_SETTING;
 
@@ -53,6 +56,8 @@ public class ExtensionsInitRequestHandler {
      */
     public InitializeExtensionResponse handleExtensionInitRequest(InitializeExtensionRequest extensionInitRequest) {
         logger.info("Registering Extension Request received from OpenSearch");
+        validateDNs(extensionInitRequest);
+
         extensionsRunner.getThreadPool().getThreadContext().putHeader("extension_unique_id", extensionInitRequest.getExtension().getId());
         SDKTransportService sdkTransportService = extensionsRunner.getSdkTransportService();
         sdkTransportService.setOpensearchNode(extensionInitRequest.getSourceNode());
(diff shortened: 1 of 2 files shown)


cwperks:
Hey @MaciejMierzwa, validation of the peer certificate's dn (distinguished name) should be wholly encapsulated within the security plugin. This setting helps mTLS between opensearch node <-> extension node be more secure by verifying that the extension's certificate (peer certificate) matches a list of known distinguished names and if not, deny the transport request as coming from an unknown origin.

Take a look at this method [here](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/transport/SecurityRequestHandler.java#L356) which is used on the receiving end of a transport request (SecurityRequestHandler) to determine if a transport request has originated from another node within the cluster (intra-cluster request) or from another cluster than is trusted (cross-cluster request where nodes share same root ca)

That method uses the [DefaultInterClusterRequestEvaluator.isInterClusterRequest](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/transport/DefaultInterClusterRequestEvaluator.java#L81-L176) which compares the peer certificates dn with the static list of `node_dns` in `opensearch.yml` and from the dynamic list of `node_dns` in the security index (sourced from the [`nodes_dn.yml` file](https://opensearch.org/docs/latest/security/configuration/yaml/#nodes_dnyml))

The logic in particular that is interesting is this condition [here](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/transport/DefaultInterClusterRequestEvaluator.java#L99) where it extracts the principal (dn) from the peer certificate and checks if it matches any known node_dns. 

The alternative to matching exactly the distinguished name is for the peer certificate to have a [known OID (Object Identifier)](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/transport/DefaultInterClusterRequestEvaluator.java#L121-L174) and then it doesn't need to match a known node_dn. More details about that can be found [here](https://security.stackexchange.com/a/133327). I am a little weaker on the subject alt names and how to properly issue certs with matching OIDs so that an exact node_dn match does not need to occur. I believe its added for convenience to prevent needed to add new node_dns every time a new certificate has been issued when nodes are added to the cluster and instead as long as they share the OID from the subjectAltName its granted permission to join the cluster.

MaciejMierzwa:
Hi @cwperks thank you for reply. I spent some time debugging the flow. The issue I'm thinking about right now is that during  handshake list of DNs is loaded from file or index. (like you mentioned).  
Current implementation sends this data in json body in request since design moved away from static config so this data is not available during initial transport request. I'm thinking about maybe loading this info into the index config at first stages of calls. For example around this place: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/extensions/ExtensionsManager.java#L294C29-L294C29. Then this info could be used during handshake and during consecutive calls between extension and cluster. 
Let me know what do you think about it

Issue #824, opened 2023-06-16
[BUG] Overriding SDK DEBUG logging configuration is needlessly difficult


dbwiddis:
### What is the bug?

The SDK contains a logging configuration file, `log4j2.xml` in `src/main/resources` which puts it on the class path of all extensions.  It is currently set to DEBUG level logging at the root level and has an extra ERROR level for netty, which we've since removed.

https://github.com/opensearch-project/opensearch-sdk-java/blob/ab2a3038a8c1f0cc266ec4ed23e4f07e0dd80b6f/src/main/resources/log4j2.xml#L11-L17

Extensions importing the SDK end up with this file on the classpath.  This logging configuration is not obvious to them, nor easily accessible without digging through the GitHub source code.  

More importantly, it is more difficult to override locally; users need to figure out how to configure their own config file and pass a command line parameter or set a system property to override the one on the class path.  Simply adding another config file doesn't always work, as the file extensions play a part in which one is chosen first.  Specifically, the priority (which no one ought to be expected to know) is:
1. Value of `log4j2.configurationFile` system property or command line argument.
2. `.properties` file on class path
3. `.yaml` file on the class path
4. `.json` file on the class path
5. `.xml` file on the class path.
6. Default log4j settings (ERROR level).

If there's multiple files and a tie, which one is loaded first is based on class loading order, which in my experience has tended to be the SDK.

### How can one reproduce the bug?

1. Run an extension.
2. Look at the logs, and observe lots of DEBUG statements.

### What is the expected behavior?

Extensions should use the default Log4j2 settings with no configuration file on the class path.

### Do you have any additional context?

Suggested solution:
 - rename the `log4j2.xml` configuration file to one that is not automatically added to the class path
 - update build.gradle or developer guide command line instructions on how to point to the updated config file

One other thought: we are locking extension authors into using Log4J2 as their logger.  We might consider changing our logging to the SLF4J API, which allows extension authors to plug in their own logging framework of choice.  So we could use log4j2 in our own build.gradle for testing but extension authors could use logback, reload4j, slf4j-simple, log4j2, etc.

dbwiddis:
Option 1: Leave logger as-is and document.

1. build.gradle changes:
    - Need to an an exclusion to the Java Client dependency to exclude the slf4j-api transitive dependency.
2. Need to add verbiage to developer guide indicating that we inherit an API dependency on log4j2 from OpenSearch and link to log4j2 documentation on how to configure logging.
[...]

PR #885, opened 2023-07-14
[Extensions] Allow default log4j behavior for users of SDK

vibrantvarun:
### Description
Fix to bug
[824](https://github.com/opensearch-project/opensearch-sdk-java/issues/824#issuecomment-1636226463)

Steps to fix
[https://github.com/opensearch-project/opensearch-sdk-java/issues/824#issuecomment-1636226463](https://github.com/opensearch-project/opensearch-sdk-java/issues/824#issuecomment-1636226463)

### Issues Resolved

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -146,7 +146,7 @@ dependencies {
     def javaxVersion  = "1"
     def guavaFailureAccessVersion = "1.0.1"
     def aopallianceVersion = "1.0"
-    //def slf4jVersion = "1.7.36"
+    def slf4jVersion = "1.7.36"
 
     api("org.opensearch:opensearch:${opensearchVersion}")
     implementation("org.apache.logging.log4j:log4j-api:${log4jVersion}")
@@ -199,10 +199,8 @@ dependencies {
             force("org.apache.logging.log4j:log4j-api:${log4jVersion}")
             force("org.apache.logging.log4j:log4j-core:${log4jVersion}")
             force("org.apache.logging.log4j:log4j-jul:${log4jVersion}")
-            //force("org.slf4j:slf4j-api:${slf4jVersion}")
+            force("org.slf4j:slf4j-api:${slf4jVersion}")
         }
-
-        exclude group: 'org.slf4j'
     }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -146,7 +146,7 @@ dependencies {
     def javaxVersion  = "1"
     def guavaFailureAccessVersion = "1.0.1"
     def aopallianceVersion = "1.0"
-    def slf4jVersion = "1.7.36"
+    //def slf4jVersion = "1.7.36"
 
     api("org.opensearch:opensearch:${opensearchVersion}")
     implementation("org.apache.logging.log4j:log4j-api:${log4jVersion}")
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -146,7 +146,7 @@ dependencies {
     def javaxVersion  = "1"
     def guavaFailureAccessVersion = "1.0.1"
     def aopallianceVersion = "1.0"
-    def slf4jVersion = "1.7.36"
+    //def slf4jVersion = "1.7.36"
```
dbwiddis, line 149: Looks great but just remove instead of commenting :)

build.gradle
```diff
@@ -199,8 +199,10 @@ dependencies {
             force("org.apache.logging.log4j:log4j-api:${log4jVersion}")
             force("org.apache.logging.log4j:log4j-core:${log4jVersion}")
             force("org.apache.logging.log4j:log4j-jul:${log4jVersion}")
-            force("org.slf4j:slf4j-api:${slf4jVersion}")
+            //force("org.slf4j:slf4j-api:${slf4jVersion}")
```
dbwiddis, line 202: ditto

src/main/resources/log4j2-sdk.xml
```diff
@@ -13,6 +13,5 @@
             <AppenderRef ref="ConsoleAppender" />
             <AppenderRef ref="FileAppender"/>
         </Root>i
-    <logger name="io.netty.util" level="ERROR"/>
```
dbwiddis, line 16: Can you also remove the stray `i` in the line above this?

dbwiddis (commented):
Looks good!   Can we fully delete instead of just commenting?

build.gradle
```diff
@@ -199,8 +199,10 @@ dependencies {
...
         }
+
+        exclude group: 'org.slf4j'
```
owaiskazi19, line 205: Is this a transitive dependency coming from some other dependency?

owaiskazi19:
The change is failing the integ test though:

```
 org.opensearch.sdk.sample.helloworld.rest.TestHelloWorldIT > testInitializeExtension FAILED
    java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory

File: build.gradle (hunk 4 above)
Comment (vibrantvarun) [reply build.gradle]: Comment addressed and verified by dan

Comment (dbwiddis) [reply build.gradle]: @owaiskazi19 this turned out to be a gnarlier issue than I thought.  I'll update the issue and create a new one.

TLDR we are keeping both imports as is.  We are just moving the log4j xml file.

**Review by dbwiddis — APPROVED** [review]
LGTM.

This will avoid spamming the debug log of SDK users.

My original jar hell fears are not accurate; those were for log4j1, so there is a distinction in classes.

We still have both api's as transitive dependencies but downstream users can exclude the one they don't want.

**Comment by dbwiddis** [comment]
> The change is failing the integ test though. You can remove the logger from IT test

That logger is log4j.  The problem is in the next line in the stack trace:
```
org.opensearch.sdk.sample.helloworld.rest.TestHelloWorldIT > testInitializeExtension FAILED
    java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory
        at org.apache.hc.client5.http.ssl.AbstractClientTlsStrategy.<clinit>(AbstractClientTlsStrategy.java:64)
```
We are forced to import SLF4J logger here: https://hc.apache.org/httpcomponents-client-5.1.x/current/httpclient5/xref/org/apache/hc/client5/http/ssl/AbstractClientTlsStrategy.html#L64

We are forced to import LOG4J2 logger here:  https://github.com/opensearch-project/OpenSearch/blob/675acdb4b92d862ccec3ebd10f390f333d81291f/server/src/main/java/org/opensearch/common/settings/Settings.java#L857

We can therefore not exclude either API dependency from SDK without either:
 - removing any usage of the Settings class, or
 - removing any usage of Apache Client SSL classes

My original concern of forcing debug logging is resolved by moving the config file. 

Downstream users end up with two logging APIs but they can at least configure them and sort it out themselves.

owaiskazi19 (approved):
Fine with this temp fix!

Issue #887, opened 2023-07-17
[Security for Extensions] Consume access tokens passed from core and utilize in REST Clients


cwperks:
As part of [Security for Extensions](https://gist.github.com/cwperks/e756e1cead72cd511d819241a11337e8#file-security_for_extensions-md) there will be 2 types of tokens ferried to an extension to be utilized to make REST Requests back to the OpenSearch cluster. These 2 types of tokens are:

- On-Behalf-Of tokens - these are short-lived access tokens issued on behalf of the currently authenticated user
- Service Account token - This would be a single token issued and ferried over during extension bootstrap and utilized by the extension to make requests on its own behalf, for instance writing to a system index

The [SDKClient](https://github.com/cwperks/opensearch-sdk-java/blob/main/src/main/java/org/opensearch/sdk/SDKClient.java) contains methods for initializing clients in an extension for making REST requests back to OpenSearch and the tokens described above will need to be utilized by the clients.

For Extensions Developers, it will need to be clear how to use each type of token described above. In order to make it clearer which token is utilized, I propose to create 2 separate REST Clients:

1. Extension REST Client - This REST Client would always utilize the _static_ service account token (_static_ meaning that there is a single instance of the token at a time.)
2. User REST Client - This REST Client would be initialized at the start of handling a REST Request and would be used to make REST Requests back to OpenSearch on behalf of the authenticated user. Initialization means that the REST Client is provided default HTTP Headers and it would populate the `Authorization` header with the auth token provided to the REST Handler.

Additional considerations for the REST Clients and Security is that they will need to be setup to use the `https://` protocol instead of the default `http://`. Note that it is possible to disable `https` in the security plugin (but **not** transport TLS), so the setting that controls https/http in the security plugin also needs to be ferried over to an extension to determine what scheme to use.

peternied:
@cwperks Did we want to consider this item for the next security sprint, is it ready for that or out of scope?

cwperks:
@peternied Its related to this issue: https://github.com/opensearch-project/security/issues/3176

There is a checkbox in there to update this PR to reflect how Service Account Tokens are ferried to an extension. In the implementation in this PR it ferries the token over in a separate transport action that is sent directly after a success extension initialization request.

PR #892, opened 2023-07-21
Use auth tokens passed from core and introduce extension and user REST clients

cwperks:
### Description

Companion example PR for integ testing in security: https://github.com/cwperks/security/pull/6
Companion example PR for integ testing in core: https://github.com/cwperks/OpenSearch/pull/94

This PR introduces an `extensionRestClient` and `userRestClient` that an extension developer can utilize to make REST Requests with Auth Tokens back to an OpenSearch Cluster.

- The `extensionRestClient` is part of the `extensionsRunner` and is available extension-wide. This REST Client uses a Service Account Token to make requests to an OpenSearch cluster on an extension's own behalf. A common example of an extension making requests to OpenSearch on its own behalf is when interacting with a system index that the extension has reserved.

- The `userRestClient` is instantiated in `handleRequest` of `BaseExtensionRestHandler` and is available to use in all extension REST Handlers. This REST client is utilized to make requests back to opensearch on behalf of the original authenticated user utilizing an On-Behalf-Of token. The On-Behalf-Of token is a short lived access token (JWT) that includes an audience claim that is populated with the destination extension's unique ID. The On-Behalf-Of token lets an extension act as the original user. @peternied is also working on mechanisms for cluster admins to impose further restrictions on these tokens. Details can be found [here](https://gist.github.com/cwperks/e756e1cead72cd511d819241a11337e8#on-behalf-of-tokens-authorization)

The last thing this PR introduces is another security setting to be placed in the extension's settings file: `ssl.http.enabled`

This setting is used to determine the scheme (`http` or `https`) of the REST Client. This is a boolean setting.

### Issues Resolved

https://github.com/opensearch-project/opensearch-sdk-java/issues/887


src/main/java/org/opensearch/sdk/SDKClient.java
```diff
@@ -556,7 +593,7 @@ public void bulk(BulkRequest request, ActionListener<BulkResponse> listener) {
          * @return the response returned by OpenSearch
          * @throws IOException in case of a problem or the connection was aborted
          */
-        public Response performRequest(Request request) throws IOException {
+        public Response uest(Request request) throws IOException {
```
peternied, line 596: typo?

src/main/java/org/opensearch/sdk/handlers/ExtensionsRestRequestHandler.java
```diff
@@ -68,11 +73,19 @@ public RestExecuteOnExtensionResponse handleRestExecuteOnExtensionRequest(Extens
             );
         }
 
+        String oboToken = request.getRequestIssuerIdentity();
```
peternied, line 76: Seems like this should be called an onBehalfOfToken both here and on the request object?

What do you think about making the type `Optional<>` ?

peternied (commented):
Took a quick look - thanks for getting this setup, seems decent to me.  Maybe there are a couple of things that could be cleaned up by influencing https://github.com/opensearch-project/OpenSearch/pull/8526/files

src/main/java/org/opensearch/sdk/handlers/ExtensionsRestRequestHandler.java
cwperks: Agreed. I think this needs to be updated in core. I can create a PR for that.

src/main/java/org/opensearch/sdk/SDKClient.java
cwperks: Changed back to `performRequest`

